PR #24, opened 2015-01-14
Update SetInternalIds.pm

arnaudxk:
Spelling mistake, updated method get_geneomeDBAdaptor into get_genomeDBAdaptor


arnaudxk:
maybe it is being updated at te same time, as one line differs from my fork,

'compliance' from the head section is missing from my fork, and should be placed back.

muffato:
"compliance" was only missing from your fork. I have restored it

PR #27, opened 2015-04-02
Fixes for Vega 79 pair aligner.

ens-ds23:
These are fixes for our Vega pipeline config fixer for Vega 59. Many of them were actually needed for Vega 58, but I didn't have a chance to commit those.


ens-ds23:
Needs further changes. Will close this to avoid creating clutter and create another in a bit.

PR #29, opened 2015-04-02
Vega59 fixes.

ens-ds23:
Including a modification to PrepareSpeciesSetsMLSS to allow master database with SS and MLSS missing.

Feel free to rename and rearrange the param, etc, if needed.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/Example/VegaProteinTrees_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/Example/VegaProteinTrees_conf.pm
@@ -74,12 +74,12 @@ sub default_options {
     # parameters that are likely to change from execution to another:
     'mlss_id'               => '100032',   # equivalent to mlss_id for PROTEIN_TREES in the db (commented out to make it obligatory to specify)
 
-    'pipeline_name'         => 'vega_genetree_20140905_76_new', #edit this each time
+    'pipeline_name'         => 'vega_genetree_20150402_79', #edit this each time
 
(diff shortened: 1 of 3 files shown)


muffato:
OK for the modifications in the config files. It is in fact similar in spirit to the edits that are made to the PairAligner. I'll try to make something generic in a base class so that you only have to override analyses_to_remove() and tweak_analyses()

I still have to look at PrepareSpeciesSetsMLSS. Nothing wrong in principle: I just want to think carefully of the use case

muffato:
What do you have in the master db ? only the genome_dbs and their dnafrags ? and you create SS/MLSS for the alignments but not for the protein trees ?

muffato:
No: according to the config, the master db is also missing the dnafrags. So you only need the master db to define a list of species to run the pipeline on (list of genome_dbs, global SS and global MLSS)

ens-ds23:
Hi Matthieu, The only thing I was aware of was the master MLSS etc, as you suggest. I'm not positive about that, but it would be great not to have to use a master db as it's an extra step for us to prepare it. We were manually adding MLSS/SS for new species but it got out of sync at some point and neither us nor the old code noticed, so it was adding MLSS/SS for new species each time. The new code is stricter, though. Our handling of master databases is clunky, so feel free to advise.

muffato:
Hi Dan,
The pipeline already has a master-free mode but there is no way of limiting the species set to build trees on: it takes all the species (core dbs of a given version) found on the configured servers. Would it help ?

If not, what about having a Registry file with the list of the core databases you want to include ? (and no master db any more)

Matthieu

PR #33, opened 2015-04-29
script shebang and mode cleanup

dbolser-ebi:
Minor changes. Mostly just want to set +x on these .pl scripts ("update_genome.pl" in particular), but thought it best to standardize shebang before setting it on all. HTH,


muffato:
Thank you Dan. I'll merge that with other changes.

Can you let me know which scripts (among those) you guys are actually using ?

Thank you,
Matthieu

dbolser-ebi:
According to this:
http://www.ebi.ac.uk/seqdb/confluence/display/EnsGen/Compara+Peptide+Pipeline

We're using (at least):
* update_genome.pl and
* create_mlss.pl

and a quick search of our other confluence docs shows up hits for:
* copy_data.pl,
* create_pair_aligner_page.pl,
* populate_new_database.pl and

Which isn't to say we're definitely using them... nor is it to say that we're not using the others... Perhaps Arnaud, Uma, and Dan S have a better idea?

dbolser-ebi:
Sorry, seems I've added a couple three more commits on to the existing pull request. The main thing is the update to modules/Bio/EnsEMBL/Compara/PipeConfig/Example/EGProteinTrees_conf.pm

Ta,
Dan.

muffato:
I can accept the extra commits at the same time if you'd like

dbolser-ebi:
So long as you're happy with all of them, I don't mind at all!
​

PR #41, opened 2015-08-25
changed analysis "set_internal_ids_collection" memory from 100 Mb to 1Gb

etapanari:
updated the memory requirement of the "set_internal_ids_collection"  analysis because this job was failing in the hive due to memory limitation. When changing memory from 100Mb to 1Gb job was running successfully in the hive.


muffato:
Hi Electra,

Can you please make the change in PipeConfig/Example/EGPairAligner_conf.pm instead ?
At the end of it, there is a section with analysis names and their new resource-classes. This way, the change only affects EG

Thank you,
Matthieu

muffato:
You can ignore the Travis CI error: Github had issues yesterday

PR #42, opened 2015-08-28
updated memory limits for the analyses: "parse_pair_aligner_conf" and…

etapanari:
… "set_internal_ids_collection"

added more memory


waakanni:
checked..

Quick Qs... Is the new memory "set_internal_ids_collection" used anywhere in the pipeline previously?

muffato:
Yep "set_internal_ids_collection" is one of the analyses of the pipeline. Previously, when it wasn't declared in the hash in EGPairAligner_conf.pm the default resource_class was used (which happens to be 100Mb). It is now overridden.

PR #44, opened 2015-09-10
typo get_GenomeDBADaptor -> get_GenomeDBAdaptor

dbolser-ebi:
Found this in our local checkout, not sure who made the fix, but I'm committing it.


waakanni:
checked

PR #46, opened 2015-11-30
EG-specific compara merging conf 

james-monkeyshines:
Uses URLs for DB location (to be specified on the command line, rather than hard-coded in the conf file), and lacks config for compara components that are not used by EG.


muffato:
Looks OK to me. I'll let other people in the team have a look at it too

PR #47, opened 2016-01-06
Script fails if preselection file is not supplied

james-monkeyshines:
My understanding is that it is optional to provide a file containing preselected species; the script currently fails if one is not provided, so I've just added a simple check.


muffato:
This is fine by me.
I think it's something that we noticed when updating a master database for the mouse project. I'll let @carlacummins or @waakanni confirm.

carlacummins:
Looks fine to me too..

PR #48, opened 2016-01-12
Need to include tables with zero rows

james-monkeyshines:
Condition needs to check for a defined value; zero is a valid possibility.


muffato:
Why ? Tables with 0 rows can be skipped from the merge / copy process.
With the previous code (that was relying on `SHOW TABLE STATUS`), there were discarded.

james-monkeyshines:
Without this modification I got an error from line 244, stating "The table homology_member exists in protein_db but not in the target database".

muffato:
OK. I had to edit your commit. It was an oversight from me when I introduced `SELECT COUNT(*) FROM` the other day. The code was supposed to get all the tables from the target database, and all the non-empty tables from the other sources

PR #49, opened 2016-01-12
Updated configs for WormBase and WBParaSitee

khowe:
Updated protein-trees configs for WormBase


muffato:
Looks good to me. I'll let someone else from the team have a look too

waakanni:
Looks ok to me

PR #50, opened 2016-02-29
move from faidx_xs module to Bio-HTS

rishidev:


muffato:
Thank you @rishidev 
Indeed faidx_xs is not used in our API itself, but on Travis we run ensembl-rest's test-suite together with ours. So many changes in the REST API have to be backported here

Thank you

PR #51, opened 2016-03-02
Setting empty defaults in EG peptide compara config

james-monkeyshines:


mateuspatricio:
Looks fine to me.

muffato:
There is a section at the end of the file (tweak_analyses I think) where you have more power to modify the config. Fortunately, taxlevels is only used in 1 analysis I believe, so I think you can add division-specific defaults there

james-monkeyshines:
That's useful to know, thanks. However, in the case of taxlevels I think it works better for us if we don't have any defaults - it's the kind of thing that people should be thinking about when initialising the pipeline, rather than relying on defaults which might get stale...

muffato:
Can you successfully set `taxlevels` on the command-line ? Last time I tried defining an array on the command-line, eHive would consider it as a string

james-monkeyshines:
I had the same problem if I provided a single 'arrayref-as-string' parameter. But because the default value is an empty array, repeatedly using -taxlevels with a string value for each desired tax level generated an appropriate arrayref.

dbolser-ebi:
Thanks @james-monkeyshines, one question though, if we accidentally forget to be thinking about taxlevels when initialising the pipeline, will it die? i.e. is this required? What happens if we forget to provide any?

muffato:
@dbolser-ebi : need to double-check but I think that for @james-monkeyshines behaviour to work we need the parameter to have [] as the default value, which means that it can be skipped at init_pipeline because eHive is happy with []. So you wouldn't have a warning telling that you forgot to define it

james-monkeyshines:
@muffato correct, you need that [] default (I tried without, in order to force a warning, but it doesn't work). So if you forget to specify it, then you won't get any dN/dS calculations. I think that's preferable to having calculations between inappropriate species, and the former is much easier to spot than the latter.

muffato:
I'm going to accept the PR and make an additional change. I've found another way of defining a default value _for each division_ (I've checked, it really works0. So plants will still have its default, and other divisions can join the party too. The default default will be `[]`, as requested by @james-monkeyshines (I also find this better than "cellular organisms") but I'll encourage other divisions to add their default value too

PR #53, opened 2016-07-01
Fix typo in SQL.

dbolser-ebi:
I see this no longer exists in the latest branch, but fixing it here none the less as it's causing problems here.


muffato:
It's the other way: the file is a new addition on this branch :)

PR #57, opened 2016-12-21
Master

dbolser-ebi:


modules/Bio/EnsEMBL/Compara/PipeConfig/Example/EGProteinTrees_conf.pm
```diff
@@ -318,14 +318,15 @@ sub tweak_analyses {
     my $analyses_by_name = shift;
 
     ## Extend this section to redefine the resource names of some analysis
+
+    ## Here we bump the resource class of some commonly MEMLIMIT
+    ## failing analyses. Are these really EG specific?
```
muffato: They are dataset-specific, yes.

muffato:
I would prefer to leave the base config file as it is and leave each division/project sub-class implement its own resource-requirements. Probably we should be moving more Ensembl defaults to EnsemblProteinTrees to leave the base file more agnostic of our dataset

dbolser-ebi:
Personally I don't think that's a good idea. If we were short of 4 or 8 Gb machines on the farm, I'd agree. As it is we have plenty. The limiting factor here isn't waiting an extra 60 seconds here or there for a job to get submitted, it's the already large configuration burden, which I think we should seek to reduce, not expand.

This strategy makes more sense in the range of 32 and 64 Gb jobs, but again, the farm is actually very well provisioned (considering there are relatively few jobs in analyses in this range).

Obviously it's your call, this is just my opinion.

muffato:
The farm is quite big, yes, but it is still useful to have tighter resource classes:
 - LSF allocates nodes only slightly faster in the general case, I agree ...
 - ... but the difference can be notable when the farm is busy ! Often, during production, we end up in situations where say many 8Gb workers are still pending, but all the 2Gb analyses are done. 
 - It is useful when debugging because you get your node quicker
 - LSF may be penalising the user's priority for requiring memory and not using it
I'd really like to keep the minimum requirements in the default pipeline

dbolser-ebi:
Can I borrow your code for post-facto evaluation of the resource usage of each analysis?

muffato:
run `load_resource_usage.pl -url (...)` to load all the stats from LSF, and then do `SELECT * FROM resource_usage_stats`
That will tell you for each analysis the min, max, and average usage of the memory (ignore the _swap_)

dbolser-ebi:
Thanks
​

dbolser-ebi:
Thanks for merging! I'm currently doing a low level test on the LSF to work out average wait times for jobs of different sizes... Was going to let that run for one release cycle (3 months) so we'd get a fair view across a realistic span of usage. I'll let you know the results.

Cheers,
Dan.

PR #60, opened 2017-01-17
EG version of the homology stats pipeline

james-monkeyshines:
production-rh7 as default queue; extra module at start, so no need to seed; consistent parameter names with merge pipeline; use 'default' clusterset, rather than one named after the collection.


muffato:
I'm fine with this

mateuspatricio:
It looks fine to me too.

PR #63, opened 2017-03-27
Zero is a valid value in conditional, so conditional should check for defined-ness not truth

james-monkeyshines:
The consequence of this bug is that the value for "n_goc_0" is inserted into the "n_goc_null" field, and the "n_goc_0" field never gets a value (i.e. is always NULL).


muffato:
Same as https://github.com/Ensembl/ensembl-compara/commit/3371f98d387acdc8bdaa3ac523f9879d8824aa9c#commitcomment-21500993

PR #64, opened 2017-03-28
In order to validate against the OrthoXML xsd, all scores need to be defined at the top of the XML document

james-monkeyshines:


muffato:
yep, we forgot to add the new scores

PR #67, opened 2017-06-27
small bugfix

dbolser-ebi:
My pipeline fell through to this highmem analysis and promptly died with the error:
`ParamError: value for param_required('dump_min_nib_size') is required and has to be defined`

Simply adding dump_min_nib_size with the same value as for dump_large_nib_for_chains fixed the issue.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/PairAligner_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/PairAligner_conf.pm
@@ -516,6 +516,7 @@ sub pipeline_analyses {
  	       -module     => 'Bio::EnsEMBL::Compara::RunnableDB::PairAligner::DumpDnaCollection',
  	       -parameters => {
 			       'faToNib_exe' => $self->o('faToNib_exe'),
+			       'dump_min_nib_size' => $self->o('dump_min_nib_size'),
                                'overwrite'=>1,
 			      },
 	       -hive_capacity => 10,


dbolser-ebi:
BTW, are you working on something like this:
`        -rc_name => [ 'crowd', 'crowd_himem' ],`
to automatically himem through and alleviate bugs like this?

muffato:
> BTW, are you working on something like this:
> `-rc_name => [ 'crowd', 'crowd_himem' ],`
> to automatically himem through and alleviate bugs like this?

I'd like to have that feature, yes, but it's quite a major internal change. It would change several very deep assumptions. I'm afraid nothing can be done soon about that

mateuspatricio:
Hi Dan.
Regarding your previous question.
In order to decide the resources requirements prior to running a particular job, we normally rely on a metric on which we'll take the decision upon.
You can grep for "decision" in our pipeconfig directory for examples.

dbolser-ebi:
Sorry, I've subverted the pull request... I'll create an 'issue'.

dbolser-ebi:
It would be really useful if each hive could 'log' it's statistics to some central place on completion so that we could see resource usage.

muffato:
You mean the `worker_resource_usage` table and the `resource_usage_stats` view ;) ?

dbolser-ebi:
Yes, those two tables could 'phone home' so we had some uber statistics
Thanks for the merge.

PR #68, opened 2017-07-19
s/rh6/rh7/

dbolser-ebi:
Hey! That was easy!


mateuspatricio:
Looks good to me.

dbolser-ebi:
Please note I just pushed another commit.. (probably bad practice!)

dbolser-ebi:
I don't know if EG/DumpMultiAlign_conf.pm is actually used by anyone... any reason we have an EG specific version of this this pipeline?

muffato:
It is used by Production. It's a variant of the default pipeline. It dumps _everything_ regardless of when the data were created, and if there is already a dump.
We run the basic pipeline "manually", i.e. for each new mlss_id

dbolser-ebi:
Thanks.
​

PR #69, opened 2017-08-30
bumping lastz version to match the cellar

dbolser-ebi:
The new cellar, under /nfs/software/ensembl/RHEL7-JUL2017-core2, has LASTZ version 1.04.00 rather than 1.02.00.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/EBI/BaseAge_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/EBI/BaseAge_conf.pm
@@ -98,7 +98,7 @@ sub default_options {
 #            'curr_core_sources_locs'    => [ $self->o('livemirror_loc') ],
 
             # executable locations:
-            'big_bed_exe' => $self->o('ensembl_cellar').'/kent/v335/bin/bedToBigBed',
+            'big_bed_exe' => $self->o('ensembl_cellar').'/kent/v335_1/bin/bedToBigBed',
 
             #Locations to write output files
             'bed_dir'        => sprintf('/hps/nobackup/production/ensembl/%s/%s', $ENV{USER}, $self->o('pipeline_name')),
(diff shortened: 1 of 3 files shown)


dbolser-ebi:
What is the best way / recommended way to backport this to, say, release/90?

muffato:
Hi @dbolser-ebi,
Have you checked that it works ? I'm not sure we have run LastZ on the new software stack but @waakanni or @carlacummins would know.
As for having the change on e90 branch, I would prefer that _not_ to happen (because we've used 1.02 for the 30+ alignments we've computed for e90) _unless_ you can confirm the pipeline works with 1.04 and it produces the same alignments. I'd like the code on the e90 branch to be a snapshot of what we've (successfully) used to make the e90 release.

@andrewyatz : the version bump is caused to recent upstream commits https://github.com/Homebrew/homebrew-science/commits/master/lastz.rb . Were we expected to pull these changes in ? I thought we were using the 2017-02-09 snapshot as in https://github.com/Ensembl/homebrew-science

Finally, regarding "backporting" to e90, it should be done this way:
 1. Make the commit on your local branch release/90 and create a pull-request against release/90
 2. We accept the pull-request
 3. We merge the release/90 branch onto master, like we've already done a few times since we've created the branch: https://github.com/Ensembl/ensembl-compara/network

dbolser-ebi:
Thanks for the info on the procedure, I didn't know it could be done like that :-) No problem /not/ backporting it, I was just curious.

So far so good with 1.04: http://guihive.ebi.ac.uk:8080/versions/80/?driver=mysql&username=ensro&host=mysql-eg-hive.ebi.ac.uk&port=4411&dbname=dbolser_lastz_90_bdis_taes

dbolser-ebi:
BTW, is there the possibility of a test that runs a pipeline from start to finish?

andrewyatz:
Hi. No we are working from the live homebrew science. I don't think I made that repo

muffato:
@dbolser-ebi : yes it's possible, there are solutions in ensembl-test and ensembl-hive. I agree it's a major lack of the Compara test-suite, and I would be thrilled to see more production scripts / pipelines being continuously tested. Currently only the few pipelines written recently by @waakanni and @carlacummins (Orthology QC) are

dbolser-ebi:
Thanks, sorry for going off topic. I may put this as a feature request.
There we can link the appropriate examples.
​

dbolser-ebi:
Just added a similar commit to this PR, namely adjusting the location of kent libraries.

carlacummins:
Sorry for the late reply. 

Yes, I also had this to commit to make. Pipeline appears to work as expected with 1.04.

PR #74, opened 2018-02-01
Explicitly set a default queue for HighConfidenceOrthologs pipeline

james-monkeyshines:


muffato:
Hi James
Instead of setting the queues in every single resource_class, we do it across the board with an environment variable:
```export LSB_DEFAULTQUEUE=production-rh7```

james-monkeyshines:
Yes, I know, but that's not documented anywhere (that I can find...), so people may not have that variable set.
But I agree that setting the parameter redundantly in every pipeline is a poor solution; perhaps it could be set in 'ComparaGeneric_conf' instead? Anyway, I'll leave that to you decide, and in the meantime I'll update the EG compara docs with the LSB_DEFAULTQUEUE information.

PR #81, opened 2018-06-19
Pre release 94 deprecation

at7:
I have an ongoing effort with core to remove variation dependencies from the core API. It turns out there is some code called on AlignSlice objects. I cannot remove the ones that are used in the webcode: AlignSlice::Slice::get_all_VariationFeatures. But get_all_VariationFeatures_by_VariationSet and get_all_genotyped_VariationFeatures are not used in the webcode and therefor can be removed. Let me know if you have any concerns. Also let me know if remove for e95 is fine or if you prefer a later release. Thanks.

### DIFF
--- a/DEPRECATED.md
+++ b/DEPRECATED.md
@@ -10,6 +10,11 @@
 
 # Deprecated methods scheduled for deletion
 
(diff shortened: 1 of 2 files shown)


mateuspatricio:
Looks OK to me.

PR #84, opened 2018-09-13
The dumping script is expecting a division name as the -reg-alias parameter

james-monkeyshines:


muffato:
Unfortunately the division name is not always the correct Registry name for the compara database. I think you can instead set `rel_db` when initializing the pipeline, possibly with `-rel_db '#division#'`

james-monkeyshines:
I do set `rel_db` in the pipeline, but other modules require it to be a url, e.g. factory_homology_range_dumps, dump_all_homologies_orthoxml...

muffato:
Those analyses should understand a registry name (that's how we run the pipeline). Have you changed `--production_registry` ?

james-monkeyshines:
I specify `-production_registry "--reg_conf $HOME/work/registries/eg-s1r.94.pm"`
So, setting `rel_db` to the name of the compara db should work, in that case?

muffato:
I think it should work with the "registry" name, yes.

james-monkeyshines:
OK thanks, I'll try that next time.

PR #85, opened 2018-09-13
dataflow_output_id needs a proper array of hashref from copy_dataflow

MatBarba:
This fix prevents "copy_compara_table" to fail in the OrthologQM_Alignment pipeline. This may also happen in 94.


muffato:
Thank you @MatBarba 
We usually run the pipeline with `alt_aln_dbs`, thus not going into the `else`

PR #87, opened 2018-10-29
Only compute Jaccard index if there are two databases to compare

wbazant:
I'm running a pipeline without reuse: `reuse_db` in `pipeline_wide_parameters` is set to `undef`. Also as far as I see, the output of the job just gets printed to stdout which is usually discarded. So I prefer if the job doesn't even get created in my pipeline.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/ProteinTrees_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/ProteinTrees_conf.pm
@@ -3096,7 +3096,7 @@ sub core_pipeline_analyses {
             -parameters         => {
                 mode            => 'stable_id_mapping',
             },
-            -flow_into          => [ 'compute_jaccard_index' ],
+            -flow_into          => WHEN( '#reuse_db#' => 'compute_jaccard_index' ),
             %hc_analysis_params,
         },
 


muffato:
That's one for @mateuspatricio to review. He's fixed that on the e95 branch (ENSCOMPARASW-1644) and will have to merge the changes

mateuspatricio:
Looks like your code is old. There was a commit to deal with cases when the Jaccard Index wasn't needed. https://github.com/Ensembl/ensembl-compara/commit/3292c2d5b7084e260ba694ffd868c223e360baa1

mateuspatricio:
Wait, I just realised. Are you using our master? You should be using the e94 or e95 branches.

wbazant:
I've ran 6ebb76027a4c8d, it was head of release/94 on 22 August when I did the checkout.

3292c2d will at least let me avoid the bug, because I can set do_jaccard_index to 0. I'll close the issue, thank you! 

The bug is still there ( Jaccard index getting calculated when it can't be because of reuse_db ) - feel free to reconsider default behaviour if you're inspired.

Also I saw
```
 open( my $plot_file, ">", $self->param('output_jaccard_file') ) || die "Could not open plot_file";
```
if it was `$plot_file` it would be easier to tell what's not being opened.

muffato:
I've already fixed `$plot_file` on the e95 branch too

PR #91, opened 2019-02-27
Recognise more segfaults

wbazant:
I had a job die on a segfault, except it had an error code 139 and not 11.

I'm not sure if exit code 11 is special for treebest. I don't know if there are cases where the exit code is 11, the log doesn't say segfault, and it is a segfault, but this change preserves the existing behaviour in that scenario.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/TreeBest.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/TreeBest.pm
@@ -123,10 +123,10 @@ sub run_treebest_best {
         return $run_cmd->out unless ($run_cmd->exit_code);
 
         my $full_cmd = $run_cmd->cmd;
-        $self->throw("'$full_cmd' resulted in a segfault") if ($run_cmd->exit_code == 11);
 
         print STDERR "$full_cmd\n";
         my $logfile = $run_cmd->err;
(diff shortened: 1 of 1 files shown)


muffato:
11 is the UNIX signal for segmentation fault. That's what we get when the command is run directly. When it's run via `bash`, I think the latter returns 128+11=139

wbazant:
I got this error. 
```
'ARRAY(0x89cd4e8)' resulted in a segfault at /nfs/production/panda/ensemblgenomes/wormbase/software/packages/ensembl/branches/branch-95/ensembl-hive/modules/Bio/EnsEMBL/Hive/Process.pm line 613.
```

I also get ARRAY(0x89cd4e8) in my STDERR - apparently `$full_cmd` is an arrayref, not a string as the code assumes now.

muffato:
Ah, yes. This needs some slight rewriting. `$run_cmd` has a method named `die_with_log` that could be used, but perhaps with an option to add "because of a segfault" or something like that

wbazant:
@muffato I'll have to leave it with you guys, I don't have a good way of testing changes any more. I was running it through Hive and don't have a way of reproducing this now. I know you can change stuff with confidence but I probably shouldn't.

If the job dies with log and exit status ( after recognizing "segmentation fault" in a log, or exit code 11) I think the message doesn't need to say "because of segfault" - and as it's an extra level of interpretation it could actually be confusing or wrong. It will be a very good message if it communicates the log, the status, and where it died.

muffato:
OK, I'll change that on master later today

PR #92, opened 2019-04-17
Update CopyData.pm

epaule:
that is more for future compatibility with newer MySQL versions that don't like function names as column ids.


muffato:
Have you checked this works and it's not running `$_` as a system command and using its output ?

epaule:
erm ... yes. I am using it already. And there is no system shell stuff in
the SQL statements.

PR #93, opened 2019-05-02
Added script to extract taxonomy information

JAlvarezJarreta:
Commit related with issue [ENSCOMPARASW-2193](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-2193).

### DIFF
--- a/scripts/taxonomy/GetTaxonomyTable.pl
+++ b/scripts/taxonomy/GetTaxonomyTable.pl
@@ -49,7 +49,8 @@ my @target_ranks = qw(class superorder order suborder family subfamily species);
 # Number of ranks (plus the default addition of "GenomeDB name" as last rank)
 my $num_ranks = $#target_ranks + 1;
 # Create the taxonomy template: if a rank is missing, assign "N/A"
-my %rank_template = map {$_ => 'N/A'} @target_ranks;
+my $null_value = 'N/A';
+my %rank_template = map {$_ => $null_value} @target_ranks;
 # @taxonomy_table will have one string per species' taxonomy in tab-separated
 # values (TSV) format
 my @taxonomy_table;
@@ -70,13 +71,9 @@ foreach my $gdb (@{$list_of_gdbs}) {
         }
         $node = $node->parent();
     }
-    # Get the taxonomy in a TSV-like string and append it to the taxonomy table
-    my $taxonomy_str = "";
-    foreach my $rank (@target_ranks) {
-        $taxonomy_str .= $taxonomy{$rank} . "\t";
-    }
-    # Add GenomeDB name as the last level/rank of the taxonomy
-    $taxonomy_str .= $gdb->name();
+    # Get the taxonomy in a TSV-like string, add GenomeDB name as the last
+    # level/rank and append it to the taxonomy table
+    my $taxonomy_str = join("\t", (map {$taxonomy{$_}} @target_ranks), $gdb->name());
     push(@taxonomy_table, $taxonomy_str);
 }
 # Sort the rows in alphabetical order
@@ -89,13 +86,27 @@ if ($group_ranks) {
         my @taxonomy = split(/\t/, $taxonomy_table[$i]);
         # Grouped version of the taxonomy (string)
         my $grouped_taxonomy = "";
+        # Flag when two consecutive species have a not-null matching rank
+        my $not_null_match_found = 0;
         # Loop over each rank to compare it with the value in the previous row
         foreach my $j (0 .. $num_ranks) {
             if ($prev_taxonomy[$j] eq $taxonomy[$j]) {
-                $grouped_taxonomy .= "''\t";
+                # Get the flag to True if the ranks are not null and equal
+                if ($taxonomy[$j] ne $null_value) {
+                    $not_null_match_found = 1;
+                }
             } else {
+                # If there is a not-null match, all the ranks matched can be
+                # replaced by "''". If not, it means all the ranks were missing
+                # until this point, so leave the null value.
+                my @values;
+                if ($not_null_match_found) {
+                    @values = ("''") x $j;
+                } else {
+                    @values = ($null_value) x $j;
+                }
                 # As soon as one rank is different, the remaining will be too
-                $grouped_taxonomy .= join("\t", @taxonomy[$j .. $num_ranks]);
+                $grouped_taxonomy .= join("\t", @values, @taxonomy[$j .. $num_ranks]);
                 last;
             }
         }
--- a/scripts/taxonomy/GetTaxonomyTable.pl
+++ b/scripts/taxonomy/GetTaxonomyTable.pl
@@ -17,99 +17,95 @@
 use strict;
 use warnings;
 use Getopt::Long;
-use Cwd qw(abs_path);
-use File::Spec::Functions qw(catfile);
 use Bio::EnsEMBL::Compara::DBSQL::DBAdaptor;
 
-$| = 1;
-
 # Parameters:
-my $help = 0;
 # URL to the Compara database
 my $compara_url;
-# Output directory where the taxonomy table will be saved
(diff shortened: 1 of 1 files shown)


scripts/taxonomy/GetTaxonomyTable.pl
```diff
@@ -0,0 +1,115 @@
...
+    # Copy taxonomy template and fill it in
+    my %taxonomy = %rank_template;
+    while ($node->name() ne 'root') {
+        my $rank = $node->rank();
+        if (defined $rank && exists $taxonomy{$rank}) {
+            $taxonomy{$rank} = $node->name();
+        }
+        $node = $node->parent();
+    }
+    # Replace species name by GenomeDB name
```
muffato: I actually meant to print the GenomeDB name alongside the species name that comes from the taxonomy. For most genomes `aaa_bbb` the species name is _Aaa bbb_ but not always, e.g. trinomial names, strains

scripts/taxonomy/GetTaxonomyTable.pl
```diff
@@ -0,0 +1,115 @@
...
+        }
+        $level = $level->{$classification};
+    }
+}
+# Flatten the taxonomy hash into an array of one element per species
+my @taxonomy_table;
+flatten($taxonomy_db, \@taxonomy_table);
+# Write the array into the output file with @target_ranks as headers
+my $out_abs_path = catfile($out_dir, $out_file);
+open(my $file, ">$out_abs_path") or
```
muffato: `perlcritic` advises the 3-argument form, i.e.
```open(my $file, '>', $out_abs_path)```

File: scripts/taxonomy/GetTaxonomyTable.pl [hunk 3]
```diff
@@ -0,0 +1,115 @@
...
+    # Replace species name by GenomeDB name
+    if (grep(/^species$/, @target_ranks)) {
+        $taxonomy{'species'} = $ref->name();
+    }
+    # Push the template into the taxonomy hash
+    my $level = $taxonomy_db;
+    foreach my $rank (@target_ranks) {
+        my $classification = $taxonomy{$rank};
+        if (! exists($level->{$classification})) {
+            # Each rank has a hash as value except "species" (this has a 1)
```
muffato: By adding the genome_db name as an extra column at the end, I think you wouldn't need this special treatment here. Always set the value to `{}` and after the loop you may do ```$level->{$genome_db->name} = 1```

scripts/taxonomy/GetTaxonomyTable.pl
```diff
@@ -0,0 +1,115 @@
...
+        if (defined $rank && exists $taxonomy{$rank}) {
+            $taxonomy{$rank} = $node->name();
+        }
+        $node = $node->parent();
+    }
+    # Replace species name by GenomeDB name
+    if (grep(/^species$/, @target_ranks)) {
+        $taxonomy{'species'} = $ref->name();
+    }
+    # Push the template into the taxonomy hash
```
muffato: Here is a very important part of the script that requires more explanation / better naming.
At this stage, you could just print the line immediately:
```print $file join("\t", map {$taxonomy{$_}} @target_ranks), "\n"; ```
but the output wouldn't be sorted by clades

What you're building in `$taxonomy_db` is essentially the NCBI tree that links all those genomes, but restricted to the allowed ranks. You then use that to sort the genomes according to the taxonomy, and you need to introduce `flatten`. `$taxonomy_db` in fact allows the grouping that Emily was showing in her diagram, e.g. only print "Mammalia" once

1) If we only care about ordering the genomes, you could simplify this by sorting the genome_dbs by the `left_index` of their NCBITaxon object (see https://en.wikipedia.org/wiki/Nested_set_model for an explanation of the "left" and "right" indices. They are used in all our tree tables). Maybe pre-sort the list of genome_dbs and print them on the fly. It's be clearer

2) But since you have the structure that allows the grouping, maybe you could add an option to display it ? e.g. instead of
```
Aves    Neognathae      Anseriformes    N/A     Anatidae        Anatinae        anas_platyrhynchos_platyrhynchos
Aves    Neognathae      Anseriformes    N/A     Anatidae        Anserinae       anser_brachyrhynchus
Aves    Neognathae      Charadriiformes N/A     Scolopacidae    N/A     calidris_pugnax
Aves    Neognathae      Charadriiformes N/A     Scolopacidae    N/A     calidris_pygmaea
```
print
```
Aves    Neognathae      Anseriformes    N/A     Anatidae        Anatinae        anas_platyrhynchos_platyrhynchos
"       "               "               "       "               Anserinae       anser_brachyrhynchus
"       "               Charadriiformes N/A     Scolopacidae    N/A     calidris_pugnax
"       "               "               "       "               "       calidris_pygmaea
```

File: scripts/taxonomy/GetTaxonomyTable.pl [hunk 5]
```diff
@@ -0,0 +1,115 @@
...
+# Parameters:
+my $help = 0;
+# URL to the Compara database
+my $compara_url;
+# Output directory where the taxonomy table will be saved
+my $out_dir = Cwd::abs_path();
+# Handle input arguments
+GetOptions("help"          => \$help,
+           "compara_url=s" => \$compara_url,
+           "outdir=s"      => \$out_dir
```
muffato: The simplest is not to have any option, just print to the standard output and people can redirect the output wherever they want with `>` or even pipe it into something.
If there is an option in this script, I would expect it to be the path to the output file, i.e. `-o /path/to/dest/file`. This way, the script can be run with its output in a given file without needing a shell.
I find having an option to change the output directory unusual. Imagine the script is going to be run across several compara databases (vertebrates, plants, fungi): you need to run it in three different directories, otherwise the output files will be overwritten

scripts/taxonomy/GetTaxonomyTable.pl
```diff
@@ -0,0 +1,115 @@
...
+# limitations under the License.
+
+use strict;
+use warnings;
+use Getopt::Long;
+use Cwd qw(abs_path);
+use File::Spec::Functions qw(catfile);
+use Bio::EnsEMBL::Compara::DBSQL::DBAdaptor;
+
+$| = 1;
```
muffato: This is only useful when the script is printing its ouput in a "stream" manner. Here you write everything at once at the end, so there won't be any visible differences

scripts/taxonomy/GetTaxonomyTable.pl
```diff
@@ -0,0 +1,111 @@
...
+    # Group ranks in taxonomy table, replacing duplicates by "''"
+    # Get the first row as reference to see how many ranks are the same
+    my @prev_taxonomy = split(/\t/, $taxonomy_table[0]);
+    foreach my $i (1 .. $#taxonomy_table) {
+        my @taxonomy = split(/\t/, $taxonomy_table[$i]);
+        # Grouped version of the taxonomy (string)
+        my $grouped_taxonomy = "";
+        # Loop over each rank to compare it with the value in the previous row
+        foreach my $j (0 .. $num_ranks) {
+            if ($prev_taxonomy[$j] eq $taxonomy[$j]) {
```
muffato: Looking at the output, I think that 'N/A' should be treated in a special manner when it comes to grouping.

These four lines can be grouped because there have the same taxon names after the initial 'N/A's
```
N/A     N/A     Testudines      Cryptodira      Emydidae        N/A     Chrysemys picta chrysemys_picta_bellii
N/A     N/A     Testudines      Cryptodira      Testudinidae    N/A     Chelonoidis abingdonii  chelonoidis_abingdonii
```
But not these two
```
N/A     N/A     Petromyzontiformes      N/A     Petromyzontidae N/A     Petromyzon marinus      petromyzon_marinus
N/A     N/A     Testudines      Cryptodira      Emydidae        N/A     Chrysemys picta chrysemys_picta_bellii
```

PR #99, opened 2019-05-30
Added retirement of species sets not longer used in any mlss

JAlvarezJarreta:
Ticket [ENSCOMPARASW-2224](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-2224).

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/PrepareMasterDatabaseForRelease_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/PrepareMasterDatabaseForRelease_conf.pm
@@ -290,7 +290,7 @@ sub pipeline_analyses {
             -flow_into => [ 'set_last_release_to_mlss' ],
         },
 
-        {   -logic_name => 'set_last_release_to_mlss',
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/Compara/PipeConfig/PrepareMasterDatabaseForRelease_conf.pm
```diff
@@ -287,6 +287,15 @@ sub pipeline_analyses {
                 'cmd'                   => 'perl #create_all_mlss_exe# --reg_conf #reg_conf# --compara #master_db# -xml #xml_file# --release --output_file #report_file# --verbose',
             },
             -rc_name        => '2Gb_job',
+            -flow_into => [ 'set_last_release_to_mlss' ],
+        },
+
+        {   -logic_name => 'set_last_release_to_mlss',
```
carlacummins: Code is perfect, but I would update the logic name for clarity, since we're not actually touching MLSSes, only species_sets

```suggestion
        {   -logic_name => 'retire_old_species_sets',
```

PR #102, opened 2019-06-18
Do not assume a division is an Ensembl division

wbazant:
I came across this while running the Load Members pipeline. It wouldn't start, because there isn't a `parasite` conf in the source code, and calling `check_file_in_ensembl` doesn't let me overwrite it., I think someone already had this problem because there's something similar in the `#NOTE:` comment above `binary_species_tree`.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/ENV.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/ENV.pm
@@ -74,8 +74,8 @@ sub shared_default_options {
         'linuxbrew_home'        => $ENV{'LINUXBREW_HOME'},
 
         # All the fixed parameters that depend on a "division" parameter
-        'reg_conf'              => $self->check_file_in_ensembl('ensembl-compara/scripts/pipeline/production_reg_'.$self->o('division').'_conf.pl'),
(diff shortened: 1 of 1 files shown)


muffato:
Hi Wojtek,
Out of curiosity, how do you configure the various database locations in your pipeline ? Do you define hashes (`-host => `, `-port =>`, etc) or do you use your own registry ?
You have probably noticed that all the resource classes now include `--reg_conf`. We don't need to edit most PipeConfigs every release any more, now that they all point at the registry file and we change the db locations in the latter.

wbazant:
We write out a registry once per release, and then --reg_conf in the instructions to beekeepers. Compara has its own registry, including also the locations of Compara db.

Gory details: https://www.ebi.ac.uk/seqdb/confluence/display/WORMBASE/ParaSite+Compara+Production

muffato:
OK. I will accept this PR and copy the commit to `release/98`, but I will also add a check that the final value `reg_conf` parameter must be a valid path. This is necessary because all the resource-classes embed a `--reg_conf` argument for the workers. In your instructions, you should add the `--reg_conf` parameter on the `init_pipeline` command-line of WBParaSiteProteinTrees_conf
BTW, this means you don't need to use `--reg_conf` when running beekeeper, as long as the database is referred by url.

PR #103, opened 2019-06-27
Add custom mapped_gene_ratio_per_taxon for the WBPS config stored in this repo

wbazant:
The default of 0.5 unique genes is not good enough as we have some really weird animals - also, their genomes are not very good. 

Proportions from WBPS14 run are in: https://gist.github.com/wbazant/7b8f398ff43f764b6acfae6051285102


muffato:
Cherry-picked onto `release/98` too

PR #106, opened 2019-07-01
Remove an unused parameter from ProteinTrees_conf.pm

wbazant:
Remove a parameter that doesn't do anything: the jaccard index and gini coefficient are calculated in one script.

Came to look at this because we don't have a reuse_db set and the job didn't run - but you've fixed this already.


muffato:
@mateuspatricio . Can you check why this parameter is not used ?

mateuspatricio:
Yes, it's OK to remove, Gini coefficient and Jaccard Index are computed by the same runnable.

PR #107, opened 2019-07-01
added triticum_turgidum self-alignments

brunocontrerasmoreira:
Added triticum_turgidum to list of species to do self-alignments. 
This species has subgenomes A & B in the species tree.


brunocontrerasmoreira:
Since triticum_turgidum was not repeat-masked we cannot do this in 98

PR #108, opened 2019-07-01
clusterset_id for stats jobs is not set everywhere

wbazant:
The job `set_default_values` came out like that in my run of the ProteinTrees pipeline:

```
Bio::EnsEMBL::Hive::RunnableDB::SqlCmd     | {"clusterset_id" => undef,"sql" => ["DELETE FROM gene_member_hom_stats","INSERT INTO gene_member_hom_stats (gene_member_id, collection)\n                     SELECT gene_member_id, \"#clusterset_id#\"\n                     FROM gene_member"]}
```

This failed, because SqlCmd didn't like inserting with the undefined value.

This is probably not a robust fix, it tries to replicate the previous behaviour but I don't really have the big picture for making this right.

@mateuspatricio do have a look! think after 1ee5d93ecc74 not every pipeline will have a set `collection` which should go there - the commit removes it from INPUT_PLUS and a few more places, and sets it in PostHomologyMerge_conf.pm, but not everywhere else.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/Parts/GeneMemberHomologyStats.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/Parts/GeneMemberHomologyStats.pm
@@ -102,7 +102,7 @@ sub pipeline_analyses_hom_stats {
         {   -logic_name => 'set_default_values',
             -module     => 'Bio::EnsEMBL::Hive::RunnableDB::SqlCmd',
             -parameters => {
-                'clusterset_id' => $self->o('collection'),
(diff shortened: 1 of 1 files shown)


wbazant:
@muffato could we make it a default in the ProteinTrees PipeConfig?

There might also be other analyses that run these jobs, also without this default.

muffato:
Yep. Let's do that. Move `'collection' => 'default'` from PipeConfig/EBI/Ensembl/ProteinTrees_conf.pm to PipeConfig/ProteinTrees_conf.pm
Would I do it ?

wbazant:
See #110

PR #109, opened 2019-07-02
stats_homologies jobs create different temp tables

wbazant:
This PR proposes making the analysis create different temp tables for each job in the fan, so that they can happen simultaneously without a lock.

In my run they failed as follows:

```
BD::mysql::db do failed: The total number of locks exceeds the lock table size at /nfs/production/panda/ensemblgenomes/wormbase/software/packages/ensembl/branches/branch-97/ensembl-hive/modules/Bio/EnsEMBL/Hive/DBSQL/CoreDBConnection.pm line 983.\n at /nfs/production/panda/ensemblgenomes/wormbase/software/packages/ensembl/branches/branch-97/ensembl-hive/modules/Bio/EnsEMBL/Hive/DBSQL/CoreDBConnection.pm line 1004.\n\tBio::EnsEMBL::Hive::DBSQL::CoreDBConnection::__ANON__(Bio::EnsEMBL::Hive::DBSQL::DBConnection=HASH(0x2872730), "CREATE TEMPORARY TABLE temp_member_hom_counts AS\\x{a}            "...)
```

This was before there was `hive_capacity` 5 on the job, and anyways I have worked around the issue by doing
```
tweak_pipeline.pl -url $EHIVE_URL --tweak "analysis[stats_homologies].analysis_capacity=1"
```
and rerunning.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/Parts/GeneMemberHomologyStats.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/Parts/GeneMemberHomologyStats.pm
@@ -160,9 +160,9 @@ sub pipeline_analyses_hom_stats {
             -parameters => {
                 'sql'   => [
                     # Clean up partial runs
-                    'DROP TABLE IF EXISTS temp_member_hom_counts',
+                    'DROP TABLE IF EXISTS temp_member_hom_counts_#min_gene_member_id#',
                     # Temporary table with the counts
-                    'CREATE TEMPORARY TABLE temp_member_hom_counts AS
+                    'CREATE TEMPORARY TABLE temp_member_hom_counts_#min_gene_member_id# AS
                      SELECT gene_member_id, SUM(method_link_id=201) AS orthologues, SUM(method_link_id=202) AS paralogues, SUM(method_link_id=206) AS homoeologues
                      FROM homology_member
                      STRAIGHT_JOIN homology ON homology_member.homology_id = homology.homology_id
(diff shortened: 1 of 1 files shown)


wbazant:
Actually I now think this PR won't do anything, and is unnecessary. Unless you disagree and I'm wrong about being wrong, do close it :$

http://www.mysqltutorial.org/mysql-temporary-table/
"Different clients can create temporary tables with the same name without causing errors because only the client that creates the temporary table can see it."

 I think this won't work because I've updated the jobs to run five at a time with differently named temp tables, and still got "the total number of locks exceeds the total table size" on two of them. The others are still running.

I'll let the queries run overnight, and if it doesn't work I'll come ask for help with this job (what it's for, how it usually runs, and if you tried other query plans that get the same information e.g. counting homology_member.description)

muffato:
Yep, I knew that concurrent sessions could create temporary tables with the same name, but I thought that somehow this was solving the locking issues on your server. I wasn't terribly surprised, MySQL servers at the EBI often have different parameters.
Anyway, even if that's not needed, I still like your trick to make the names unique. It would also allow removing the `TEMPORARY`, which would perhaps make debugging easier ?
We can discuss the purpose of the analysis and the table it generates tomorrow

muffato:
Hi @wbazant 
We discussed a few strategies for this query. Which one worked the best for you ?

wbazant:
@muffato I've not done the alternative query based on investigating the field homology_member.description to get the counts, after Carla's fix the job worked fast enough (maybe 20 mins each, if I remember), and I've never came back to my Compara to try the other one. I'll set it up to benchmark it tonight.

wbazant:
@muffato yeah it's slower to look at the description column than the current average (24 minutes):

mysql> CREATE TEMPORARY TABLE temp_member_hom_counts_1 AS                     SELECT gene_member_id, SUM(description LIKE "ortholog%") AS orthologues, SUM(description LIKE "%paralog%" OR description = "gene_split") AS paralogues, SUM(description LIKE "%homoeolog%" OR description = "alt_allele") AS homoeologues FROM homology_member JOIN homology ON homology_member.homology_id = homology.homology_id WHERE gene_member_id BETWEEN 1 AND 1000000 GROUP BY gene_member_id;
Query OK, 846981 rows affected (46 min 13.46 sec)
Records: 846981  Duplicates: 0  Warnings: 0

mysql> CREATE TEMPORARY TABLE temp_member_hom_counts_2 AS                     SELECT gene_member_id, SUM(description in ("ortholog_one2one", "ortholog_one2many", "ortholog_many2many")) AS orthologues, SUM(description in ("within_species_paralog", "other_paralog", "gene_split")) AS paralogues, SUM(description in ("alt_allele", "homoeolog_one2one", "homoeolog_one2many", "homoeolog_many2many")) AS homoeologues FROM homology_member JOIN homology ON homology_member.homology_id = homology.homology_id WHERE gene_member_id BETWEEN 1 AND 1000000 GROUP BY gene_member_id;
Query OK, 846981 rows affected (32 min 46.83 sec)
Records: 846981  Duplicates: 0  Warnings: 0

muffato:
Thanks for testing, I'll update the query accordingly (on master).

PR #111, opened 2019-07-12
Include the last ensembl-analysis modules

muffato:
This is the last piece of work for [ENSCOMPARASW-1488](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-1488): migrating the pairwise-aligner pipeline (LastZ and TBlat). I have copied the ensembl-analysis modules with their full history, and then eHive-ified them (i.e. made them use `$self->param`, etc).
The result is that:
1. we don't need an ensembl-analysis checkout any more for Compara (which is quirky to set up, e.g. `General.pm` etc)
2. we have full control on the implementation of LastZ, which will help making changes like reading from the genome dumps etc [ENSCOMPARASW-1727](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-1727)

I have also modified the production table `dnafrag_chunk`, which allows reusing the Locus object and other methods.


modules/Bio/EnsEMBL/Compara/RunnableDB/PairAligner/AlignmentChains.pm
```diff
@@ -162,54 +148,51 @@ sub fetch_input {
...
       # If there is no .nib file, preload the sequence
-      if ($query_nib_dir and (-d $query_nib_dir) and (-e $query_nib_dir . "/" . $query_slice->seq_region_name . ".nib")) {
-          print STDERR "reusing the query nib file\n";
-          $parameters{'-query_nib_dir'} = $query_nib_dir;
+      if ($query_nib_dir and (-d $query_nib_dir) and (-e $query_nib_dir . "/" . $self->param('query_dnafrag')->name . ".nib")) {
+          print STDERR "reusing the query nib file ". $query_nib_dir . "/" . $self->param('query_dnafrag')->name . ".nib" . "\n";
       } else {
           print STDERR "fetching the query sequence\n";
+          my $query_slice = $self->param('query_dnafrag')->slice;
+          $self->param('query_slice', $query_slice);
```
carlacummins: Why do we only want to store the query_slice as a param if we're not reusing a .nib file?

carlacummins: Found the answer in Bio::EnsEMBL::Compara::Production::Analysis::AlignmentChains

modules/Bio/EnsEMBL/Compara/RunnableDB/PairAligner/PairAligner.pm
```diff
@@ -145,53 +130,11 @@ sub fetch_input {
 		"  your system can cope with so many alignments.");
       }
   }
-  $self->compara_dba->dbc->disconnect_if_idle();
```
carlacummins: Why is this not needed anymore?

modules/Bio/EnsEMBL/Compara/RunnableDB/PairAligner/AlignmentChains.pm
muffato: Nice. I didn't remember and was digging back into the code :)

modules/Bio/EnsEMBL/Compara/RunnableDB/PairAligner/PairAligner.pm
muffato: Because it was the last line in `fetch_input` and eHive will immediately reopen the connection to change the job status to RUN

carlacummins:
Are ImportChains and ImportNets just for legacy support? I can't find a reference to them in the committed code here

muffato:
Yes, they'e used by this pipeline: PipeConfig/Legacy/ImportUcscChainNet_conf.pm
I don't think it's ever been run since I joined

modules/Bio/EnsEMBL/Compara/RunnableDB/PairAligner/Blat.pm
```diff
@@ -108,36 +110,15 @@ sub configure_runnable {
...
-  $self->param('runnable', []);
-  my $runnable = Bio::EnsEMBL::Compara::Production::Analysis::Blat->
-    new(
-	-query      => $qyChunkFile,
-	-database   => $dbChunkFile,
-	-options    => $options,
-       );
-  push @{$self->param('runnable')}, $runnable;
+  my $starttime = time();
+  my $output = Bio::EnsEMBL::Compara::Production::Analysis::Blat::run_blat($self, $qyChunkFile, $dbChunkFile);
```
carlacummins: Why do we need to call it like this instead of just `$self->run_blat($qyChunkFile, $dbChunkFile);` ?

carlacummins:
All looks good to me! It's quite a large commit, so I hope I haven't missed anything. Am happy to merge.

PR #112, opened 2019-07-15
Removed test_mode flag and RegisterMLSS runnable

JAlvarezJarreta:
`test_mode` flag is now obsolete, making `RegisterMLSS` runnable obsolete too.

JIRA ticket [ENSCOMPARASW-2430](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-2430).


modules/Bio/EnsEMBL/Compara/PipeConfig/EpoLowCoverage_conf.pm
```diff
@@ -403,22 +403,14 @@ sub pipeline_analyses {
                                                                            {'test' => 'conservation_scores','method_link_species_set_id'=>'#cs_mlss_id#'},
                                                                 ],
                                     } ),
-                               'A->1' => ['register_mlss'],
+                               'A->1' => WHEN( 'not #skip_multiplealigner_stats#' => [ 'multiplealigner_stats_factory' ] ),
```
muffato: in eHive, the _funnel_ is mandatory, so you need to make sure there is always a job created for `A->1`. You can perhaps add an _ELSE_ clause that points at a Dummy analysis. Or maybe the pipeline can be slightly restructured.

modules/Bio/EnsEMBL/Compara/PipeConfig/Parts/EPOAlignment.pm
```diff
@@ -365,7 +365,7 @@ return
                                                                            {'test' => 'conservation_scores'},
                                                                 ],
                                     } ),
-                               'A->1' => ['register_mlss'],
+                               'A->1' => WHEN( 'not #skip_multiplealigner_stats#' => [ 'multiplealigner_stats_factory' ] ),
```
muffato: Same comment as above.

modules/Bio/EnsEMBL/Compara/PipeConfig/EpoLowCoverage_conf.pm
JAlvarezJarreta: I think adding an _ELSE_ Dummy analysis seems the best option. The restructuration seems more complex and would still need the last Dummy as this analysis comes from a previous Factory-Fan-Funnel group.

PR #114, opened 2019-07-16
updated goc_taxlevels

brunocontrerasmoreira:
Updated the plant clades for which GOC is computed to include recently added species


muffato:
@JAlvarezJarreta . If the pipeline hasn't reach that stage yet (I don't think it has), then please update the parameters to use them in e98

JAlvarezJarreta:
The changes should not have affected Travis CI but it is failing.

PR #115, opened 2019-07-22
Remove trailing whitespace in newick trees

muffato:
This would make the convention used in the Plants species-tree to add trailing commas everywhere redundant.
The code now handles graciously unnamed nodes (instead of warning about undefined values in several places).


carlacummins:
All looks good to me!

PR #119, opened 2019-08-22
Added some helper methods to run commands

muffato:
I have looked at all the occurrences of the \` operator, `system`, and `open` (when used to read from / write to a command), and converted them to use new helper methods, all flavours of BaseRunnable's `run_command`. I have added these methods in BaseRunnable:
- `get_command_output`. A replacement for \`
- `read_from_command`. A replacement for some `open` calls. A callback method will be called with the file-handle of the command's standard input
- `write_to_command`. Similar to `read_from_command` but the file-handle allows reading from the command's standard output

In all three methods, `die_on_failure` is automatically set, unless told otherwise. I think it is a safer default. I could make it the default in `run_command` too since it is set in about two thirds of the calls ?

Advantages of using these helper methods are:
- automatic disconnection from the database
- capture of errors, standard output and standard error
- logging of the command run-time
- unified way of showing debug information
- aware of LSF's killing mechanism (will sleep to allow a proper MEMLIMIT kill)


modules/Bio/EnsEMBL/Compara/Production/EPOanchors/FindDfrStrand.pm
```diff
@@ -108,11 +108,11 @@ sub fetch_input {
 	foreach my $query_file (@$q_files) {
 		foreach my $target_file (@$t_files) {
 			my $command = $self->_bl2seq_command( $query_file, $target_file );
-			my $bl2seq_fh;
-			open($bl2seq_fh, "$command |") or throw("Error opening command: $command"); # run the command
-			# parse_bl2seq returns a hashref of the scores and the number of hits to each query strand
-			push(@$blastResults, $self->parse_bl2seq($bl2seq_fh));
-			close($bl2seq_fh);
+			$self->read_from_command($command, sub {
+				my $bl2seq_fh = shift;;
```
carlacummins:
```suggestion
				my $bl2seq_fh = shift;
```

modules/Bio/EnsEMBL/Compara/Production/EPOanchors/MapAnchors.pm
```diff
@@ -132,12 +132,10 @@ sub run {
...
 	}
 	my $command = join(" ", $program, $option_st, $query_file, $target_file); 
-	print $command, "\n";
-	my $out_fh;
-	open( $out_fh, '-|', $command ) or die("Error opening exonerate command: $? $!"); #run mapping program
-	$self->param('out_file', $out_fh);
-
         my $hits;
+	$self->read_from_command($command, sub {
+	my $out_fh = shift;;
```
carlacummins:
```suggestion
	my $out_fh = shift;
```

modules/Bio/EnsEMBL/Compara/RunnableDB/DumpMemberSequencesIntoFasta.pm
```diff
@@ -85,7 +85,7 @@ sub run {
 
     close FASTAFILE;
 
-    my $n_seq_in_file = `grep -c "^>" "$fasta_name"`;
+    my $n_seq_in_file = $self->run_command(['grep', '-c', '^>', $fasta_name])->out;
```
carlacummins: I'm a bit unsure about how `grep` will handle the `'^>'` here - I think we should add double quotes around it to be safe, as it was in the original command

modules/Bio/EnsEMBL/Compara/RunnableDB/DumpMembersIntoFasta.pm
```diff
@@ -121,7 +121,7 @@ sub run {
     $member_set->print_sequences_to_file($fasta_file);
 
     my $n_seq_expected = scalar(@$members);
-    my $n_seq_in_file = `grep -c "^>" "$fasta_file"`;
+    my $n_seq_in_file = $self->run_command(['grep', '-c', '^>', $fasta_file])->out;
```
carlacummins: Comment as above
```suggestion
    my $n_seq_in_file = $self->run_command(['grep', '-c', '"^>"', $fasta_file])->out;
```

modules/Bio/EnsEMBL/Compara/RunnableDB/DumpMemberSequencesIntoFasta.pm
muffato: I will double-check, but with the array form, each element becomes a separate argument as is. I think if we use `'">"'`, `grep` would get `"^>"` as pattern and would be looking for double-quotes in the fasta file

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/StoreTree.pm
```diff
@@ -691,7 +691,7 @@ sub call_hcs_all_trees {
 
 sub examl_exe_decision {
     my $self = shift;
-    my $avx = `grep avx /proc/cpuinfo`;
+    my $avx = $self->run_command(['grep', 'avx', '/proc/cpuinfo'])->out;
```
carlacummins: Why not use `get_command_output` here?

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/StoreTree.pm
```diff
@@ -712,7 +712,7 @@ sub raxml_exe_decision {
     my $self = shift;
     my $no_cores = $self->param('raxml_number_of_cores');
 
-    my $avx = `grep avx /proc/cpuinfo`;
+    my $avx = $self->run_command(['grep', 'avx', '/proc/cpuinfo'])->out;
```
carlacummins: Why not use get_command_output here?

modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromMetadataFactory.pm
```diff
@@ -73,8 +73,8 @@ sub fetch_input {
 
 	# check there are no seq_region changes in the existing species
 	my $list_cmd = "perl $list_genomes_script $metadata_script_options";
-	my $list_run = $self->run_command($list_cmd);
-	my @release_genomes = split( /\s+/, $list_run->out );
+	my $list_out = $self->get_command_output($list_cmd);
+	my @release_genomes = split( /\s+/, $list_out );
```
carlacummins: Doesn't `get_command_output` detect `wantarray` and do this splitting automatically?
```suggestion
	my @release_genomes = $self->get_command_output($list_cmd);
```

modules/Bio/EnsEMBL/Compara/RunnableDB/ProteinTrees/ExaML.pm
```diff
@@ -76,13 +76,13 @@ sub fetch_input {
...
 
-	my @dir = `find $source_dir -name $root_id.binary | xargs ls -t`;
+	my @dir = $self->get_command_output("find $source_dir -name $root_id.binary | xargs ls -t");
 	my @tok = split(/\//,$dir[0]);
 	my $worker_dir = $tok[-2];
-	my @list = `ls -t $source_dir/$worker_dir/ExaML_binaryCheckpoint*`;
-	if ((scalar(@list) > 0) && (!-z $list[0])){
-		chomp($list[0]);
-		$self->param('newest_checkPointFile',$list[0]);
+	my $newest_checkPointFile = $self->run_command("ls -t $source_dir/$worker_dir/ExaML_binaryCheckpoint* | head -n 1")->out;
```
carlacummins:
```suggestion
	my $newest_checkPointFile = $self->get_command_output("ls -t $source_dir/$worker_dir/ExaML_binaryCheckpoint* | head -n 1");
```

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/StoreTree.pm
muffato: Because `get_command_output` sets `die_on_failure`, but `grep` returns 1 when it doesn't find any hit, making it look like a failure. I had two choices:
- `run_command(...)->out`
- `get_command_output(..., {die_on_failure => 0})`
I chose the former because it's shorter

modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromMetadataFactory.pm
muffato: `get_command_output` splits on new-lines (`\n`). Would that work here (the pattern currently is `\s+`) ?

modules/Bio/EnsEMBL/Compara/RunnableDB/ProteinTrees/ExaML.pm
muffato: I'll double-check but I think I've tested this and the command returns a non-zero code when there are no files matching, so `get_command_output` would have died (same as with the AVX check)

modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromMetadataFactory.pm
carlacummins: Yes, I've just checked and splitting on newline is fine for this case.

carlacummins:
All looks good to me now!

modules/Bio/EnsEMBL/Compara/RunnableDB/DumpMemberSequencesIntoFasta.pm
muffato: Confirmed: it doesn't work with the double quotes

modules/Bio/EnsEMBL/Compara/RunnableDB/ProteinTrees/ExaML.pm
muffato: `get_command_output` actually works here (because `use_bash_pipefail` is not set)

PR #120, opened 2019-08-28
Renamed JIRA version from 'Release XX' to 'Ensembl XX'

JAlvarezJarreta:
Ticket [ENSCOMPARASW-2458](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-2458).

+ bugfix: extra `my` declaration in JIRA class
+ deleted old file that is not used by our codebase anymore


muffato:
It's necessary to replace "Release" in the tickets summary, but that can be done if you think the wording is better. If I were to change it, I would suggest "Ensembl <Division> <version>"

JAlvarezJarreta:
> It's necessary to replace "Release" in the tickets summary, but that can be done if you think the wording is better. If I were to change it, I would suggest "Ensembl "

Unless I don't understand what you are suggesting, I think that is what I've done in every `jira_recurrent_tickets.<division>.json` file.

muffato:
Gosh, my message looks completely scrambled. I meant this:
It's _not_ necessary to replace "Release" in the tickets summary, but that can be done if you think the wording is better. But if I were to change it, I would suggest `Ensembl <Division> <version> - XXX` or `Ensembl <version> <Division> - XXX` as a template, whichever is the most natural

muffato:
Can you squash the first and third commits together ? This way the history won't show those lines being changed to something and then changed back to what they were

PR #126, opened 2019-09-18
Support both version nomenclatures

muffato:
I've realised that with #120 deployed, the dashboard and graphs were not displaying anything for e98 (and previous releases), simply because there are indeed no tickets for a version "Ensembl 98".

Since JIRA only filters the version names as text, without checking that they really exist etc, I would propose to filter across both nomenclatures, so that the e98 pages still work.

PS: the new versions are already under ~compara_ensembl/public_html. If you want I can restore the release/99 version to demonstrate they can't fetch anything ?

### DIFF
--- a/scripts/jira_tickets/merged_release_graph.js
+++ b/scripts/jira_tickets/merged_release_graph.js
@@ -95,7 +95,7 @@ function process_tickets(json) {
 
 
 // fetch ticket status from REST API
-var endpoint_ticket_list = 'https://www.ebi.ac.uk/panda/jira/rest/api/2/search/?jql=project=ENSCOMPARASW+AND+fixVersion="Ensembl+__RELEASE__"+AND+component+IN+("Relco+tasks","Production+tasks")+AND+labels+IS+NOT+EMPTY+ORDER+BY+created+ASC,id+ASC&maxResults=100';
+var endpoint_ticket_list = 'https://www.ebi.ac.uk/panda/jira/rest/api/2/search/?jql=project=ENSCOMPARASW+AND+(+fixVersion="Ensembl+__RELEASE__"+OR+fixVersion="Release+__RELEASE__"+)+AND+component+IN+("Relco+tasks","Production+tasks")+AND+labels+IS+NOT+EMPTY+ORDER+BY+created+ASC,id+ASC&maxResults=100';
 var release = $.urlParam("release");
 var pipelines = {};
 $('#progress').text("Loading e" + release + " graph");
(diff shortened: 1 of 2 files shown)


scripts/jira_tickets/merged_release_graph.js
```diff
@@ -95,7 +95,7 @@ function process_tickets(json) {
 
 
 // fetch ticket status from REST API
-var endpoint_ticket_list = 'https://www.ebi.ac.uk/panda/jira/rest/api/2/search/?jql=project=ENSCOMPARASW+AND+fixVersion="Ensembl+__RELEASE__"+AND+component+IN+("Relco+tasks","Production+tasks")+AND+labels+IS+NOT+EMPTY+ORDER+BY+created+ASC,id+ASC&maxResults=100';
+var endpoint_ticket_list = 'https://www.ebi.ac.uk/panda/jira/rest/api/2/search/?jql=project=ENSCOMPARASW+AND+(+fixVersion="Ensembl+__RELEASE__"+OR+fixVersion="Release+__RELEASE__"+)+AND+component+IN+("Relco+tasks","Production+tasks")+AND+labels+IS+NOT+EMPTY+ORDER+BY+created+ASC,id+ASC&maxResults=100';
```
JAlvarezJarreta: Alternatively, you could use the following syntax to make the URL more compact:
`(+fixVersion+IN+("Ensembl+__RELEASE__"+,+"Release+__RELEASE__")+)`
Although it is not that relevant as I don't expect the field `fixVersion` to be changed in the near future.

scripts/jira_tickets/release_dashboard.js
```diff
@@ -13,7 +13,7 @@ String.prototype.capitalize = function() {
 }
 
 // fetch ticket status from REST API
-var endpoint_ticket_list = 'https://www.ebi.ac.uk/panda/jira/rest/api/2/search/?jql=project=ENSCOMPARASW+AND+issuetype=Task+AND+fixVersion="Ensembl+__RELEASE__"+AND+component="Relco+tasks"+AND+cf[11130]=__DIVISION__+ORDER+BY+created+ASC,id+ASC&maxResults=500';
+var endpoint_ticket_list = 'https://www.ebi.ac.uk/panda/jira/rest/api/2/search/?jql=project=ENSCOMPARASW+AND+issuetype=Task+AND+(+fixVersion="Ensembl+__RELEASE__"+OR+fixVersion="Release+__RELEASE__"+)+AND+component="Relco+tasks"+AND+cf[11130]=__DIVISION__+ORDER+BY+created+ASC,id+ASC&maxResults=500';
```
JAlvarezJarreta: Same as above.

JAlvarezJarreta:
This also raises the problem that the JIRA module will allow the creation of new tickets for an old release with the same summary as an existing ticket since the `fixVersion` field will be different. But I guess the chances of this scenario as almost null.

muffato:
Good idea, I prefer the `IN`. Will change that now

muffato:
It's unlikely to happen indeed, but still useful to know

PR #125, opened 2019-09-16
Improved the detection of genome reusability

muffato:
It all started when we found out that ProteinTrees/CheckGenomedbReusability.pm was only comparing some properties of _protein-coding_ genes only to assess the reusabiity of genomes. I have fixed this and the query now includes (almost) all the fields that we use to make members.
Then I have rearranged the modules into a parent class and specific sub-modules for each pipeline (LoadMembers, ProteinTrees, EPO) so that they only test what matters.

The previous code caused two data bugs:
1. By only considering protein-coding genes, it was blind to updates to ncRNAs. Since some species have had their ncRNA-set updated recently, our current database is out of sync as it still includes the previous set of ncRNAs ... In e99, we will likely have to mark those species as "do_not_reuse" (although I'm also considering whether the e98 database could be patched instead).
2. The other issue is that our seq_member stable_ids for yeast _don't exist_. We still have the stable_ids that were in the e96 core db, but they have been renamed in the e97 core database ...

PR #128, opened 2019-09-28
Force fresh reload of non-coding genes

muffato:
Follow-up of #125. Since the non-coding genes were ignored when assessing the reusability of genomes, we have reused gene-sets that have in reality changed and are out of sync with many core databases: 34 species have a different gene-set, 36 more species have the same gene-set but with different biotypes.

Note: this is only needed for Vertebrates because we don't run any pipelines on non-coding genes for Plants

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/DBSQL/GeneMemberAdaptor.pm
+++ b/modules/Bio/EnsEMBL/Compara/DBSQL/GeneMemberAdaptor.pm
@@ -254,6 +254,16 @@ sub store {
 
 }
 
+sub delete {
+    my ($self, $gene_member) = @_;
+    foreach my $seq_member (@{$gene_member->get_all_SeqMembers}) {
+        $seq_member->adaptor->delete($seq_member);
+    }
+    $self->dbc->do('DELETE FROM gene_member_qc          WHERE gene_member_stable_id = ?', undef, $gene_member->stable_id);
+    $self->dbc->do('DELETE FROM member_xref             WHERE gene_member_id = ?', undef, $gene_member->dbID);
+    $self->dbc->do('DELETE FROM gene_member_hom_stats   WHERE gene_member_id = ?', undef, $gene_member->dbID);
+    $self->dbc->do('DELETE FROM gene_member             WHERE gene_member_id = ?', undef, $gene_member->dbID);
+}
 
 1;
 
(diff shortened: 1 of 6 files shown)


modules/Bio/EnsEMBL/Compara/DBSQL/SeqMemberAdaptor.pm
```diff
@@ -452,6 +452,18 @@ sub _set_member_as_canonical {
...
+    $self->dbc->do('DELETE FROM seq_member_projection_stable_id WHERE target_seq_member_id = ?', undef, $seq_member->dbID);
+    $self->dbc->do('DELETE FROM seq_member_projection           WHERE source_seq_member_id = ?', undef, $seq_member->dbID);
+    $self->dbc->do('DELETE FROM seq_member_projection           WHERE target_seq_member_id = ?', undef, $seq_member->dbID);
+    $self->dbc->do('DELETE FROM gene_member_qc          WHERE seq_member_id = ?', undef, $seq_member->dbID);
+    $self->dbc->do('DELETE FROM hmm_annot               WHERE seq_member_id = ?', undef, $seq_member->dbID);
+    $self->dbc->do('DELETE FROM exon_boundaries         WHERE seq_member_id = ?', undef, $seq_member->dbID);
+    $self->dbc->do('DELETE FROM other_member_sequence   WHERE seq_member_id = ?', undef, $seq_member->dbID);
+    $self->dbc->do('DELETE FROM seq_member              WHERE seq_member_id = ?', undef, $seq_member->dbID);
+}
+
```
carlacummins: What about `homology_member`?

modules/Bio/EnsEMBL/Compara/DBSQL/GeneMemberAdaptor.pm
```diff
@@ -254,6 +254,16 @@ sub store {
...
 }
 
+sub delete {
+    my ($self, $gene_member) = @_;
+    foreach my $seq_member (@{$gene_member->get_all_SeqMembers}) {
+        $seq_member->adaptor->delete($seq_member);
+    }
+    $self->dbc->do('DELETE FROM gene_member_qc          WHERE gene_member_stable_id = ?', undef, $gene_member->stable_id);
+    $self->dbc->do('DELETE FROM member_xref             WHERE gene_member_id = ?', undef, $gene_member->dbID);
+    $self->dbc->do('DELETE FROM gene_member_hom_stats   WHERE gene_member_id = ?', undef, $gene_member->dbID);
```
JAlvarezJarreta: Shouldn't `ortholog_goc_metric` be also targeted?

muffato: Yes it should. But also, since @carlacummins rewrote the GOC analyses this table is not populated any more. I would actually suggest to remove it from our production schema

PR #135, opened 2019-10-15
Added script to generate JIRA tickets from failed datachecks

JAlvarezJarreta:
It is based on `scripts/jira_tickets/create_healthcheck_tickets.pl` (basically I have replaced the method `parse_healthchecks` by `parse_datachecks`).


JAlvarezJarreta:
> Could you make it more generic (dynamic project name - standard fields) and submit it to datacheck repo @JAlvarezJarreta ?

I'm going to need an example of a datacheck TAP file with failures from another group/database than Compara to make sure the script works before submitting a PR.

PR #137, opened 2019-10-16
Allow using a more fitting term than strain, e.g. breed

muffato:
See https://www.ebi.ac.uk/panda/jira/browse/ENSINT-317 for more information.

Here I change the content of `genome_db.strain_name` to explicitly include "strain", "breed", etc.
There will be some SQL to apply to the vertebrates and plants databases (e99 and master), see https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-2532, following the staging-patch https://github.com/Ensembl/staging-patches/pull/325/files

Note that `strain.type` is currently optional but will at some point be made mandatory via a DataCheck. Then we should be able to remove default value `|| 'strain'`


muffato:
Not just wheat, but all the other species too, then

muffato:
Actually, I shouldn't add a key to the wheat database since the key is optional and Plants themselves haven't populated it yet. It is (will) only be present for some Vertebrates

PR #140, opened 2019-10-21
Improved and reenabled the multiple-alignment statistics

muffato:
# Use case

cf https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-1746. The stats will be used to compare our alignments to Cactus, but we are unable to compute them at the moment because they are too slow. They include:
1. the "depth", i.e. the number of sequences aligned to a given sequence,
2. the pairwise coverage.

# Description

In this PR I have reimplemented all the modules from scratch. I have tested the new stats module on all e98 multiple alignments. The new code can compute the stats for everything in about 950 hours on a single worker. But it works fine with 500 workers in parallel too :relieved:, so speed is not an issue any more.
All the important methods are in `Utils::Cigars` and the Runnables merely fetching the cigar-lines, calling those methods and aggregating / dataflowing the results. All have unit-tests.
I have also changed a few things regarding the handling of cigar-lines in other pipelines / modules, consolidating the methods in Utils::Cigars, e.g.:
- making sure the EPO pipelines don't generate `0X` elements in the cigar-lines
- making sure 0* elements are ignored when parsing cigar-lines
- extending `Utils::Cigars` to understand `G` and `X` elements
- changing the interface to return references instead of objects
- removing methods from AlignedMember and GenomicAlign in order to explicitly call `Utils::Cigars`

Note: I might be able to take these changes out to a separate PR if there is too much in this one.

# Implementation

The most important new method is `column_iterator`. It is able to parse a multiple alignment (given as cigar lines / cigar arrays) to extract consecutive columns that share the same pattern of match / gaps. The processing of the columns is deferred to a callback subroutine. This method is not only used in the two new stats but also in two other functions of `Utils::Cigars` that it significantly speeds up.

In the Stats pipeline I have unified both Runnables into one, so that the data (cigar lines) are only fetched and transformed into arrays once. Both sets of stats being implemented in two different methods of `Utils::Cigars`, column_iterator is actually called twice. And since both callbacks are quite similar too, I think the runtime could be significantly improved (halved ??) by merging both functions into one. I haven't done it yet because I felt they don't belong together in Utils::Cigars. Maybe I should define a callback and call `column_iterator` in the Runnable itself to make things more efficient ? I would still leave the methods in Utils::Cigars, though.

Each alignment block can be processed in parallel and will have its stats written to eHive's accu. For the largest alignment (mammals _EPO_LOW_COVERAGE_) that's on average >15,000 stats per block. In my early tests I noticed that the accu table was the bottleneck as the jobs are very quick to complete. Since the stats are aggregated simply being summed up, I have made each job process several blocks, do the partial sums, and send these to the accu tables. 20 blocks / job seems to work well for _EPO_ alignments, but I had to increase this to 50 for _EPO_LOW_COVERAGE_ alignments. The longest _EPO_ job only takes a few minutes, but the longest _EPO_LOW_COVERAGE_ job takes up to 3 hours. That's what allows running more in parallel, though.


JAlvarezJarreta:
> In the Stats pipeline I have unified both Runnables into one, so that the data (cigar lines) are only fetched and transformed into arrays once. Both sets of stats being implemented in two different methods of Utils::Cigars, column_iterator is actually called twice. And since both callbacks are quite similar too, I think the runtime could be significantly improved (halved ??) by merging both functions into one. I haven't done it yet because I felt they don't belong together in Utils::Cigars. Maybe I should define a callback and call column_iterator in the Runnable itself to make things more efficient ? I would still leave the methods in Utils::Cigars, though.
Unless we struggle a lot with the performance (and based on what you have said that's unlikely for now), I would say I agree with you that it makes more sense to have them separated. There is some common code, but making a merge that would keep both functions separated wouldn't be straightforward and I don't think it is worth it at this point.

modules/Bio/EnsEMBL/Compara/PipeConfig/MultipleAlignerStats_conf.pm
```diff
@@ -107,7 +108,7 @@ sub pipeline_wide_parameters {
 
 
 
-sub pipeline_analyses {
+sub core_pipeline_analyses {
```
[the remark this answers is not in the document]
carlacummins: It's in the ComparaGeneric_conf

modules/Bio/EnsEMBL/Compara/RunnableDB/EpoLowCoverage/LowCoverageGenomeAlignment.pm
```diff
@@ -412,7 +413,11 @@ sub _parse_results {
 			#print "before cigar_line " . $genomic_align->cigar_line . "\n";
 
 			
-			$genomic_align->cigar_line($start_X . "X" .$genomic_align->cigar_line . $end_X . "X");
+			$genomic_align->cigar_line(
+			    Bio::EnsEMBL::Compara::Utils::Cigars::_cigar_element('X', $start_X) .
```
carlacummins: I find it unusual for a method to be called like this - my understanding of `_*` methods is that they are internal to the module and would not be called outside of that module. This doesn't follow that convention. Am I too strict with my definition of internal methods?

modules/Bio/EnsEMBL/Compara/Utils/Cigars.pm
```diff
@@ -100,14 +102,19 @@ sub compose_sequence_with_cigar {
     throw("Invalid cigar_line '$cigar_line'\n") if $cigar_line !~ /^[0-9A-Z]*$/;
```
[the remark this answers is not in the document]
muffato: OK, but with a question mark instead of the second `*` ?
```suggestion
     throw("Invalid cigar_line '$cigar_line'\n") if $cigar_line !~ /^(([1-9][0-9]*)?[A-Z])*$/;
```

modules/Bio/EnsEMBL/Compara/Utils/Cigars.pm
```diff
@@ -182,6 +189,7 @@ sub expand_cigar {
     my $expanded_cigar = '';
     #$cigar =~ s/(\d*)([A-Z])/$2 x ($1||1)/ge; #Expand
     while ($cigar =~ /(\d*)([A-Za-z])/g) {
+        next if $1 =~ /^0+$/;
```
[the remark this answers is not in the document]
muffato: you mean if I introduce a `throw("Invalid cigar line")`

modules/Bio/EnsEMBL/Compara/Utils/Cigars.pm
```diff
@@ -634,4 +656,240 @@ sub check_cigar_line {
...
+            }
+        }
+        # "- 1" because the depth is the number of *other* sequences
+        my $this_depth = scalar(keys %n_aligned_ids) - 1;
+
+        # Update the counters
+        foreach my $id (keys %n_aligned_ids) {
+            my $n_pos = $n_aligned_ids{$id} * $length;
+            $depth_breakdown{$id}->{$this_depth} += $n_pos;
+            $n_total_pos{$id} += $n_pos;
```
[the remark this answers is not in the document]
muffato: Yes, that's Perl's autovivification

modules/Bio/EnsEMBL/Compara/Utils/Cigars.pm
```diff
@@ -634,4 +656,240 @@ sub check_cigar_line {
...
+                            throw("Not all the cigars have the same length\n");
+                        } elsif ($curr_cigar_elem_index[$j] != $length_cigar_line_array[$j]-1) {
+                            throw("Not all the cigars have the same length\n");
+                        }
+                    }
+                    return;
+                }
+                my $e = $cigar_lines_arrays[$i]->[ $curr_cigar_elem_index[$i] ];
+                $curr_cigar_elem_codes[$i]    = $e->[0];
+                $curr_cigar_elem_lengths[$i]  = $e->[1];
```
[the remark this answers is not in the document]
muffato: I agree about the `else`, but I would keep the `return`. There are two loops to escape from: a `for` and a `while`. To use `last` I would need to introduce a label, I think

modules/Bio/EnsEMBL/Compara/RunnableDB/ProteinTrees/AlignmentFilteringTagging.pm
```diff
@@ -190,87 +163,17 @@ sub _get_gappiness {
...
-    foreach my $member ( keys %sum ) {
-        my $seq_avg = $sum{$member}/$seq_num_of_aligned_positions[$member];
-        #print $sum{$member} . "/" . $seq_num_of_aligned_positions[$member] . " = $seq_avg\n" if ($self->debug);
-        $total += $seq_avg;
+    my $s = 0;
+    my $n = 0;
+    foreach my $val (values %$total_depths) {
+        $s += $val->{'depth_sum'} / $val->{'n_total_pos'};
+        $n++;
     }
```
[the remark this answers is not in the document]
muffato: Yes it'd be the same

modules/Bio/EnsEMBL/Compara/RunnableDB/EpoLowCoverage/LowCoverageGenomeAlignment.pm
muffato: I have factored this pattern out to a new method in Utils::Cigars

modules/Bio/EnsEMBL/Compara/Utils/Cigars.pm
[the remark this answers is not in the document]
JAlvarezJarreta: Yes, this way we catch earlier cigar lines that include zeros (when they shouldn't).

modules/Bio/EnsEMBL/Compara/RunnableDB/Ortheus.pm
```diff
@@ -607,7 +608,7 @@ sub parse_results {
 
 			print "start_X $start_X end_X $end_X subseq_length " . length($subseq) . "\n" if ($self->debug);
 
-			$genomic_align->cigar_line($start_X . "X" .$genomic_align->cigar_line . $end_X . "X");
+			$genomic_align->cigar_line( Bio::EnsEMBL::Compara::Utils::Cigars::pad_with_x($genomic_align->cigar_line, $start_X, $end_X) );
```
JAlvarezJarreta:
```suggestion
                        $genomic_align->cigar_line( Bio::EnsEMBL::Compara::Utils::Cigars::pad_with_x($genomic_align->cigar_line, $start_X, $end_X) );
```
I think I have replaced the tabs by the right number of spaces.

modules/Bio/EnsEMBL/Compara/Utils/Cigars.pm
```diff
@@ -97,17 +99,22 @@ sub compose_sequence_with_cigar {
     my $alignment_string = "";
     my $seq_start = 0;
 
-    throw("Invalid cigar_line '$cigar_line'\n") if $cigar_line !~ /^[0-9A-Z]*$/;
+     throw("Invalid cigar_line '$cigar_line'\n") if $cigar_line !~ /^(([1-9][0-9]*)?[A-Z])*$/;
```
JAlvarezJarreta: I added an extra white space by mistake...

modules/Bio/EnsEMBL/Compara/RunnableDB/ProteinTrees/AlignmentFilteringTagging.pm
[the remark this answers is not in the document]
JAlvarezJarreta: My point is that I don't see the need to do an incremental update of `$n` on every iteration when you can extract its value the other way. Is there a performance aspect that I'm missing?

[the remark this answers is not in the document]
muffato: I don't think it matters. I find the current code clear enough. I actually have concerns about your suggestion. Does `my $n = keys %$total_depths;` make Perl traverse the whole hash and instantiate a new list with all the keys (are those keys -strings- copied or will it reuse the references ?) just to take the length and trash the array ?

modules/Bio/EnsEMBL/Compara/RunnableDB/Ortheus.pm
muffato: I would actually leave the tabs as they are as they fit with the rest of the function. But I would change the other occurrence in Ortheus.pm

muffato: I'm also going to restore the indentation in RunnableDB/EpoLowCoverage/LowCoverageGenomeAlignment.pm

modules/Bio/EnsEMBL/Compara/RunnableDB/GenomicAlignBlock/CalculateBlockStats.pm
```diff
@@ -0,0 +1,142 @@
...
+
+    my $depth_by_genome = $self->param('depth_by_genome');
+
+    # Convert the cigar strings to arrays beforehand and just once
+    my @all_cigar_arrays;
+    my @all_genome_db_ids;
+    while(my $ga = shift @$genomic_aligns) {
+        my $cigar_array  = Bio::EnsEMBL::Compara::Utils::Cigars::get_cigar_array($ga->cigar_line);
+        my $genome_db_id = $ga->dnafrag->genome_db_id;
+        my $i_length     = sum(map {$_->[1]} grep {$_->[0] eq 'I'} @$cigar_array);
```
carlacummins: Why are `I`s being handled here?

muffato: Right. I will add a comment block to explain this.

It's an optimisation. `I`s represent bits that are not aligned to anything, so their depth is automatically 0 and they don't add anything to the coverage. They are handled (correctly) by `column_iterator` but it's relatively expensive as it has to mimic a column of `D`s and `M`s and then call the function that will compute the depth / coverage. It is faster to count them and splice them out here

modules/Bio/EnsEMBL/Compara/Utils/Cigars.pm
```diff
@@ -634,4 +706,241 @@ sub check_cigar_line {
...
+        # Will contain the group_ids that are present on these columns
+        # and the number of sequences they contain
+        my %n_aligned_ids;
+        for (my $i = 0; $i < $n_cigars; $i++ ) {
+            if ( $codes->[$i] eq 'M' ) {
+                $n_aligned_ids{ $group_ids->[$i] } ++;
+            }
+        }
+        # "- 1" because the depth is the number of *other* sequences
+        my $this_depth = scalar(keys %n_aligned_ids) - 1;
```
carlacummins: Should it be other sequences (as the comment suggests) or other groups (as the code suggests)?

modules/Bio/EnsEMBL/Compara/Utils/Cigars.pm
```diff
@@ -226,9 +252,11 @@ sub collapse_cigar {
 
 sub alignment_length_from_cigar {
     my $cigar = shift;
+
+    assert_valid_cigar($cigar);
+
     my $length = 0;
     while ($cigar =~ /(\d*)([A-Za-z])/g) {
```
JAlvarezJarreta: This should be change as the others:
```suggestion
    while ($cigar =~ /(\d*)([A-Z])/g) {
```

modules/Bio/EnsEMBL/Compara/Utils/Cigars.pm
```diff
@@ -179,6 +209,9 @@ sub cigar_from_alignment_string {
 
 sub expand_cigar {
     my $cigar = shift;
+
+    assert_valid_cigar($cigar);
+
     my $expanded_cigar = '';
     #$cigar =~ s/(\d*)([A-Z])/$2 x ($1||1)/ge; #Expand
     while ($cigar =~ /(\d*)([A-Za-z])/g) {
```
JAlvarezJarreta: This should be change as the others:
```suggestion
    while ($cigar =~ /(\d*)([A-Z])/g) {
```

muffato: This should be change as the others:
```suggestion
     while ($cigar =~ /(\d*)([A-Z])/g) {
```

modules/Bio/EnsEMBL/Compara/Utils/Cigars.pm
muffato: Fixing the whitespace
```suggestion
     while ($cigar =~ /(\d*)([A-Z])/g) {
```

modules/Bio/EnsEMBL/Compara/Utils/Cigars.pm
```diff
@@ -72,9 +72,35 @@ use strict;
...
 use Bio::EnsEMBL::Utils::Exception qw(throw);
 
 
+=head2 assert_valid_cigar
+
+  Arg [1]     : String $cigar_line
+  Example     : assert_valid_cigar('3M4D');  # will return
+  Example     : assert_valid_cigar('3M0X');  # will throw an exception
+  Description : Tests the validity of a cigar line, which is:
+                 1) is a sequence of digits and letters
```
muffato:
```suggestion
                 1) is a sequence of digits and upper-case letters
```

modules/Bio/EnsEMBL/Compara/Utils/Cigars.pm
muffato: Other groups indeed

muffato:
Thank you ! :sweat_smile:

PR #144, opened 2019-10-31
Fix/speciestree

CristiGuijarro:
The species tree worked perfectly for the large default species_sets such as vertebrates but was not set up to work for smaller species_sets such as 10 rice species. There were complications with the outgroup and species_set parameters conflicting with the collection parameters, now only the outgroup species name is needed, the genome_db_id will be collected through the genome_db adaptor. The root is also now decided based on species_set rather than forced to be the tree of life ancestral root. The un-rooting step worked well for the large trees, but was unable to unroot the smaller trees without a forced midpoint rooting first.


carlacummins:
I think this work should go to master, not release/99 - can you make a new PR?

PR #151, opened 2019-10-31
Removed EBI layer on Homologies-related pipelines

JAlvarezJarreta:
Ticket [ENSCOMPARASW-2486](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-2486) (part 6/7).


modules/Bio/EnsEMBL/Compara/PipeConfig/Vertebrates/OrthologQM_Alignment_conf.pm
```diff
@@ -0,0 +1,80 @@
...
+
+=head1 SYNOPSIS
+
+    For Vertebrates:
+        init_pipeline.pl Bio::EnsEMBL::Compara::PipeConfig::Vertebrates::OrthologQM_Alignment_conf -host mysql-ens-compara-prod-X -port XXXX \
+            -member_type <protein_or_ncrna>
+
+    For a specific strain/breed:
+        init_pipeline.pl Bio::EnsEMBL::Compara::PipeConfig::Vertebrates::OrthologQM_Alignment_conf -host mysql-ens-compara-prod-X -port XXXX \
+            -member_type <protein_or_ncrna> -collection <murinae_or_sus>
```
carlacummins: Also need to include the `compara_db`, since the default is to use the `member_type` to point to either `compara_ptrees` or `compara_nctrees`, rather than `sus_ptrees`, for example.

modules/Bio/EnsEMBL/Compara/PipeConfig/Plants/Synteny_conf.pm
```diff
@@ -51,11 +54,11 @@ sub default_options {
     my ($self) = @_;
     return {
         %{$self->SUPER::default_options},   # inherit the generic ones
-
-        'master_db' => 'compara_master',
-        'work_dir'  => $self->o('pipeline_dir'),
+        'division'     => 'plants',
+        
+        'alignment_db'    => 'compara_curr',
+        'curr_release_db' => 'compara_curr',
```
carlacummins: These seem like the kind of thing you would have moved to the base class in other cases (since the setting is the same for verts and plants). Any reason this is different?

carlacummins:
Github reporting conflicts here too

modules/Bio/EnsEMBL/Compara/PipeConfig/Vertebrates/StrainsOrthologQM_Alignment_conf.pm
```diff
@@ -0,0 +1,56 @@
...
+
+use base ('Bio::EnsEMBL::Compara::PipeConfig::Vertebrates::OrthologQM_Alignment_conf');
+
+
+sub default_options {
+    my ($self) = @_;
+    return {
+        %{$self->SUPER::default_options},   # inherit the generic ones
+
+        'collection'  => undef, # undef to force to be passed as an argument
```
muffato: This doesn't force anything. There will be Perl warnings such as
> Use of uninitialized value in substitution iterator at /home/matthieu/workspace/src/hive/master/modules/Bio/EnsEMBL/Hive/DependentOptions.pm line 200.
and the database will be named like `matthieu_vertebrates__protein_orth_qm_wga_99`

To really force something you can:
- either edit `$self->SUPER::default_options` to remove `collection`, but we never do this pattern
- or introduce an alias like `'collection' => $self->o('collection_name'),`

modules/Bio/EnsEMBL/Compara/PipeConfig/Vertebrates/StrainsOrthologQM_Alignment_conf.pm
```diff
@@ -0,0 +1,56 @@
...
+        
+        'compara_db'  => '#expr( (#member_type# eq "protein") ? "#collection#_ptrees" : "#collection#_nctrees" )expr#',
```
muffato: `collection` must be put in pipeline_wide_parameters for this to work

PR #147, opened 2019-10-31
Removed EBI layer on PWA-related pipelines

JAlvarezJarreta:
Ticket [ENSCOMPARASW-2486](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-2486) (part 2/7).

PR #154, opened 2019-11-04
Removed EBI layer on newly added Metazoa pipelines

JAlvarezJarreta:
Ticket ENSCOMPARASW-2486.


JAlvarezJarreta:
The last commit has to do with the last changes in [PR#151](https://github.com/Ensembl/ensembl-compara/pull/151).

JAlvarezJarreta:
> Same as in #147: use `tweak_analyses` rather than `pipeline_analyses` in their LastZ
> 
> Also,
> 
> ```perl
>         'alignment_db'    => 'compara_curr',
>         'curr_release_db' => 'compara_curr',
> ```
> 
> should be in the main Synteny_conf since it's the default way we want it to run

The last commit has to do with the last changes in [PR#151](https://github.com/Ensembl/ensembl-compara/pull/151).

PR #156, opened 2019-11-11
Calculate ortholog_quality for polyploid genomes

CristiGuijarro:
The pipeline was running on both the components and principal genomes, this meant that the data were not correctly collaborating. Only protein tree data is stored for the components and not on the principal genome db id for the polyploids, and the rest of the alignment data and mlsses are only available on the principal. The fix takes the genomic alignment data from the principal to calculate the scores for the components. Now the ortholog_quality writes output for the components and not the principal genome db ids.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/CalculateWGACoverage.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/CalculateWGACoverage.pm
@@ -122,7 +122,9 @@ sub fetch_input {
 =cut
 
 sub run {
-	my $self = shift;	
+	my $self = shift;
+	my $dba  = $self->get_cached_compara_dba('pipeline_url');
+	my $gdba = $dba->get_GenomeDBAdaptor;
 
 	my @orth_batch = @{ $self->param_required('orth_batch') };
 	my %aln_ranges = %{ $self->param_required('aln_ranges') };
@@ -130,8 +132,6 @@ sub run {
 	my (@qual_summary, @orth_ids);
 	foreach my $orth ( @orth_batch ) {
 		my @orth_dnafrags = @{ $orth->{ 'orth_dnafrags'} };
-		my @aln_mlss_ids  = @{ $self->param_required( 'aln_mlss_ids' ) };
-
 		my $orth_ranges    = $orth->{ 'orth_ranges' };
 		my $homo_id        = $orth->{'id'};
 		my $this_aln_range = $aln_ranges{ $homo_id  };
@@ -144,7 +144,11 @@ sub run {
 		if ( defined $exon_ranges ){
 			foreach my $aln_mlss ( keys %{ $this_aln_range } ){
 				foreach my $gdb_id ( sort {$a <=> $b} keys %{ $orth_ranges } ){
-					my $combined_coverage = $self->_combined_coverage( $orth_ranges->{$gdb_id}, $this_aln_range->{$aln_mlss}->{$gdb_id}, $exon_ranges->{$gdb_id} );
+					# Make sure that if this is a component gdb_id it refers back to the principal. The exon and ortholog data is on the components and not on the principal, but the aln_mlss is only on principal
+					my $gdb = $gdba->fetch_by_dbID($gdb_id);
+					my $principal_gdb = $gdb->principal_genome_db;
+					my $principal_gdb_id = $principal_gdb ? $principal_gdb->dbID : $gdb_id;
+					my $combined_coverage = $self->_combined_coverage( $orth_ranges->{$gdb_id}, $this_aln_range->{$aln_mlss}->{$principal_gdb_id}, $exon_ranges->{$gdb_id} );
 					push( @qual_summary, 
 						{ homology_id              => $homo_id, 
 						  genome_db_id             => $gdb_id,


modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/CalculateWGACoverage.pm
```diff
@@ -122,16 +122,16 @@ sub fetch_input {
...
+	my $dba = $self->get_cached_compara_dba('pipeline_url');
+	my $gdba = $dba->get_GenomeDBAdaptor;
 
 	my @orth_batch = @{ $self->param_required('orth_batch') };
 	my %aln_ranges = %{ $self->param_required('aln_ranges') };
 
 	my (@qual_summary, @orth_ids);
 	foreach my $orth ( @orth_batch ) {
 		my @orth_dnafrags = @{ $orth->{ 'orth_dnafrags'} };
-		my @aln_mlss_ids  = @{ $self->param_required( 'aln_mlss_ids' ) };
```
CristiGuijarro: This array is redundant, so I removed it

modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/CalculateWGACoverage.pm
```diff
@@ -144,7 +144,12 @@ sub run {
 		if ( defined $exon_ranges ){
 			foreach my $aln_mlss ( keys %{ $this_aln_range } ){
 				foreach my $gdb_id ( sort {$a <=> $b} keys %{ $orth_ranges } ){
-					my $combined_coverage = $self->_combined_coverage( $orth_ranges->{$gdb_id}, $this_aln_range->{$aln_mlss}->{$gdb_id}, $exon_ranges->{$gdb_id} );
+					# Make sure that if this is a component gdb_id it refers back to the principal. The exon data is all on the components and not on the principal, but the ortholog info and aln_mlss is only on principal
+					my $gdb = $gdba->fetch_by_dbID($gdb_id);
+					my $components = $gdb->component_genome_dbs($gdb_id);
```
JAlvarezJarreta:
```suggestion
```
You don't seem to be using this for anything.

CristiGuijarro: Yes, I meant to remove that, thanks

modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/CalculateWGACoverage.pm
```diff
@@ -122,16 +122,16 @@ sub fetch_input {
 =cut
 
 sub run {
-	my $self = shift;	
+	my $self = shift;
+	my $dba = $self->get_cached_compara_dba('pipeline_url');
```
JAlvarezJarreta:
```suggestion
	my $dba  = $self->get_cached_compara_dba('pipeline_url');
```

modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/CalculateWGACoverage.pm
```diff
@@ -144,7 +144,11 @@ sub run {
 		if ( defined $exon_ranges ){
 			foreach my $aln_mlss ( keys %{ $this_aln_range } ){
 				foreach my $gdb_id ( sort {$a <=> $b} keys %{ $orth_ranges } ){
-					my $combined_coverage = $self->_combined_coverage( $orth_ranges->{$gdb_id}, $this_aln_range->{$aln_mlss}->{$gdb_id}, $exon_ranges->{$gdb_id} );
+					# Make sure that if this is a component gdb_id it refers back to the principal. The exon data is all on the components and not on the principal, but the ortholog info and aln_mlss is only on principal
+					my $gdb = $gdba->fetch_by_dbID($gdb_id);
+					my $principal_gdb = $gdb->principal_genome_db;
+					my $principal_gdb_id = $principal_gdb ? $principal_gdb->dbID : $gdb_id;
+					my $combined_coverage = $self->_combined_coverage( $orth_ranges->{$gdb_id}, $this_aln_range->{$aln_mlss}->{$principal_gdb_id}, $exon_ranges->{$gdb_id} );
```
carlacummins: You said in the comment (line 147) that ortholog info is on the principal, so shouldn't this be
```suggestion
					my $combined_coverage = $self->_combined_coverage( $orth_ranges->{$principal_gdb_id}, $this_aln_range->{$aln_mlss}->{$principal_gdb_id}, $exon_ranges->{$gdb_id} );
```
?

CristiGuijarro: I need to alter the comment, I counted the ortholog information under protein trees

muffato:
It took me a while to understand what is going on. In the end, what helped me is @carlacummins 's explanation at the group meeting. The homology dumps include the genome_db_id of each member, which are still the component ones when the dumps are created, and these genome_dbs_ids are here the keys of `$orth_ranges`. On the other hand, `$this_aln_range->{$aln_mlss}` has the genome_db_ids of the alignment (line 106), which are the principal ones.
So the bugfix is to map the component genome_db_ids of the orthologs to the principal genome_db_ids so that they can match. However I think your change would fit better in `prepare_orthologs` because the latter is supposed to hash the orthologs info the right way for `calculate_wga_coverage`.

In terms of timing and synchronisation:
1. `prepare_orthologs` queries the `gene_member` and `exon_boundaries` tables by genome_db_id. It needs to have the component genome_db_ids if run _before_ the members are moved, and the principal genome_db_ids if run _after_ the members are moved. In the pipeline, these genome_db_ids come from `ortholog_mlss_factory` which filters the list of all possible pairs of genome_db_ids (incl. principal _and_ component ones) down to the pairs that have an ENSEMBL_ORTHOLOGUES mlss. Since the latter are only created for the principal genome_db_ids (see Utils::MasterDatabase::create_homology_mlsss), this means that currently the pipeline must run _after_ the members have been moved back to the principal genomes in order to be able to get exon boundaries. If we want to be able to run the pipeline before the members are moved, we could for instance make the fetching of exons expand the list of genome_db_ids to the component ones.
2. `prepare_orthologs` gets the dnafrag_id of each member, and passes the information to `calculate_wga_coverage` to query the alignment database (line 94). Since the alignments are on the principal genome_dbs too, here again this means that the orthologs must have been moved to the principal genome _before_ the pipeline is run. If we want to be able to run the pipeline before the members are moved, we could for instance make `prepare_orthologs` convert the dnafrag_ids when in presence of component ones.

I don't think we need to make these changes, but we need to document the dependency very clearly. The WGA pipeline can start:
* for Vertebrates: when `homology_id_mapping` is done 
* for Plants: when `move_back_component_genes` is done

On a related note, `ortholog_mlss_factory` is currently not able to select ENSEMBL_HOMEOELOGUES mlss_ids. I will create another ticket for that, that requires self-alignments of polyploids to be sorted out. That would require moving away from using genome_db_ids to differentiate between the members in the input of `calculate_wga_coverage`

carlacummins:
@muffato : Currently, `prepare_orthologs` is reading orthologs from the homology dumps and there is no step to alter the dump files to move things back to components, so as it stands, the dependency for plants to wait for `move_back_component_genes` is not exactly correct either. Given what you've said above, we can either make the changes you've suggested __or__ make `move_back_component_genes` update the genome_db_ids in the dump files.

muffato:
Why is not correct to say that the WGA pipeline requires `move_back_component_genes` ? (for Plants)

muffato:
I was only exposing what I think would be needed if we want to be able to start the WGA pipeline earlier, i.e. as soon as `homology_id_mapping` is done, before `move_back_component_genes`. But I don't think we need to. The pipeline is quick enough and we can easily manage that dependency. You may face that, though, in https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-2024 but that's another story

carlacummins:
> Why is not correct to say that the WGA pipeline requires `move_back_component_genes` ?

Because WGA now uses the homology dumps to fetch orthologs and `move_back_component_genes` doesn't touch the files, so they will still contain component-level data afterwards

muffato:
We talked with Carla and decided to file two more Jira tickets about the handling of polyploids:
- Make the pipeline work on orthologies that are still on the component genome_db_id/dnafrag_ids, to allow running the pipeline before `move_back_component_genes`
- Make the pipeline work on homeoelogues too
And one (not much related to this PR) about running WGA off the production alignment databases directly. We believe some changes that have been made to the code alleviate the need to copy all the alignments to the local database. This would remove a serious overhead and would simplify embedding the WGA pipeline in the protein-trees pipeline

PR #158, opened 2019-11-13
homology_id INT to BIGINT

CristiGuijarro:
At 32-bits the number of homology_ids that could populate the database was limited to around 370 genomes. Given we expect thousands more genomes to come this way, it makes sense to update to 64-bits (INT->BIGINT) for the homology_id column, providing more space to grow. This patch and schema change changes the homology_id from INT to BIGINT.


sql/patch_99_100_b.sql
```diff
@@ -0,0 +1,31 @@
...
+#   room for another 130 genomes (a total of 370 in vertebrates). 64-bits
+#   means potential for another 100,000 genomes. 
+
+
+ALTER TABLE homology MODIFY homology_id bigint(20) UNSIGNED NOT NULL AUTO_INCREMENT;
+ALTER TABLE homology_member MODIFY homology_id bigint(20) UNSIGNED NOT NULL;
+
+# Patch identifier
+INSERT INTO meta (species_id, meta_key, meta_value)
+  VALUES (NULL, 'patch', 'patch_99_100_b.sql|homology_id_64-bits');
```
JAlvarezJarreta: I'm not sure if not having and end of line at the end of the file will actually commit this query to the database when passed with `<`.

sql/table.sql
```diff
@@ -2253,3 +2253,5 @@ INSERT INTO meta (species_id, meta_key, meta_value) VALUES (NULL, 'schema_type',
 # Patch identifier
 INSERT INTO meta (species_id, meta_key, meta_value)
   VALUES (NULL, 'patch', 'patch_99_100_a.sql|schema_version');
+INSERT INTO meta (species_id, meta_key, meta_value)
+  VALUES (NULL, 'patch', 'patch_99_100_b.sql|homology_id_64-bits');
```
JAlvarezJarreta: Same as before.

modules/t/test-genome-DBs/homology/compara/table.sql
```diff
@@ -388,8 +388,14 @@ CREATE TABLE `homology` (
   KEY `gene_tree_root_id` (`gene_tree_root_id`)
 ) ENGINE=MyISAM AUTO_INCREMENT=100990070 DEFAULT CHARSET=latin1;
 
+CREATE TABLE `homology_id_mapping` (
+  `curr_release_homology_id` int(10) unsigned NOT NULL,
+  `prev_release_homology_id` int(10) unsigned NOT NULL,
+  `mlss_id` int(10) unsigned NOT NULL
+) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```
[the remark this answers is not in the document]
muffato: That's because `homology_id_mapping` comes from `pipeline-tables.sql` which uses InnoDB for all tables (as opposite to `table.sql` which uses MyISAM by default).
BTW this table can be removed from the schema because it's not used any more, see https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-2788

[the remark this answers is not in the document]
muffato: Oh, wait, actually @carlacummins you've already removed it :smile: ! https://github.com/Ensembl/ensembl-compara/commit/d12ebc9ce153891e86c88be3f75e30840e686971 . 
@CristiGuijarro can you remove references to this table from all the test databases ? (and maybe assign the Jira ticket to yourself and explain what's actually needed)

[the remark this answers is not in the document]
CristiGuijarro: Oh, I just misread that and created the ticket linked. I thought you meant Carla had already removed it... Can someone delete the Jira ticket I just made?

CristiGuijarro:
I am a little lost @muffato, what would you like me to change? Would you like me to take the newline at the end of the files back out again?

sql/table.sql
```diff
@@ -2253,3 +2253,6 @@ INSERT INTO meta (species_id, meta_key, meta_value) VALUES (NULL, 'schema_type',
...
+
```
muffato: It's the warning here. It means that the line 2258 does not follow the convention of ending with a newline itself. GitHub doesn't show it well but the line only has two whitespace characters and no newline. So in fact, the two whitespace should be removed.

sql/patch_99_100_b.sql
```diff
@@ -0,0 +1,32 @@
...
+
```
muffato: Same here there are two whitespaces on line 32

CristiGuijarro: Ah, I didn't realise

PR #160, opened 2019-11-27
Check that our gene-reuse detection matches the ensembl-metadata report

muffato:
With this change _PrepareMasterForRelease_ will write to a file the list of genomes that are reported to have a new annotation set. Then _LoadMembers_ will compare this list against the set of genomes that are not reused and complains about the differences.

I have introduced a mechanism to manually OK a difference as things like gene name / description updates are probably listed as "new annotation" but it's not something we care about in our reuse detection.
@JAlvarezJarreta made the comment this morning that _LoadMembers_ could read from the JSON reports directly. I decided to have a separate file because for Plants we need only certain species from the Vertebrates report, and I didn't want to list these species again in _LoadMembers_.

What I don't like in this implementation is that the file is created in the pipeline directory, so _LoadMembers_ is configured with a path that we hope is the path of the other pipeline, but if someone else is running it, or if the version suffix differs, this will break.

PS: I have also introduced a new method in _BaseRunnable_: `die_no_retry`, which sets "transient_error" to 0 since there is no point recomparing the lists.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/LoadMembers_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/LoadMembers_conf.pm
@@ -110,7 +110,7 @@ sub default_options {
 
         # list of species that got an annotation update
         # ... assuming the same person has run both pipelines
-        'expected_updates_file' => '/hps/nobackup2/production/ensembl/' . $ENV{'USER'} . '/prep_' . $self->o('division') . '_master_for_rel_' . $self->o('rel_with_suffix') .  '/annotation_updates.list',
+        'expected_updates_file' => $self->o('shared_hps_dir') . '/ensembl-metadata/annotation_updates.' . $self->o('division') . '.' . $self->o('ensembl_release') . '.list',
     };
 }
 
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/LoadMembers_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/LoadMembers_conf.pm
@@ -111,7 +111,6 @@ sub default_options {
         # list of species that got an annotation update
         # ... assuming the same person has run both pipelines
         'expected_updates_file' => '/hps/nobackup2/production/ensembl/' . $ENV{'USER'} . '/prep_' . $self->o('division') . '_master_for_rel_' . $self->o('rel_with_suffix') .  '/annotation_updates.list',
-
     };
 }
 
@@ -613,4 +612,3 @@ sub pipeline_analyses {
 }
 
 1;
-
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/LoadMembers_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/LoadMembers_conf.pm
@@ -107,6 +107,11 @@ sub default_options {
         # members
         'include_nonreference' => 0,
         'include_patches'      => 0,
+
+        # list of species that got an annotation update
+        # ... assuming the same person has run both pipelines
+        'expected_updates_file' => '/hps/nobackup2/production/ensembl/' . $ENV{'USER'} . '/prep_' . $self->o('division') . '_master_for_rel_' . $self->o('rel_with_suffix') .  '/annotation_updates.list',
+
     };
 }
 
@@ -263,7 +268,16 @@ sub pipeline_analyses {
         {   -logic_name => 'create_reuse_ss',
             -module     => 'Bio::EnsEMBL::Compara::RunnableDB::CreateReuseSpeciesSets',
             -rc_name => '2Gb_job',
-            -flow_into => [ 'nonpolyploid_genome_reuse_factory' ],
+            -flow_into  => [ 'compare_non_reused_genome_list' ],
+        },
+
+        {   -logic_name => 'compare_non_reused_genome_list',
+            -module     => 'Bio::EnsEMBL::Compara::RunnableDB::LoadMembers::CompareNonReusedGenomeList',
+            -parameters => {
+                'expected_updates_file' => $self->o('expected_updates_file'),
+                'current_release'       => $self->o('ensembl_release'),
+            },
+            -flow_into  => [ 'nonpolyploid_genome_reuse_factory' ],
         },
 
         {   -logic_name => 'check_reuse_db_is_patched',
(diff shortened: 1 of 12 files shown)


modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMembers/CompareNonReusedGenomeList.pm
```diff
@@ -0,0 +1,120 @@
...
+        -expected_updates_file <(echo "homo_sapiens"; echo "mus_musculus"; echo "equus_caballus") \
+        -current_release 99 -nonreuse_ss_id 10000002
+
+=head1 DESCRIPTION
+
+This RunnableDB compares the list of genomes that are expected to have an
+annotation update (provided in a file) to the actual list of genomes that
+cannot be reused.
+
+Both lists must match, and the job fill fail if there are any references.
```
CristiGuijarro:
```suggestion
Both lists must match, and the job will fail if there are any references.
```

modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMembers/CompareNonReusedGenomeList.pm
```diff
@@ -0,0 +1,120 @@
...
+=head1 NAME
+
+Bio::EnsEMBL::Compara::RunnableDB::LoadMembers::CompareNonReusedGenomeList
+
+=head1 SYNOPSIS
+
+    standaloneJob.pl Bio::EnsEMBL::Compara::RunnableDB::LoadMembers::CompareNonReusedGenomeList \
+        -compara_db $(mysql-ens-compara-prod-8 details url jalvarez_vertebrates_load_members_99) \
+        -expected_updates_file <(echo "homo_sapiens"; echo "mus_musculus"; echo "equus_caballus") \
+        -current_release 99 -nonreuse_ss_id 10000002
```
JAlvarezJarreta: I didn't know runnables should have this synopsis format... I like it :)
But I would replace `SYNOPSIS` by `EXAMPLE` in this case, as I think the synopsis shouldn't have specific values, e.g. the release number.

modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMembers/CompareNonReusedGenomeList.pm
```diff
@@ -0,0 +1,120 @@
...
+
+This RunnableDB compares the list of genomes that are expected to have an
+annotation update (provided in a file) to the actual list of genomes that
+cannot be reused.
+
+Both lists must match, and the job fill fail if there are any references.
+If a difference happens to be fine, the species can be manually okay-ed
+by adding a parameter named "ok_${species_name}".
+Note: the Runnable respects the "do_not_reuse_list" parameter and won't
+complain that those species are not listed in the file.
```
JAlvarezJarreta: Since we were thinking about improving/updating our PODs, I was wondering if all the "non-eHive-runnable" arguments should be defined with `=item`. Would it be useful?

modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMembers/CompareNonReusedGenomeList.pm
```diff
@@ -0,0 +1,120 @@
...
+
+use Bio::EnsEMBL::Utils::IO qw/slurp_to_array/;
+
+use base ('Bio::EnsEMBL::Compara::RunnableDB::BaseRunnable');
+
+
+sub param_defaults {
+    my $self = shift @_;
+    return {
+        %{ $self->SUPER::param_defaults },
```
JAlvarezJarreta:
```suggestion
        %{ $self->SUPER::param_defaults },

```

modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromMetadataFactory.pm
```diff
@@ -53,7 +53,7 @@ sub fetch_input {
 	my $genome_db_adaptor = $self->get_cached_compara_dba('master_db')->get_GenomeDBAdaptor;
 
 	# use metadata script to report genomes that need to be updated
-	my ($genomes_to_update, $renamed_genomes, $genomes_with_assembly_patches) = $self->fetch_genome_report($release, $division);
+	my ($genomes_to_update, $renamed_genomes, $genomes_with_assembly_patches, $updated_annotations) = $self->fetch_genome_report($release, $division);
```
JAlvarezJarreta:
```suggestion
    my ($genomes_to_update, $renamed_genomes, $genomes_with_assembly_patches, $updated_annotations) = $self->fetch_genome_report($release, $division);
```
Since you are already changing the line...

modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMembers/CompareNonReusedGenomeList.pm
muffato: Agreed. Will do that.

modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMembers/CompareNonReusedGenomeList.pm
muffato: Will use `EXAMPLE` instead. With the `=item` list below I think it will be clear how to invoke the runnable

modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMembers/CompareNonReusedGenomeList.pm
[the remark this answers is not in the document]
muffato: `s/references/differences/`

modules/Bio/EnsEMBL/Compara/PipeConfig/LoadMembers_conf.pm
```diff
@@ -107,6 +107,11 @@ sub default_options {
         # members
         'include_nonreference' => 0,
         'include_patches'      => 0,
+
+        # list of species that got an annotation update
+        # ... assuming the same person has run both pipelines
+        'expected_updates_file' => '/hps/nobackup2/production/ensembl/' . $ENV{'USER'} . '/prep_' . $self->o('division') . '_master_for_rel_' . $self->o('rel_with_suffix') .  '/annotation_updates.list',
```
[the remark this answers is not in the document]
muffato: OK. My preference is to use the shared hps space for this file, rather than the warehouse. The JSON report, however, would fit well in warehouse.

PR #162, opened 2019-11-28
Populate the Category field of the Jira tickets

muffato:
Categories are mostly handled the same way as Components (multi-value fields, but using the `value` key in JSON rather than `name`).

HC and DC tickets are under _Bug::Internal_ and _Production::Tasks_.

Instead of defining the Category in every entry of our JSON config files (for the relco / production tickets), I've made `Utils::JIRA` automatically populate it from the Component, since in this case there is a a 1-to-1 mapping. Categories are automatically set _only_ if none is given, and it will only work for the components that can be mapped (relco and production).

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/Utils/JIRA.pm
+++ b/modules/Bio/EnsEMBL/Compara/Utils/JIRA.pm
@@ -32,6 +32,10 @@ use Bio::EnsEMBL::Utils::Argument qw(rearrange);
 use Bio::EnsEMBL::Utils::IO qw (slurp);
 use Bio::EnsEMBL::Utils::Logger;
 
+# JIRA identifiers of the custom field used in Compara (ENSCOMPARASW)
+use constant DIVISION_CUSTOM_FIELD_ID => 'customfield_11130';
+use constant CATEGORY_CUSTOM_FIELD_ID => 'customfield_11333';
+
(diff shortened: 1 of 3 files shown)


modules/Bio/EnsEMBL/Compara/Utils/JIRA.pm
```diff
@@ -359,6 +377,24 @@ sub _json_to_jira {
...
+    if ($json_hash->{'category'}) {
+        push @{$jira_hash{CATEGORY_CUSTOM_FIELD_ID}}, { 'value' => $json_hash->{'category'} };
+    } elsif ($json_hash->{'categories'}) {
+        push @{$jira_hash{CATEGORY_CUSTOM_FIELD_ID}}, { 'value' => $_ } for @{$json_hash->{'categories'}};
+    }
+    if ($extra_categories) {
+        push @{$jira_hash{CATEGORY_CUSTOM_FIELD_ID}}, { 'value' => $_ } for @{$extra_categories};
+    }
+    unless (scalar(@{$jira_hash{CATEGORY_CUSTOM_FIELD_ID}})) {
+        # Fallback to automatically setting the categories from the component names
```
JAlvarezJarreta:
```suggestion
        # Fallback to automatically set the categories from the component names
```

PR #163, opened 2019-11-29
Run podchecker on Travis

muffato:
This PR enables `podchecker`, a tool to check the syntax of Perl documentation blocks (PODs), on Travis. At the same time, I fixed the errors on existing files (there were not that many), and you can see the typical errors we had, e.g.:
 - it wants 1 blank line before and after each `=XXX` command, not more than 1 on the inside of the block
 - inside `=NAME` text should not be indented
 - sections cannot be empty
 - `=head2` must be preceded by `=head1`

I have also removed a few `=pod` that had no effect.

### DIFF
--- a/scripts/compare/flatten_mouse_files.pl
+++ b/scripts/compare/flatten_mouse_files.pl
@@ -15,10 +15,10 @@
 # limitations under the License.
 
 
-=pod
+=head1 DESCRIPTION
 
-    This script fetches data from MGI and maps its ids with Ensembl stable ids.
-    MGI annotations come dorectly from the web [http://www.genenames.org]
+This script fetches data from MGI and maps its ids with Ensembl stable ids.
+MGI annotations come dorectly from the web [http://www.genenames.org]
 
 =cut
 
--- a/scripts/compare/flatten_mouse_files.pl
+++ b/scripts/compare/flatten_mouse_files.pl
@@ -15,9 +15,11 @@
 # limitations under the License.
 
 
-=head1
+=pod
+
     This script fetches data from MGI and maps its ids with Ensembl stable ids.
     MGI annotations come dorectly from the web [http://www.genenames.org]
+
 =cut
 
 use strict;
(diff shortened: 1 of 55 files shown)


muffato:
Damn it, I was using an old `podchecker` on my laptop. Newer ones find more stuff. Will add an extra commit soon

muffato:
It doesn't like lines that are only whitespace. I believe everything is now fixed

muffato:
A little bit of context. I am working on adding the Python checkers to Travis (upcoming PR, preview here: https://travis-ci.org/Ensembl/ensembl-compara/builds/619578551 ), and thought I could add a Perl one as well.

scripts/compare/flatten_mouse_files.pl
```diff
@@ -15,9 +15,11 @@
 # limitations under the License.
 
 
-=head1
+=pod
+
```
JAlvarezJarreta: Based on the convention we are following in other files, I would take the opportunity to fix this, replacing `=pod` by:
```
=cut

=head1 DESCRIPTION

```

scripts/compare/flatten_zebrafish_files.pl
```diff
@@ -14,11 +14,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-=head1
+=pod
```
JAlvarezJarreta: Sames as previous comment.

scripts/projection/map_coordinates.pl
```diff
@@ -14,14 +14,15 @@
 # limitations under the License.
 
 
+=pod
```
JAlvarezJarreta: Same as my previous comment.

scripts/compare/flatten_mouse_files.pl
muffato: `DESCRIPTION` is a good idea

PR #165, opened 2019-12-02
Update runnable to use new Ortheus parameters

JAlvarezJarreta:
These new parameters were introduced to avoid exceptions raised due to command lines being too long for bash to process them.


modules/Bio/EnsEMBL/Compara/Production/EPOanchors/TrimAnchorAlign.pm
```diff
@@ -161,6 +148,7 @@ sub _dump_fasta {
       $self->complete_early( "Not enough sequences in current anchor - omitting" );
   }
 
+  my $fasta_files;
```
muffato:
```suggestion
  my $fasta_files = '';
```

JAlvarezJarreta: I was hoping Perl was clever enough to do this, but I guess it is not :'(

PR #157, opened 2019-11-12
Dev/citest

dthybert:
This is the initial generic tool for CITest. The tool is able to make db comparison with tests specified in json file. The test has been developed using the LoadMember pipeline where strict comparison are done.Note that while comparison with margin error are authorised it has not been tested in the context of LoadMember.
The package includes:
 - LoadMember.json : describing the test to be done for LoadMemeber
 - LoadMembers_conf.pm : the configuration file for runiing the the pipeline under CITest environment
 - CITest.py : the main script to be executed for running the tests
 - citest.py: file describing the the main class citest which read config file and run test
 - test.py: file describing the base class for a test. This class has to be inherited and extended to impleemnt specific tests (dbtests, file test, tre etc...)
 -tabletest.py file describing the class that test database tables (inherit from Test)

PR #166, opened 2019-12-02
Updated Travis configuration, incl. some Python tests

muffato:
I have refactored the tests that run on Travis, separating the ones that are Perl-specific, from the others. This allows running some tests in parallel and makes the whole build faster. They now all have a name explaining what it does. See https://travis-ci.org/Ensembl/ensembl-compara/builds/619708656

I have introduced a new module `Bio::EnsEMBL::Compara::Utils::Test` with a function that returns all the files in the repository. It is used by several .t tests (license, perlcritic, etc)

There is obviously room to add the Python unit-tests when we have them (#157 ?). In the meantime, this PR enables _pylint_ and this is the time to agree on the rules we want to follow. There is now a `pylintrc` file at the root of the repository. I have copied a list of exceptions from https://github.com/Ensembl/template-python . These tests are disabled:
* similarities
* invalid-name
* too-many-statements
* too-many-arguments
* too-many-locals
* too-few-public-methods
* relative-import
* no-self-use

The first one would for instance complain that `reroot_newick.py` and `unroot_newick.py` are too similar and should be refactored. `invalid-name` requires all the global variables to be named in UPPER_CASE. I know that some files in #157 would not pass `too-many-arguments`.
And I have increased the maximum line width from 100 to 110 because the Ensembl copyright line is 102 characters long ...

You can see how the two existing Python scripts had to be changed. All minor changes. We should also all configure our editor / IDE to run pylint (with this config file if possible)


modules/Bio/EnsEMBL/Compara/Utils/Test.pm
```diff
@@ -0,0 +1,100 @@
...
+
+my $repository_root;
+sub get_repository_root {
+    return $repository_root if $repository_root;
+    my $file_dir = dirname(__FILE__);
+    my $original_dir = cwd();
+    chdir($file_dir);
+    my $cur_dir = cwd();
+    chdir($original_dir);
+    $repository_root = File::Spec->catdir($cur_dir, File::Spec->updir(), File::Spec->updir(), File::Spec->updir(), File::Spec->updir(), File::Spec->updir());
```
JAlvarezJarreta: I'm always a little bit afraid of this way of writing code as if anything changes it will break. What about the following?
```suggestion
($repository_root = $cur_dir) =~ s/ensembl_compara.*/ensembl_compara/;
```
I know it forces to have the repository under `ensembl_compara`, but who would dare to change it? :P

modules/Bio/EnsEMBL/Compara/Utils/Test.pm
```diff
@@ -0,0 +1,100 @@
...
+=cut
+
+my $repository_root;
+sub get_repository_root {
+    return $repository_root if $repository_root;
+    my $file_dir = dirname(__FILE__);
+    my $original_dir = cwd();
+    chdir($file_dir);
+    my $cur_dir = cwd();
+    chdir($original_dir);
```
JAlvarezJarreta: I'm struggling to see the reason for this instead of using `abs_path($file_dir)`.

modules/Bio/EnsEMBL/Compara/Utils/Test.pm
```diff
@@ -0,0 +1,100 @@
...
+    $repository_root = File::Spec->catdir($cur_dir, File::Spec->updir(), File::Spec->updir(), File::Spec->updir(), File::Spec->updir(), File::Spec->updir());
+    return $repository_root;
+}
+
+=head2 find_all_files
+
+  Description : Return the list of all the files in the repository
```
JAlvarezJarreta:
```suggestion
  Description : Return the list of all the files in the repository.
```

travisci/pylint_harness.sh
```diff
@@ -0,0 +1,28 @@
...
+# 
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+tmpfile=$(mktemp)
+
+pylint --rcfile pylintrc scripts/ | grep -v "\-\-\-\-\-\-\-\-\-" | grep -v "Your code has been rated" | grep -v "\n\n" | sed '/^$/d' > "$tmpfile"
```
JAlvarezJarreta: Should we include also the future `src/python/`? Or will it fail/complain because that path doesn't exist at the moment?

.travis.yml
```diff
@@ -53,21 +46,61 @@ install:
...
+      perl: 5.30
+      name: "Perl linter"
+      install:
+        - cpanm --quiet --installdeps --notest --cpanfile ensembl/cpanfile .
+      before_script:
+        - echo
+      script: ./travisci/perl-linter_harness.sh
+
+    - language: python
+      python: 3.6             # 3.8 is not available on Trusty builds
```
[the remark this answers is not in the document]
muffato: I'm following the same pattern as for Perl:
 - linter on the latest version
 - unit tests on the minimum supported version and the latest one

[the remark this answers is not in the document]
muffato: BTW, I'm going to push another commit that will bump this one to 3,8 as I've found a way of using this version

travisci/all-housekeeping/apache2.t
```diff
@@ -16,34 +16,15 @@
...
 
-#chdir into the file's target & request cwd() which should be fully resolved now.
-#then go back
-my $file_dir = dirname(__FILE__);
-my $original_dir = cwd();
-chdir($file_dir);
-my $cur_dir = cwd();
-chdir($original_dir);
-my $root = File::Spec->catdir($cur_dir, File::Spec->updir(),File::Spec->updir());
+use Bio::EnsEMBL::Test::TestUtils;
```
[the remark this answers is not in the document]
muffato: It's needed. It's where `has_apache2_licence` is defined

modules/Bio/EnsEMBL/Compara/Utils/Test.pm
muffato: Can't find any reason not too :)

modules/Bio/EnsEMBL/Compara/Utils/Test.pm
muffato: This regex or a _while_ loop that traverses the directories up until it finds `ensembl-compara` ?

travisci/pylint_harness.sh
muffato: It fails if asked to run on a file that doesn't exist. It should be added in #157

modules/Bio/EnsEMBL/Compara/Utils/Test.pm
JAlvarezJarreta: Performance-wise I guess the former, but either could work if you think the latter will produce a more readable code.

muffato:
I want to push a few more commits to define stubs for the Python unit-tests and run _mypy_. I can add them to this PR, make another PR, or just push. I anyway want to discuss _mypy_ with everyone in the office this afternoon as well

carlacummins:
Would it be useful to have a unit test for the new `Utils::Test` module? (I see coveralls is not happy)

.travis.yml
```diff
@@ -53,21 +46,95 @@ install:
...
+      perl: 5.30
+      name: "Perl linter"
+      install:
+        - cpanm --quiet --installdeps --notest --cpanfile ensembl/cpanfile .
+      before_script:
+        - echo
+      script: ./travisci/perl-linter_harness.sh
+
+    - language: python
+      python: 3.4
```
carlacummins: Should be 3.5?

modules/Bio/EnsEMBL/Compara/Utils/Test.pm
```diff
@@ -0,0 +1,113 @@
...
+
+sub get_repository_root {
+    return $repository_root if $repository_root;
+    my $file_dir = dirname(abs_path(__FILE__));
+    if ($file_dir =~ /^(.*\/ensembl-compara)\b/) {
+        $repository_root = $1;
+        return $repository_root;
+    } else {
+        die "Can't find a parent directory named 'ensembl-compara' in '$file_dir'";
+    }
```
carlacummins: I use several checkouts of compara in development. They can be named things like `ensembl-compara-master` or `ensembl-compara-release`. While I'm able to just change this regex in each one, I would prefer not to have to..

I have a script that performs exactly this function which continually trims the last dir from a path until `-e $dir/.git` - could we use this means of finding the repo root? (I see you've already rewritten this method once, so no problem if you don't want to do it again)

travisci/python-unittest_harness.sh
```diff
@@ -26,7 +26,12 @@ else
   rt1=0
 fi
 
-pytest "${PYTHON_SOURCE_LOCATIONS[@]}"
+PYTEST_OPTIONS=()
+if [ "$COVERAGE" = 'true' ]
+then
```
JAlvarezJarreta: More a suggestion than anything:
```
if [ "$COVERAGE" = 'true' ]; then
````

File: travisci/python-linter_harness.sh [hunk 10]
```diff
@@ -0,0 +1,35 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+PYTHON_SOURCE_LOCATIONS=('scripts')
+
+PYLINT_OUTPUT_FILE=$(mktemp)
+PYLINT_ERRORS=$(mktemp)
```
carlacummins: Can we rely on `mktemp` being available everywhere?

JAlvarezJarreta: Not in windows: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/mktemp?view=vs-2019

modules/Bio/EnsEMBL/Compara/Utils/Test.pm
JAlvarezJarreta: I must confess I haven't been able to find any solution that I'm 100% happy with. If I was to run your suggestion in my computer, my unison is not copying the `.git` folders, so it wouldn't work...
To make it work for you too, we could just update the regex:
```
    if ($file_dir =~ /^(.*\/ensembl-compara[^\\\/])\b/) {
```

travisci/python-linter_harness.sh
muffato: On linux, yes it's present everywhere. If I try to remove `coreutils` (the package it's part of), it says
```
root@d48ff7c56ce2:/# apt-get remove coreutils
(...)
WARNING: The following essential packages will be removed.
This should NOT be done unless you know exactly what you are doing!
  coreutils
(...)
You are about to do something potentially harmful.
To continue type in the phrase 'Yes, do as I say!'
 ?```

modules/Bio/EnsEMBL/Compara/Utils/Test.pm
muffato: I'm not 100% happy with this implementation either. Maybe I could be looking for `modules` and consider the parent directory to be the root ? That would work for you @carlacummins ?

.travis.yml
muffato: 3.4 there was the minimum I could get pytest to run on (I've tried the previous ones). But yes, 3.4 not being supported any more, it should be 3.5

modules/Bio/EnsEMBL/Compara/Utils/Test.pm
JAlvarezJarreta: Actually, that's a good point @muffato . Most things wouldn't work if that is manually modified by users... I like it :)

muffato:
I have put in `requirements-dev.txt` the development-time dependencies (_pylint_ etc). For production, and for external users wanting to run our scripts, the dependencies found in `requirements.txt` suffice (_ete3_ etc).

modules/Bio/EnsEMBL/Compara/Utils/Test.pm
carlacummins: Yes, that would work well! 👍

muffato: What about reverting to the initial implementation (go 5 directories up) but also add a unit-test checking that the directory returned by the function does contain say `modules`, `scripts` and `sql`. So if we move the function and break the test we would know ? That would also answer this comment:
> Would it be useful to have a unit test for the new `Utils::Test` module? (I see coveralls is not happy)

carlacummins: Killing two birds with one stone - nice 🥇

PR #172, opened 2020-01-09
Add new mlss_tag for EPO2x

JAlvarezJarreta:
## Description

In e99, a bug was discovered when we added the new pig EPO alignment. This is due to some slightly hacky code in the webcode that selects the 'principal' species in an EPO2x alignment for display purposes.

**Related JIRA tickets:**
- ENSCOMPARASW-3022

## Overview of changes
Added a new tag for the 'principal' species in EPO2x alignments. To avoid more comparisons than those necessary, this tag has been introduced in the `Bio::EnsEMBL::Compara::Utils::MasterDatabase::create_multiple_wga_mlsss` method.

## Testing
No testing performed yet.

## Notes
The value of this tag has been hardcoded as it is unlikely to change frequently.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/Utils/MasterDatabase.pm
+++ b/modules/Bio/EnsEMBL/Compara/Utils/MasterDatabase.pm
@@ -826,6 +826,7 @@ sub create_multiple_wga_mlsss {
         $principal_species = 'gallus_gallus'   if ($mlsss[0]->name =~ /sauropsids/);
         $principal_species = 'oryzias_latipes' if ($mlsss[0]->name =~ /fish/);
         $principal_species = 'sus_scrofa'      if ($mlsss[0]->name =~ /pig/);
+        die "Unexpected EPO_LOW_COVERAGE MLSS: cannot set a principal species\n" unless $principal_species;
         $mlsss[0]->add_tag('principal_species', $principal_species);
     }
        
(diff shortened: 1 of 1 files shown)


CristiGuijarro:
Is this something that may affect the new rice epo? Does there need to be a special case for rice?

JAlvarezJarreta:
> Is this something that may affect the new rice epo? Does there need to be a special case for rice?

If you are doing both EPO and EPO2x, the answer is yes. What principal species do you want me to set for rice?

CristiGuijarro:
> > Is this something that may affect the new rice epo? Does there need to be a special case for rice?
> 
> If you are doing both EPO and EPO2x, the answer is yes. What principal species do you want me to set for rice?

Oryza_sativa would probably be most appropriate? Just because it is a reference genome, and listed on the site. Otherwise, how does one choose?

muffato:
Sorry but I really don't like this fix. What you're describing is a mandatory tag that must be present for every EPO_LOW_COVERAGE alignment. Its place is in `mlss_conf.xml` if we want to declare it ourselves.

Otherwise, I think it could be set dynamically by the pipeline itself
- Option 1: choose the species that contributes the highest number of pairwise alignments
- Option 2: record as tags (instead of a pipeline-wide parameter) the mapping of reference species

JAlvarezJarreta:
I have found that this information was already there in a different way, so there is no need to update our codebase (see [ENSCOMPARASW-3022](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-3022)).

PR #173, opened 2020-01-21
Feature/on demand previous registry

muffato:
## Description

I want to avoid having to uncomment / comment out the bit in `production_reg_conf.pl` that loads the previous core databases into the Registry.

## Overview of changes

Here I wrap this part of the code in a function that belongs to our `Utils::Registry` and I add helper methods in `Utils::Registry` to search for a previous core database and do the loading automatically (just once) when needed. I have updated `CheckGenomeReusability.pm` and `Utils/MasterDatabase.pm` to use the new interface.

The new interface is still compatible with "old-style" production_reg_conf files (as long as they use the `__cut_here__` string and the correct version number).

You don't need to review these two functions as they serve another purpose.
- `cleanup_adaptors_from_vertebrates_server` in `production_reg_conf.pl`
- `load_collection_core_database` in `Utils::Registry`

## Testing

This is the code I've used during the Pan-Compara production.

## Further thoughts

Doing this made it more apparent (to me) that we do `$release` minus 1 in many places, but since we only ever load one set of previous core databases, what about using a fixed suffix like `__previous_database` ? Then it's up to the production_reg_conf to load the required previous version, and consumers ( `CheckGenomeReusability.pm` and `Utils/MasterDatabase.pm`) don't need to bother about checking the release number.

### DIFF
--- a/conf/pan/production_reg_conf.pl
+++ b/conf/pan/production_reg_conf.pl
@@ -52,34 +52,37 @@ Bio::EnsEMBL::Compara::Utils::Registry::load_collection_core_database(
 # ---------------------- PREVIOUS CORE DATABASES---------------------------------
 
 # previous release core databases will be required by PrepareMasterDatabaseForRelease and LoadMembers only
-# !!! COMMENT THIS SECTION OUT FOR ALL OTHER PIPELINES (for speed) !!!
-#my $suffix_separator = '__cut_here__';
-#Bio::EnsEMBL::Registry->load_registry_from_db(
-#    -host   => 'mysql-ens-mirror-1',
-#    -port   => 4240,
-#    -user   => 'ensro',
-#    -pass   => '',
-#    -db_version     => $prev_release,
-#    -species_suffix => $suffix_separator.$prev_release,
-#);
-#Bio::EnsEMBL::Registry->remove_DBAdaptor('saccharomyces_cerevisiae'.$suffix_separator.$prev_release, 'core'); # never use Vertebrates' version of yeast
-#Bio::EnsEMBL::Registry->load_registry_from_db(
-#    -host   => 'mysql-ens-mirror-3',
-#    -port   => 4275,
-#    -user   => 'ensro',
-#    -pass   => '',
-#    -db_version     => $prev_release,
-#    -species_suffix => $suffix_separator.$prev_release,
-#);
-# Bacteria server: all species used in Pan happen to be in this database
-#Bio::EnsEMBL::Compara::Utils::Registry::load_collection_core_database(
-#    -host   => 'mysql-ens-mirror-4',
-#    -port   => 4495,
-#    -user   => 'ensro',
-#    -pass   => '',
-#    -dbname => "bacteria_0_collection_core_${prev_eg_release}_${prev_release}_1",
-#    -species_suffix => $suffix_separator.$prev_release,
-#);
+*Bio::EnsEMBL::Compara::Utils::Registry::load_previous_core_databases = sub {
+    my $release_number  = shift;
+    my $species_suffix  = Bio::EnsEMBL::Compara::Utils::Registry::SUFFIX_SEPARATOR.$release_number;
+
+    Bio::EnsEMBL::Registry->load_registry_from_db(
+        -host   => 'mysql-ens-mirror-1',
+        -port   => 4240,
+        -user   => 'ensro',
+        -pass   => '',
+        -db_version     => $release_number,
+        -species_suffix => $species_suffix,
+    );
+    Bio::EnsEMBL::Registry->remove_DBAdaptor('saccharomyces_cerevisiae'.$species_suffix, 'core'); # never use Vertebrates' version of yeast
+    Bio::EnsEMBL::Registry->load_registry_from_db(
+        -host   => 'mysql-ens-mirror-3',
+        -port   => 4275,
+        -user   => 'ensro',
+        -pass   => '',
+        -db_version     => $release_number,
+        -species_suffix => $species_suffix,
+    );
+    # Bacteria server: all species used in Pan happen to be in this database
+    Bio::EnsEMBL::Compara::Utils::Registry::load_collection_core_database(
+        -host   => 'mysql-ens-mirror-4',
+        -port   => 4495,
+        -user   => 'ensro',
+        -pass   => '',
+        -dbname => "bacteria_0_collection_core_".($release_number-53)."_${release_number}_1",
+        -species_suffix => $species_suffix,
+    );
+};
 
 #------------------------COMPARA DATABASE LOCATIONS----------------------------------
 
--- a/conf/pan/production_reg_conf.pl
+++ b/conf/pan/production_reg_conf.pl
@@ -59,26 +59,23 @@ Bio::EnsEMBL::Compara::Utils::Registry::load_collection_core_database(
 
 # previous release core databases will be required by PrepareMasterDatabaseForRelease and LoadMembers only
 *Bio::EnsEMBL::Compara::Utils::Registry::load_previous_core_databases = sub {
-    my $release_number  = shift;
-    my $species_suffix  = Bio::EnsEMBL::Compara::Utils::Registry::SUFFIX_SEPARATOR.$release_number;
-
     Bio::EnsEMBL::Registry->load_registry_from_db(
         -host   => 'mysql-ens-mirror-1',
         -port   => 4240,
         -user   => 'ensro',
         -pass   => '',
-        -db_version     => $release_number,
-        -species_suffix => $species_suffix,
+        -db_version     => $prev_release,
+        -species_suffix => Bio::EnsEMBL::Compara::Utils::Registry::PREVIOUS_DATABASE_SUFFIX,
     );
-    Bio::EnsEMBL::Compara::Utils::Registry::remove_species(\@overlap_species, $species_suffix);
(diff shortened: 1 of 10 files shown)


muffato:
Another thought: I could implement a similar system for the _current_ Core databases.This would make the registry faster to initialise because it would defer the loading Core databases to when required, if required at all (e.g. not if the locators are written in). I can make it totally transparent as long as one is accessing the Core databases through `GenomeDB::db_adaptor`, but a script that expects the Registry to be fully loaded and does directly `get_adaptor($species_name, 'core', 'slice')` would have to be adjusted (not much, just 1 function call)

conf/pan/production_reg_conf.pl
```diff
@@ -35,35 +35,72 @@
...
+    # Bacteria server: all species used in Pan happen to be in this database
+    Bio::EnsEMBL::Compara::Utils::Registry::load_collection_core_database(
+        -host   => 'mysql-ens-mirror-4',
+        -port   => 4495,
+        -user   => 'ensro',
+        -pass   => '',
+        -dbname => "bacteria_0_collection_core_".($release_number-53)."_${release_number}_1",
+        -species_suffix => $species_suffix,
+    );
+};
```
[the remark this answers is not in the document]
muffato: How to do this ? The list of the servers, their order, and the cleanup bit will change from one division to another

modules/Bio/EnsEMBL/Compara/Utils/Registry.pm
```diff
@@ -55,14 +55,83 @@ use strict;
...
+  Description : Wrapper around load_previous_core_databases to only call it once
+  Returntype  : none
+  Exceptions  : none
+
+=cut
+
+my %loaded_previous;
+sub load_previous_core_databases_if_needed {
+    my ($release_number) = @_;
+    return unless defined &load_previous_core_databases;
```
carlacummins: Wondering if we should issue at least a warning in this case? In what scenario are we happy to call this loading method, but allow it to silently not do anything?

muffato: I made it this way to allow "old-style" production_reg_conf to coexist with my new-style file for Pan-Compara, but I can roll it out to all divisions and then Registry.pm can assume `load_previous_core_databases` always exists

muffato:
Looks quite simpler with the release numbers out of the way !

muffato:
There are now too many non-trivial changes to be put on the release branch. I will push the e100 branch:
 - the new `load_collection_core_database` and `remove*` functions in Utils::Registry
 - a valid production_reg_conf for Pan but without the `load_previous_core_databases` handler

and move this pull-request to master (in which I merged release/100 this morning)

muffato:
I don't really need the `remove*` functions. I can do our usual
```
Bio::EnsEMBL::Registry->remove_DBAdaptor('saccharomyces_cerevisiae');
```
since it's only yeast that's important to remove. The other adaptors are only cleaned up to avoid unnecessary warnings

conf/pan/production_reg_conf.pl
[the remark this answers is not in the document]
JAlvarezJarreta: You could make a function that accepts the list of mirrors and the `\@overlap_species`, and loads them in order, running the `remove_*` part between the first and the second mirrors (if more than 1 is given). So in the registry you would have something like the following in Pan:
```
Bio::EnsEMBL::Compara::Utils::Registry::load_previous_core_databases(['mysql-ens-mirror-1', 'mysql-ens-mirror-3', 'mysql-ens-mirror-4'], \@overlap_species);
```
and this for plants:
```
Bio::EnsEMBL::Compara::Utils::Registry::load_previous_core_databases(['mysql-ens-mirror-3', 'mysql-ens-mirror-1'], \@overlap_species);
```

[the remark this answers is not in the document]
muffato: I'm not keen on having this specific operation inserted between the 1st and 2nd servers, but not between 2nd and 3rd, or 3rd and 4th, etc, with no control.
Thinking about this in the car, what about:
```perl
Bio::EnsEMBL::Compara::Utils::Registry::setup_previous_core_databases($prev_release,
    ['load_all_from_server', 'mysql-ens-mirror-3'],
    ['remove_species', \@overlap_species],
    ['remove_multi, undef],
    ['load_from_server', 'mysql-ens-mirror-1'],
);
```

By adding `load_all_from_server` to Utils::Registry and giving it the interface `($server_name, $release_number, $optional_species_suffix)`, `setup_previous_core_databases` could simply iterate over the array and for each element `$a` call the function named `$a->[0]` with the arguments `($a->[1], $prev_release, PREVIOUS_DATABASE_SUFFIX)`. For Pan we'd need a wrapper around `load_collection_core_database` with the same interface.

Then, with a slight adjustment, we can have `setup_current_core_databases` too, which I want to address in another PR anyway because it is touching more files and I don't want to make it too confusing.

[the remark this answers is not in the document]
JAlvarezJarreta: I agree with you, that would be the way to approach it if you want to have control on the steps within the method. But then I feel it starts to reduce the advantage of moving the function away from the registry, as "most" of the logic/code is still there...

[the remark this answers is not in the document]
muffato: To me the advantage is that we move towards abstracting the definition of the Registry as a structure of hash / lists that could be moved into JSON and be read either from `production_reg_conf.pl` or whatever equivalent we will have in Python. Something like
```json
{
  "current_release": 100,
  "current_cores": [
    ["load_all_from_server", "mysql-ens-sta-3"],
    ["cleanup_overlap", ["saccharomyces_cerevisiae", "drosophila_melanogaster", "caenorhabditis_elegans"]],
    ["load_all_from_server", "mysql-ens-sta-1"],
  ],
  "previous_cores": [
    ["load_all_from_server", "mysql-ens-mirror-3"],
    ["cleanup_overlap", ["saccharomyces_cerevisiae", "drosophila_melanogaster", "caenorhabditis_elegans"]],
    ["load_all_from_server", "mysql-ens-mirror-1"],
  ],
  "aliases": {
    "compara": {
      "compara_master" : ["mysql-ens-compara-prod-5", "ensembl_compara_master_plants"],
      "compara_curr"   : ["mysql-ens-compara-prod-5", "ensembl_compara_plants_<curr_eg_release>_<curr_release>"],
      "compara_prev"   : ["mysql-ens-compara-prod-5", "ensembl_compara_plants_<prev_eg_release>_<prev_release>"],
    },
    "core": {
      "ancestral_curr" : ["mysql-ens-compara-prod-5", "ensembl_ancestral_plants_<curr_eg_release>_<curr_release>"],
    },
    "taxonomy": {
      "ncbi_taxonomy"  : ["mysql-ens-sta-1", "ncbi_taxonomy_<curr_release>" ],
    },
  },
}
```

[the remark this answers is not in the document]
JAlvarezJarreta: That's a really interesting approach! I quite like it :+1:

PS: I would add the `prev_release` key as an optional parameter just in case we want to use a different one than the default `-1`.

conf/plants/production_reg_conf.pl
```diff
@@ -35,35 +35,45 @@
...
+# Use our mirror (which has all the databases)
 Bio::EnsEMBL::Registry->load_registry_from_url("mysql://ensro\@mysql-ens-vertannot-staging:4573/$curr_release");
+
+# Or use the official staging servers
 #Bio::EnsEMBL::Registry->load_registry_from_url("mysql://ensro\@mysql-ens-sta-3:4160/$curr_release");
-#Bio::EnsEMBL::Registry->remove_DBAdaptor('saccharomyces_cerevisiae', 'core'); # never use EG's version of yeast
+# and remove the NV version of the shared species
+#Bio::EnsEMBL::Compara::Utils::Registry::remove_species(\@overlap_species);
+#Bio::EnsEMBL::Compara::Utils::Registry::remove_multi();
+# before loading the V version
```
carlacummins: What is 'the V version'?

carlacummins: Vertebrates version! Nvrmd 🤦‍♀️

muffato: I can make it clearer :+1:

muffato:
I would make it a new PR

conf/plants/production_reg_conf.pl
```diff
@@ -35,35 +35,45 @@
...
 # ---------------------- CURRENT CORE DATABASES----------------------------------
 
-# most cores are on EG servers, but some are on ensembl's vertannot-staging
+# Use our mirror (which has all the databases)
 Bio::EnsEMBL::Registry->load_registry_from_url("mysql://ensro\@mysql-ens-vertannot-staging:4573/$curr_release");
+
+# Or use the official staging servers
 #Bio::EnsEMBL::Registry->load_registry_from_url("mysql://ensro\@mysql-ens-sta-3:4160/$curr_release");
-#Bio::EnsEMBL::Registry->remove_DBAdaptor('saccharomyces_cerevisiae', 'core'); # never use EG's version of yeast
+# and remove the NV version of the shared species
```
muffato:
```suggestion
# and remove the Non-Vertebrates version of the shared species
```

conf/plants/production_reg_conf.pl
muffato:
```suggestion
# before loading the Vertebrates version
```

PR #176, opened 2020-02-05
Run stats after CAFE

JAlvarezJarreta:
## Description

In release 100, CAFE stats were missing from the `gene_member_hom_stats` table because the `stats_homologies` step was run before CAFE.

**Related JIRA tickets:**
- ENSCOMPARASW-3100

## Overview of changes
As Carla proposes in the ticket, to ensure CAFE is run before stats it can be moved to `rib_group_1` or stats can be moved to `rib_group_3`. I have done some numbers on those two groups for e100 for vertebrates, and `rib_group_1` took 2h 30m and `rib_group_3` took 8 seconds. Furthermore, CAFE rib took ~37h and the stats rib took ~14h. Thus, I have seen more logical to move CAFE to `rib_group_1` as the group is already in a "similar" time cost.

The new pipeline graph can be found [here](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-8&port=4618&dbname=jalvarez_plants_plants_protein_trees_100).

## Testing
There hasn't been any testing as the change is pretty straightforward and this rib doesn't have any dependencies with the analyses in `rib_group_1`.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/ProteinTrees_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/ProteinTrees_conf.pm
@@ -3267,6 +3267,7 @@ sub core_pipeline_analyses {
                 '1->A' => [
                     'rib_fire_gene_qc',
                     'rib_fire_homology_id_mapping',
+                    'rib_fire_cafe',
                 ],
                 'A->1' => 'rib_group_2'
             },
@@ -3277,7 +3278,6 @@ sub core_pipeline_analyses {
             -flow_into  => {
                 '1->A' => [
                     'rib_fire_dnds',
-                    'rib_fire_cafe',
                     'rib_fire_homology_stats',
                     'rib_fire_hmm_build',
                     'rib_fire_goc'


muffato:
This PR should be against release/100 because it's a bug that would affect anyone using the branch

JAlvarezJarreta:
Since it is a tiny change, I'm going to close this PR without applying it to `master` and commit the change to `release/100`. It seems easier than having to rebase the whole branch.

carlacummins:
But we didn't use this code in release 100. The point of the branches is to capture the exact code we used for the release.

muffato:
I think we haven't been following this exact rule. Regressions bugs are sometimes pushed to the release branch that long-standing bugs, which can often wait. The reason is that EG will be running pipelines on this branch.

If EG are running Cafe I think it should definitely be fixed on e100. If not, I guess we can live with the bug since it only affects a by-product of the pipeline which is only used for the website (the `gene_member_hom_stats` table) and won't affect any tests / benchmarks we might do against the e100 code ?

JAlvarezJarreta:
This change will be pushed with a commit to `release/100` branch instead.

PR #177, opened 2020-02-20
Pan compara needs more memory

thomasmaurel:
## Description

We need more memory for the pan division. 
Associated PR here: https://github.com/Ensembl/ensembl-production/pull/385

**Related JIRA tickets:**
- ENSCOMPARASW-XXXX

## Overview of changes
Increase memory for pan division.
#### Change #1
- _detail #1.1_

#### Change #2
- _detail #2.1_

## Testing
I ran the pipeline

## Notes
_Optional extra information._


muffato:
Sorry, didn't see your comment, @carlacummins

thomasmaurel:
> If this is expected to affect other divisions as well, then it is fine. If it has only to do with Pan, the standard is to create a division-specific pipeconfig file that tweaks these three analyses (see [Pan::ProteinTrees_conf](https://github.com/Ensembl/ensembl-compara/blob/release/100/modules/Bio/EnsEMBL/Compara/PipeConfig/Pan/ProteinTrees_conf.pm) for reference).

Some other divisions will benefit for more memory. Thanks for the example

thomasmaurel:
> It should be `1Gb_job` (note lower case `b`), but is fine otherwise

Thanks for letting me know

PR #180, opened 2020-03-03
NcRNA_trees homology mlss_id factory replaced

CristiGuijarro:
## Description

_The homology_dumps_mlss_id_factory was using analyses from the protein tree `Bio::EnsEMBL::Compara::PipeConfig::Parts::DumpHomologiesForPosttree`  which did not fit with the rest of the pipeline, causing conflict further along with the homologies that were dumped and the stats that depend on those dumps._

**Related JIRA tickets:**
- ENSCOMPARASW-3077

## Overview of changes

#### Change #180
- _The_ `homology_dumps_mlss_id_factory` _was replaced with_ `Bio::EnsEMBL::Compara::RunnableDB::MLSSIDFactory`

#### Change #180
- `Bio::EnsEMBL::Compara::RunnableDB::MLSSIDFactory` _was altered to _

## Testing
_This was tested in Standalone mode for different scenarios:_
```
- { 'methods' => {  'SYNTENY' => 2 },  'line_count' => 1 } 
- { 'methods' => {  'SYNTENY' => 2 } }
- { 'methods' => {  'ENSEMBL_ORTHOLOGUES' => 2 }, 'line_count' => 1 }
- { 'methods' => {  'ENSEMBL_ORTHOLOGUES' => 2 } }
- { 'methods' => {  'ENSEMBL_PARALOGUES' => 2 }, 'line_count' => 1 }
- { 'methods' => {  'ENSEMBL_PARALOGUES' => 2 } }
- { 'methods' => {  'LASTZ_NET'  => 2 }, 'line_count' => 1 }
```

## Notes
_There are 3 separate analyses for mlss_id factory in the_ `rib_fire_homology_dumps`. _There doesn't need to be. But in this case I have just swapped out the one analysis, so everything else remains the same._

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/MLSSIDFactory.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/MLSSIDFactory.pm
@@ -97,10 +97,9 @@ sub _get_line_count {
         $adaptor = $self->compara_dba->get_SyntenyRegionAdaptor;
     }
     else {
-        return;
+        die "_get_line_count does not support method: " . $mlss->method->type;
     }
-
-    my $allitems = $adaptor->fetch_all_by_MethodLinkSpeciesSet($mlss);
-    return scalar(@$allitems);
+    my $line_count = $adaptor->count_by_mlss_id($mlss->dbID);
+    return $line_count;
 }
 1;
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/MLSSIDFactory.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/MLSSIDFactory.pm
@@ -71,10 +71,36 @@ sub write_output {
     foreach my $method (keys %$methods) {
         my $branch_number = $methods->{$method};
         my $mlsss = $mlss_a->fetch_all_by_method_link_type($method);
-        foreach my $mlss (@$mlsss) {
-            $self->dataflow_output_id( { 'mlss_id' => $mlss->dbID }, $branch_number);
+
+        if ( $self->param('line_count') ) {
+            foreach my $mlss ( @$mlsss ) {
+                my $line_count = $self->_get_line_count($mlss);
+                $self->dataflow_output_id( { 'mlss_id' => $mlss->dbID, 'exp_line_count' => $line_count }, $branch_number);
+            }
+        }
+        else {
+            foreach my $mlss ( @$mlsss ) {
+                $self->dataflow_output_id( { 'mlss_id' => $mlss->dbID }, $branch_number);
+            }
         }
     }
 }
 
+sub _get_line_count {
+    my ($self, $mlss) = @_;
+    my $adaptor;
+
+    if ($mlss->method->class =~ /homology/i) {
+        $adaptor = $self->compara_dba->get_HomologyAdaptor;
+    }
+    elsif ($mlss->method->class =~ /synteny/i) {
+        $adaptor = $self->compara_dba->get_SyntenyRegionAdaptor;
+    }
+    else {
+        return;
+    }
+
+    my $allitems = $adaptor->fetch_all_by_MethodLinkSpeciesSet($mlss);
+    return scalar(@$allitems);
+}
 1;
--- a/modules/Bio/EnsEMBL/Compara/DBSQL/HomologyAdaptor.pm
+++ b/modules/Bio/EnsEMBL/Compara/DBSQL/HomologyAdaptor.pm
@@ -703,4 +703,11 @@ sub update_wga_coverage {
   return $self;
 }
 
+sub count_by_mlss_id {
+    my ($self, $mlss_id) = @_;
+
+    $self->bind_param_generic_fetch($mlss_id, SQL_INTEGER);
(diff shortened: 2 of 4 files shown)


muffato:
Can you explain something ? The change looks like you _need_ files to be created for _every_ mlss_id, even if there are no homologies and the file end up being empty (I guess it may not be strictly empty and still have the header line).
Where was it causing a problem in the rest of the ncRNA-tree pipeline ? Can the same problem affect the protein-tree pipeline as well, but it currently doesn't because the latter has data for every mlss_id ? If that's the case, I would suggest implementing your change in `homology_dumps_mlss_id_factory` in `PipeConfig/Parts/DumpHomologiesForPosttree.pm` instead of introducing `mlss_id_factory_for_homology_dumps`. Even if the protein-tree pipeline did create data for every mlss_id in default Vertebrates and Plants, can we really be sure it will always be the case for every division ? Let's not take the risk :smile: 

Then, I've got another question. `dump_per_mlss_homologies_tsv` (`RunnableDB::GeneTrees::DumpHomologiesTSV`) is configured to run the `line_count` healthcheck, which requires the `exp_line_count` parameter (should be using `param_required` line 134, by the way). Your new factory doesn't provide this parameter so you need to either populate it through another means or disable the healhcheck.

CristiGuijarro:
> Can you explain something ? The change looks like you _need_ files to be created for _every_ mlss_id, even if there are no homologies and the file end up being empty (I guess it may not be strictly empty and still have the header line).
> Where was it causing a problem in the rest of the ncRNA-tree pipeline ? 

The issue in the pipeline this release was that the homology dumps were not dumping all the necessary files that were needed further down the pipeline (in the stats). This appears to have been the issue you had last release running ncrna_trees also, as you hacked the same change (looking at the 99 pipeline).

> Can the same problem affect the protein-tree pipeline as well, but it currently doesn't because the latter has data for every mlss_id ? If that's the case, I would suggest implementing your change in `homology_dumps_mlss_id_factory` in `PipeConfig/Parts/DumpHomologiesForPosttree.pm` instead of introducing `mlss_id_factory_for_homology_dumps`. Even if the protein-tree pipeline did create data for every mlss_id in default Vertebrates and Plants, can we really be sure it will always be the case for every division ? Let's not take the risk 

I agree, because the protein_trees pipeline does do exactly the same further down the line with the dependent homology flatfile. But I was going for the change to make the ncrna_trees work because they didn't in the last couple of releases.

> Then, I've got another question. `dump_per_mlss_homologies_tsv` (`RunnableDB::GeneTrees::DumpHomologiesTSV`) is configured to run the `line_count` healthcheck, which requires the `exp_line_count` parameter (should be using `param_required` line 134, by the way). Your new factory doesn't provide this parameter so you need to either populate it through another means or disable the healhcheck.

Gotcha, I could stick reporoduce the line count parameter in the `Bio::EnsEMBL::Compara::RunnableDB::MLSSIDFactory`? And replace the exclusive analysis with this runnable to be available for both protein_trees and ncrna_trees?

modules/Bio/EnsEMBL/Compara/RunnableDB/MLSSIDFactory.pm
```diff
@@ -71,10 +71,36 @@ sub write_output {
...
+    my $adaptor;
+
+    if ($mlss->method->class =~ /homology/i) {
+        $adaptor = $self->compara_dba->get_HomologyAdaptor;
+    }
+    elsif ($mlss->method->class =~ /synteny/i) {
+        $adaptor = $self->compara_dba->get_SyntenyRegionAdaptor;
+    }
+    else {
+        return;
```
muffato: throw an error in this case

modules/Bio/EnsEMBL/Compara/RunnableDB/MLSSIDFactory.pm
```diff
@@ -71,10 +71,36 @@ sub write_output {
...
+    }
+
+    my $allitems = $adaptor->fetch_all_by_MethodLinkSpeciesSet($mlss);
+    return scalar(@$allitems);
```
muffato: Fetching the objects will be "slow" because it needs to transfer the data (rows) from the server to the Runnable and then instantiate all the objects. If you can count the objects directly in SQL and that would be much faster.

There is a standard way of doing that (introduced for `copy_data.pl`). Some adaptors implement a method named `count_by_mlss_id`. The standard way of implement this method is like in `SyntenyRegionAdaptor` or `ConstrainedElementAdaptor` or `GenomicAlignAdaptor`: you can use the parent class' `generic_count` method, which will read from the table that is set in `_tables`. If you implement the method in HomologyAdaptor, you can use it here, and any other method_link that needs a line_count would be added by implementing the method in its adaptor.

modules/Bio/EnsEMBL/Compara/RunnableDB/MLSSIDFactory.pm
CristiGuijarro: Can you give me an example?

carlacummins: `die "Unsupported method type";`

PR #181, opened 2020-03-03
New wrapper scripts that take a registry configuration file

muffato:
## Description

This is to to make our SOPs more generic, and be able to use our usual `production_reg_conf.pl` instead of having to provide the database name and server.

This removes the need to have separate instructions for Vertebrates (mysql-ens-compara-prod-1 ensembl_compara_XX), Plants (mysql-ens-compara-prod-5 ensembl_compara_plants_XX_YY) and Pan (given `$COMPARA_DIV` doesn't match)

## Overview of changes

The three new scripts are mostly wrappers around existing scripts (from other repos) that are not `production_reg_conf`-aware. 

`scripts/production/enable_mysql_keys.sh` is implemented differently. I've made it work with `db_cmd.pl`, e.g.
```
db_cmd.pl $COMPARA_REG compara_curr -executable $ENSEMBL_CVS_ROOT_DIR/ensembl-compara/scripts/production/enable_mysql_keys.sh
```

## Testing

I have tested each script by running it.

## Notes
_Optional extra information._

### DIFF
--- a/scripts/production/backup_master.pl
+++ b/scripts/production/backup_master.pl
@@ -41,9 +41,8 @@ URL of the database to dump.
 
 =item B<[--reg_conf registry_configuration_file]>
 
-The Bio::EnsEMBL::Registry configuration file. If none given and no URL is
-given, the one set in ENSEMBL_REGISTRY will be used if defined, if not
-~/.ensembl_init will be used.
+The Bio::EnsEMBL::Registry configuration file. Must be given to refer to
+the database by registry name (alias) instead of a URL.
 
 =item B<[--reg_type reg_type]>
 
--- a/scripts/production/backup_master.pl
+++ b/scripts/production/backup_master.pl
@@ -65,6 +65,12 @@ Defaults to "compara_master".
 
 Where to store the dumps. Defaults to the shared warehouse directory.
 
+=item B<[--username username]>
+
+Name of the user used to create the dumps.
+Defaults to "compara_ensembl". Set this to an empty string to create the
+backup as yourself.
+
 =item B<[--label str]>
 
 Label to add to the file name
@@ -85,7 +91,7 @@ use Bio::EnsEMBL::Compara::DBSQL::DBAdaptor;
 use Bio::EnsEMBL::Compara::Utils::Registry;
 
 my ($url, $reg_conf, $reg_type, $reg_alias);
-my ($dump_path, $label);
+my ($dump_path, $username, $label);
 
 # Arguments parsing
 GetOptions(
@@ -94,6 +100,7 @@ GetOptions(
     'reg_type=s'                    => \$reg_type,
     'reg_alias|regname|reg_name=s'  => \$reg_alias,
     'dump_path=s'                   => \$dump_path,
+    'username=s'                    => \$username,
     'label=s'                       => \$label,
 ) or die "Error in command line arguments\n";
 
@@ -121,11 +128,16 @@ my @params = (
     '--user'    => 'ensro',
 );
 
-$dump_path ||= '/nfs/production/panda/ensembl/warehouse/compara/master_db_dumps';
-
+$dump_path //= '/nfs/production/panda/ensembl/warehouse/compara/master_db_dumps';
+$username  //= 'compara_ensembl';
 my $date = strftime '%Y%m%d', localtime;
 my $dump_name = "ensembl_compara_master_${division}.${date}" . ($label ? ".$label" : ''). ".sql";
 
 my $cmd = join(' ', 'mysqldump', @params, $dba->dbc->dbname, '>', "$dump_path/$dump_name");
-print "Executing: $cmd\n\n";
-exec('sudo', -u => 'compara_ensembl', '/bin/bash', -c => $cmd);
+if ($username) {
+    print "Executing as $username: $cmd\n\n";
+    exec('sudo', -u => $username, '/bin/bash', -c => $cmd);
+} else {
+    print "Executing: $cmd\n\n";
+    exec('/bin/bash', -c => $cmd);
+}
--- a/scripts/production/backup_master.pl
+++ b/scripts/production/backup_master.pl
@@ -112,6 +112,11 @@ unless ($url or $reg_conf) {
     exit 1;
 }
 
+if ($url and $reg_alias) {
+    print "\nERROR: Both --url and --reg_alias are defined. Don't know which one to use\n\n";
+    exit 1;
+}
+
 if ($reg_conf) {
     Bio::EnsEMBL::Registry->load_all($reg_conf);
 }
--- a/scripts/production/backup_master.pl
+++ b/scripts/production/backup_master.pl
@@ -1,0 +1,133 @@
+#!/usr/bin/env perl
+# Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+# Copyright [2016-2020] EMBL-European Bioinformatics Institute
+# 
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# 
+#      http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+=head1 NAME
+
+backup_master.pl
+
+=head1 DESCRIPTION
+
+This script is a wrapper around sudo and mysqldump to dump the master database,
+taking its location from a production_reg_conf file.
+
+=head1 SYNOPSIS
+
+  perl backup_master.pl $COMPARA_REG compara_curr -group compara
+
+  perl backup_master.pl \
+    --reg_conf registry_configuration_file --reg_alias compara_curr
+
+=head1 ARGUMENTS
+
+The script reads these arguments and passes the other ones straight to run-configurable-testrunner.sh
+
+=head2 DATABASE SETUP
+
+=over
+
+=item B<[--url mysql://user[:passwd]@host[:port]/dbname]>
+
+URL of the database to test.
+
+=item B<[--reg_conf registry_configuration_file]>
+
+The Bio::EnsEMBL::Registry configuration file. If none given and no URL is
+given, the one set in ENSEMBL_REGISTRY will be used if defined, if not
+~/.ensembl_init will be used.
+
+=item B<[--reg_type reg_type]>
+
+The "type" or "group" under which the database is to be found in the Registry.
+
+=item B<[--reg_alias|--reg_name name]>
+
+The name or "species" under which the database is to be found in the Registry.
+Defaults to "compara_master".
+
+=back
+
+=head2 DUMP SETUP
+
+=over
+
+=item B<[--dump_path /path/to_dumps]>
+
+Where to store the dumps. Defaults to the shared warehouse directory.
+
+=item B<[--label str]>
+
+Label to add to the file name
+
+=back
+
+=cut
+
+use strict;
+use warnings;
+
+use Getopt::Long;
+use POSIX qw(strftime);
+
+use Bio::EnsEMBL::Registry;
+
+use Bio::EnsEMBL::Compara::DBSQL::DBAdaptor;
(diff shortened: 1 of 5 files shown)


scripts/production/backup_master.pl
```diff
@@ -0,0 +1,143 @@
...
+my @params = (
+    '--host'    => $dba->dbc->host,
+    '--port'    => $dba->dbc->port,
+    '--user'    => 'ensro',
+);
+
+$dump_path //= '/nfs/production/panda/ensembl/warehouse/compara/master_db_dumps';
+$username  //= 'compara_ensembl';
+my $date = strftime '%Y%m%d', localtime;
+my $dump_name = "ensembl_compara_master_${division}.${date}" . ($label ? ".$label" : ''). ".sql";
```
JAlvarezJarreta:
```suggestion
my $dump_name = "ensembl_${$reg_alias}_${division}.${date}" . ($label ? ".$label" : '') . ".sql";
```

scripts/production/backup_master.pl
```diff
@@ -0,0 +1,143 @@
...
+=head1 DESCRIPTION
+
+This script is a wrapper around sudo and mysqldump to dump the master database,
+taking its location from a production_reg_conf file.
+
+=head1 SYNOPSIS
+
+  perl backup_master.pl --reg_conf $COMPARA_REG_PATH
+
+  perl backup_master.pl $COMPARA_REG compara_master --label "pre100"
```
JAlvarezJarreta:
```suggestion
    perl backup_master.pl --reg_conf $COMPARA_REG_PATH

    perl backup_master.pl $COMPARA_REG compara_master --label "pre100"
```

scripts/production/patch_database.pl
```diff
@@ -0,0 +1,128 @@
...
+
+=head1 ARGUMENTS
+
+=head2 DATABASE SETUP
+
+=over
+
+=item B<[--url mysql://user[:passwd]@host[:port]/dbname]>
+
+URL of the database to test.
```
JAlvarezJarreta:
```suggestion
URL of the database to patch.
```

scripts/production/patch_database.pl
```diff
@@ -0,0 +1,128 @@
...
+    'reg_alias|regname|reg_name=s'  => \$reg_alias,
+    'schema_patcher=s'              => \$schema_patcher,
+) or die "Error in command line arguments\n";
+
+if (@ARGV) {
+    die "ERROR: There are invalid arguments on the command-line: ". join(" ", @ARGV). "\n";
+}
+
+unless ($url or ($reg_conf and $reg_alias)) {
+    print "\nNeither --url nor --reg_conf and --reg_alias are defined. Some of those are needed to refer to the database being tested\n\n";
```
JAlvarezJarreta:
```suggestion
    print "\nNeither --url nor --reg_conf and --reg_alias are defined. Some of those are needed to refer to the database being patched\n\n";
```

scripts/production/backup_master.pl
```diff
@@ -0,0 +1,143 @@
...
+    'username=s'                    => \$username,
+    'label=s'                       => \$label,
+) or die "Error in command line arguments\n";
+
+if (@ARGV) {
+    die "ERROR: There are invalid arguments on the command-line: ". join(" ", @ARGV). "\n";
+}
+
+unless ($url or $reg_conf) {
+    print "\nNeither --url nor --reg_conf are defined. Some of those are needed to refer to the database being tested\n\n";
```
JAlvarezJarreta:
```suggestion
    print "\nNeither --url nor --reg_conf are defined. Some of those are needed to refer to the database being dumped\n\n";
```

scripts/production/patch_database.pl
```diff
@@ -0,0 +1,128 @@
...
+
+=item B<[--reg_conf registry_configuration_file]>
+
+The Bio::EnsEMBL::Registry configuration file. If none given and no URL is
+given, the one set in ENSEMBL_REGISTRY will be used if defined, if not
+~/.ensembl_init will be used.
+
+=item B<[--reg_type reg_type]>
+
+The "type" or "group" under which the database is to be found in the Registry.
```
JAlvarezJarreta:
```suggestion
The "type" or "group" under which the database is to be found in the Registry.
Defaults to "compara".
```

scripts/production/patch_database.pl
```diff
@@ -0,0 +1,128 @@
...
+unless ($schema_patcher) {
+    die "Need to give the --schema_patcher option or set the ENSEMBL_CVS_ROOT_DIR environment variable to use the default" unless $ENV{ENSEMBL_CVS_ROOT_DIR};
+    $schema_patcher = $ENV{ENSEMBL_CVS_ROOT_DIR} . '/ensembl/misc-scripts/schema_patcher.pl';
+}
+die "'$schema_patcher' is not a valid executable" unless -x $schema_patcher;
+
+
+if ($dba->dbc->user eq 'ensro') {
+    warn "Switching to the ensadmin user\n";
+}
```
JAlvarezJarreta: Don't we usually follow the following convention?
```suggestion
warn "Switching to the ensadmin user\n" if ($dba->dbc->user eq 'ensro');
```

scripts/production/patch_database.pl
```diff
@@ -0,0 +1,128 @@
...
+
+my @params = (
+    '--host'    => $dba->dbc->host,
+    '--port'    => $dba->dbc->port,
+    '--user'    => Bio::EnsEMBL::Compara::Utils::Registry::get_rw_user($dba->dbc->host),
+    '--pass'    => Bio::EnsEMBL::Compara::Utils::Registry::get_rw_pass($dba->dbc->host),
+    '--database'    => $dba->dbc->dbname,
+);
```
JAlvarezJarreta:
```suggestion
my @params = (
    '--host'     => $dba->dbc->host,
    '--port'     => $dba->dbc->port,
    '--user'     => Bio::EnsEMBL::Compara::Utils::Registry::get_rw_user($dba->dbc->host),
    '--pass'     => Bio::EnsEMBL::Compara::Utils::Registry::get_rw_pass($dba->dbc->host),
    '--database' => $dba->dbc->dbname,
);
```

scripts/production/run_datachecks.pl
```diff
@@ -0,0 +1,158 @@
...
+accepts our production_reg_conf-based configuration and automatically sets
+the connection details of the database being tested and the previous one.
+
+=head1 SYNOPSIS
+
+  perl run_datachecks.pl $COMPARA_REG compara_curr -group compara
+
+  perl run_datachecks.pl \
+    --reg_conf registry_configuration_file --reg_alias compara_curr \
+    -name CompareMSANames
```
JAlvarezJarreta:
```suggestion
    perl run_datachecks.pl $COMPARA_REG compara_curr --group compara

    perl run_datachecks.pl \
        --reg_conf registry_configuration_file --reg_alias compara_curr \
        --name CompareMSANames
```

scripts/production/run_datachecks.pl
```diff
@@ -0,0 +1,158 @@
...
+    $dc_runner = $ENV{ENSEMBL_CVS_ROOT_DIR} . '/ensembl-datacheck/scripts/run_datachecks.pl';
+}
+die "'$dc_runner' is not a valid executable" unless -x $dc_runner;
+
+if ($reg_conf) {
+    Bio::EnsEMBL::Registry->load_all($reg_conf);
+}
+my $dba = $reg_alias
+    ? Bio::EnsEMBL::Registry->get_DBAdaptor( $reg_alias, $reg_type || 'compara' )
+    : Bio::EnsEMBL::Compara::DBSQL::DBAdaptor->new( -URL => $url );
```
JAlvarezJarreta: I must confess I have found it interesting how in the previous two scripts you were testing for `$url` and in this one you have gone for `$reg_alias` :nerd_face: Thinking about it more thoroughly, do we want `$url` be the main parameter (in case both are passed)? Should we set a consensus in our scripts?

scripts/production/run_datachecks.pl
```diff
@@ -0,0 +1,158 @@
...
+
+# Common parameters
+my @params = (
+    '--host'    => $dba->dbc->host,
+    '--port'    => $dba->dbc->port,
+    '--user'    => 'ensro',             # Fallback to ensro to ensure we don't write to it by accident
+    '--dbname'  => $dba->dbc->dbname,
```
JAlvarezJarreta:
```suggestion
    '--host'   => $dba->dbc->host,
    '--port'   => $dba->dbc->port,
    '--user'   => 'ensro',             # Fallback to ensro to ensure we don't write to it by accident
    '--dbname' => $dba->dbc->dbname,
```

scripts/production/run_datachecks.pl
```diff
@@ -0,0 +1,158 @@
...
+);
+
+if ($reg_conf) {
+    push @params, (
+        '--registry_file'   => $reg_conf,
```
JAlvarezJarreta:
```suggestion
        '--registry_file' => $reg_conf,
```

scripts/production/backup_master.pl
```diff
@@ -0,0 +1,143 @@
...
+
+=item B<[--url mysql://user[:passwd]@host[:port]/dbname]>
+
+URL of the database to test.
+
+=item B<[--reg_conf registry_configuration_file]>
+
+The Bio::EnsEMBL::Registry configuration file. If none given and no URL is
+given, the one set in ENSEMBL_REGISTRY will be used if defined, if not
+~/.ensembl_init will be used.
```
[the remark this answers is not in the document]
muffato: 1. The behaviour explained here (`$ENSEMBL_REGISTRY` and `~/.ensembl_init`) is defined in the Registy itself. We can't change it
2. The script actually forbids being run without a URL or a reg_conf. So we will never fall back to the Registry's default behaviour. I should change this documentation

muffato:
Something to consider as well: when we start using the `run_datachecks.pl` wrapper, since it's going to set the `old_server_uri` automatically, we don't need to have mentions of `compara_prev` in the DC any more, e.g. https://github.com/Ensembl/ensembl-datacheck/blob/release/100/lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm#L44-L45
`compara_prev` is a Compara-only convention, and I'd be happy to keep it away from the DCs if possible.
And by the way, James is going to fix `$self->get_old_dba` so that it works even if we don't set `old_server_uri`, cf https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-4976

scripts/production/backup_master.pl
muffato: FYI 1 space is enough to trigger a "verbatim" paragraph: https://perldoc.perl.org/perlpod.html#Verbatim-Paragraph

scripts/production/patch_database.pl
muffato: FYI I use "Tab" to align the `=>` signs, so they end up align to _4*n_ columns, and there could be more than 1 whitespace between the end of the longest string and `=>`. I also often make multiple levels of alignments depending of the length of the parameters, like here 1 level for `host`, port`, etc, and another level for `database`

scripts/production/patch_database.pl
muffato: I wasn't aware this was a convention ...

scripts/production/run_datachecks.pl
carlacummins: Functionality is not affected. I don't think we need a consensus.

JAlvarezJarreta: I must confess I'm slightly concerned that we will have two scripts that, if you pass both `$reg_alias` and `$url` (by mistake, I assume), they will behave differently. But then again, only one should be passed, so we can ignore my previous comment.

muffato: Right, that's an understandable concern. I think it should be addressed by making the script die if both are set

scripts/production/backup_master.pl
```diff
@@ -0,0 +1,143 @@
...
+
+=item B<[--dump_path /path/to_dumps]>
+
+Where to store the dumps. Defaults to the shared warehouse directory.
+
+=item B<[--username username]>
+
+Name of the user used to create the dumps.
+Defaults to "compara_ensembl". Set this to an empty string to create the
+backup as yourself.
```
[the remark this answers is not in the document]
muffato: I've chosen the default values (in this script and the others) that are the closest to what we need in production in order to simplify the command-lines.

scripts/production/backup_master.pl
muffato: I am not expecting this script to be used for master databases owned by individuals, and in this case both `--username` and `--dump_path` would have to be given on the command line because the default values wouldn't be sensible.

PR #184, opened 2020-03-19
Dnafrag loading

muffato:
## Description

The metadata service considers all seq_region (dnafrag) related changes in its "updated_assemblies" category, meaning that we get there changes of the primary assembly (e.g. new macaque MT) but also new LRGs and new patches.

Human get new LRGs almost every release, meaning that the PrepareMasterDatabase pipeline will run update_genome for human, which will fail and require the "force" option. Once enabled, the analysis that specifically loads LRGs is pointless since they will have already been loaded. Similar problem with patches.

**Related JIRA tickets:**
- ENSCOMPARASW-3283

#### Only update dnafrags in the standard `add_species_to_master` analysis

No need to load LRGs and assembly patches in separate analyses

#### Report which dnafrags would change when requesting the force option

Then we can see if the changes are about patches, LRGs or main regions. This is useful to double-check what we expect of the assembly change and to make sure we don't bring in unscheduled changes.

#### Assembly patch reports.

I moved the analysis that creates the reports about new patches _before_ the dnafrags get updated. I turned it back to was it initially was: a factory on a list of species names. This is because we only care about assembly patches in the Vertebrates division

## Testing

Being tested on a backup copy of the pre-e100 master database


muffato:
Just pushed another commit I had missed about the assembly patch reports. A minor bugfix: do not report missing dnafrags / missing slices as differing lengths

modules/Bio/EnsEMBL/Compara/Utils/MasterDatabase.pm
```diff
@@ -313,9 +313,24 @@ sub _update_genome_db {
...
+        # dnafrags_match_core_slices uses print. This is to capture the output
+        local *STDOUT;
+        open (STDOUT, '>', \$output);
+        $match = dnafrags_match_core_slices($compara_dba, $genome_db);
+      }
+      my $msg = "\n\n** GenomeDB with this name [$species_production_name] and assembly".
+        " [$this_assembly] is already in the compara DB **\n";
+      if ($match) {
+        $msg .= "** And it has the right set of DnaFrags **\n";
+        $msg .= "** You can use the --force option to update the other GenomeDB fields IF YOU REALLY NEED!! **\n\n";
```
JAlvarezJarreta:
```suggestion
        $msg .= "** You can use the --force option to update the other GenomeDB fields IF YOU REALLY NEED TO!! **\n\n";
```

PR #185, opened 2020-03-19
Smart reusability detection

muffato:
## Description

Reusability is assessed on _all_ members even though in some divisions we only care about coding genes on the primary assembly. Here I add the required flags to select what needs to be assessed.

**Related JIRA tickets:**
- ENSCOMPARASW-3284

## Overview of changes

All flags but `include_lrg` were already present in the pipeline, but only passed to `load_members`. They are now passed to `check_reusability` as well

## Testing
_How was this tested? Have new unit tests been included?_

Tested as standaloneJob with custom connection parameters. I could make it mark mouse as either reusable or non-reusable depending on what biotypes etc were requested.

## Notes

The default parameters of CheckMembersReusability are such that it doesn't need a specific configuration in the MercatorPecan pipeline (the only other pipeline where it's used).

:warning: Note that the base branch is not `release/101` but `feature/dnafrag_loading`. This is because some commits here depend on commits in the latter, and to simplify the review.


muffato:
For the record (I forgot to mention it), I had already tested in standaloneJob mode the two runnables that I changed (the one that lists the assembly patches and the one that updates a genome_db)

PR #168, opened 2019-12-16
Update LastZ to avoid connections to core databases as much as possible

JAlvarezJarreta:
Ticket [ENSCOMPARASW-1727](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-1727).

Made LastZ pipeline to use our Genome Dumps instead of retrieving the information from core databases. There are some other additions that I have summarised in the following list:
* Removed storage analyses that used to store sequences from core databases to the pipeline's database
* All dumps (`DnaFragChunk`, `DnaFragChunkSet` and nibs) are generated locally instead of dumping them in a shared location (`$pipeline_dir` @ `/hps`): the cost of the accessing these large files through the network was higher than generating some of them several times
* Updates some old `wait` parts of the pipeline with semaphores (only the simple/easy ones)
* Locators on `genome_db` table have to be generated for both principal and component genomes so the stats will work properly (in case the registry is missing some core databases)
* The PODs and documentation have been updated according in almost every file I have explored

I created a testing division called `sprouts` to add the polyploid handling to LastZ pipeline and I have used the same one to test the improvements included in this PR.

Old pipeline: [jalvarez_sprouts_lastz](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-8&port=4618&dbname=jalvarez_sprouts_lastz)
New pipeline: [jalvarez_sprouts_lastz_v3](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-8&port=4618&dbname=jalvarez_sprouts_lastz_v3)

Overall, the new version takes a total of 2 hours and 20 minutes less (in terms of the sum of all job runtimes) than the old pipeline (about a 19% less time cost). Furthermore, after the 3rd analysis the registry can have the core databases removed and the pipeline will work as expected (proof that it is using the Genome Dumps).

Finally, I have removed the parameter `dump_loc` in every example at https://www.ebi.ac.uk/seqdb/confluence/display/EnsCom/Haplotypes+and+assembly+patches+against+other+species to match this update (can be reverted if we decide to leave this bit at the end).

[...]

PR #186, opened 2020-03-19
Remove core dependencies in LastZ pipeline

JAlvarezJarreta:
## Description

Related with previous work on PR#168, we are trying to remove all the dependencies (that can be removed) in this pipeline from Core dbs, using our genome dumps instead. @muffato detected a few more cases (mentioned [here](https://github.com/Ensembl/ensembl-compara/pull/168#discussion_r358552815)) that have been now removed.

**Related JIRA tickets:**
- ENSCOMPARASW-3302

## Overview of changes
Removed the core dependencies in `alignment_chains` analysis. For that I have used the `DnaFragChunk` class (instead of just Locus) to also take advantage of the method `dump_chunks_to_fasta_file()`, which saves a lot of memory requirements without affecting its performance significantly (in my test there was a ~5% increment).

A few more minor details have been addressed too.

## Testing
I have generated two LastZ pipelines for CITest, one with the previous code (mysql://ensro@mysql-ens-compara-prod-10:4648/jalvarez_citest_lastz_default) and one with the new one (mysql://ensro@mysql-ens-compara-prod-10:4648/jalvarez_citest_lastz_update). Before running `alignment_chains` analysis, I removed the core dbs from the registry, and for the updated code I also removed the locators from the `genome_db` table. All jobs have finished succesfully. I manually checked one nib file from the same job and both pipelines generated the same `.fa` file.

## Notes
The `genomic_align` and `genomic_align_block` tables present differences, but I'm not sure if this is due to the update or the reproducibility of this analysis.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/Production/Analysis/AlignmentChains.pm
+++ b/modules/Bio/EnsEMBL/Compara/Production/Analysis/AlignmentChains.pm
@@ -79,9 +79,8 @@ sub run_chains {
     mkdir $target_nib_dir;
 
     my $target_chunks = $self->param('target_chunks');
-    foreach my $nm (keys %{$target_chunks}) {
-      my $target = $target_chunks->{$nm};
-      my $target_name = $target->dnafrag->name;
+    foreach my $target_name (keys %{$target_chunks}) {
+      my $target = $target_chunks->{$target_name};
 
       $target->dump_chunks_to_fasta_file("$target_nib_dir/$target_name.fa");
 
--- a/modules/Bio/EnsEMBL/Compara/Production/Analysis/AlignmentChains.pm
+++ b/modules/Bio/EnsEMBL/Compara/Production/Analysis/AlignmentChains.pm
@@ -26,8 +26,6 @@ package Bio::EnsEMBL::Compara::Production::Analysis::AlignmentChains;
 use warnings ;
 use strict;
 
-use Bio::SeqIO;
-
 use Bio::EnsEMBL::Utils::Exception qw(throw);
 use Bio::EnsEMBL::DnaDnaAlignFeature;
 
@@ -68,18 +66,7 @@ sub run_chains {
     $query_nib_dir = "$work_dir/query_nib";
     mkdir $query_nib_dir;
 
-    my $seqio = Bio::SeqIO->new(-format => 'fasta',
-                                -file   => ">$query_nib_dir/$query_name.fa");   
-
-    # prevent extensive disconnections when fetching sequence length etc.
-    my $query_slice = $self->param('query_slice');
-    $query_slice->adaptor()->dbc->prevent_disconnect( sub {
-
-    $seqio->write_seq($query_slice); 
-
-    } );
-
-    $seqio->close;
+    $self->param('query_chunk')->dump_chunks_to_fasta_file("$query_nib_dir/$query_name.fa");
     
     $self->run_command([$self->param_required('faToNib_exe'), "$query_nib_dir/$query_name.fa", "$query_nib_dir/$query_name.nib"], { die_on_failure => 1 });
     push @nib_files, "$query_nib_dir/$query_name.nib";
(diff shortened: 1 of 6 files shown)


modules/Bio/EnsEMBL/Compara/Locus.pm
```diff
@@ -433,7 +433,7 @@ sub get_sequence {
         # Coordinates are 0-based
         $seq = $faidx_helper->get_sequence2_no_length($self->dnafrag_id, $self->dnafrag_start-1, $self->dnafrag_end-1);
         die "sequence length doesn't match !" if CORE::length($seq) != ($self->dnafrag_end-$self->dnafrag_start+1);
-        reverse_comp(\$seq) if $self->dnafrag_strand < 0;
+        reverse_comp(\$seq) if (defined $self->dnafrag_strand) and ($self->dnafrag_strand < 0);
```
muffato: You had this change in an early version of #168 and @CristiGuijarro asked why it was needed. So as far as I understand, it's because DnaFragChunk doesn't have the concept of strand ? An alternative is to remove the ambiguity in DnaFragChunk itself and set `dnafrag_strand` to 1 in its constructor.

modules/Bio/EnsEMBL/Compara/Production/Analysis/AlignmentChains.pm
```diff
@@ -91,16 +78,13 @@ sub run_chains {
...
-      
-      my $seqio =  Bio::SeqIO->new(-format => 'fasta',
-                                -file   => ">$target_nib_dir/$target_name.fa");
-      $seqio->write_seq($target);
-      $seqio->close; 
-      
+    my $target_chunks = $self->param('target_chunks');
+    foreach my $nm (keys %{$target_chunks}) {
+      my $target = $target_chunks->{$nm};
+      my $target_name = $target->dnafrag->name;
```
carlacummins: Isn't `$target_name` the same as `$nm`? Why defined it twice? 

```suggestion
    foreach my $target_name (keys %{$target_chunks}) {
      my $target = $target_chunks->{$target_name};
```

modules/Bio/EnsEMBL/Compara/RunnableDB/PairAligner/ChunkAndGroupDna.pm
```diff
@@ -135,7 +135,6 @@ sub create_chunks {
     } else {
         my $dnafrag_list = $dnafrag_dba->fetch_all_by_GenomeDB(
             $genome_db,
-            -COORD_SYSTEM_NAME  => 'toplevel',
```
carlacummins: Is it safe to remove this? I know that newer cores are coming with the `primary_assembly` coord system, but what if we run the pipeline with an older core?

modules/Bio/EnsEMBL/Compara/Production/Analysis/AlignmentChains.pm
JAlvarezJarreta: I have tried to leave as much of the original code as possible, but you are right that before these two weren't the same and now they are. Well spotted!

modules/Bio/EnsEMBL/Compara/RunnableDB/PairAligner/ChunkAndGroupDna.pm
JAlvarezJarreta: This PR is tricky because it comes from comments in a closed PR. Here is the comment from @muffato that made me change this bit: https://github.com/Ensembl/ensembl-compara/pull/168#discussion_r384217116

muffato: `-COORD_SYSTEM_NAME  => 'toplevel' is ignored by `fetch_all_by_GenomeDB`, cf https://github.com/Ensembl/ensembl-compara/blob/release/99/modules/Bio/EnsEMBL/Compara/DBSQL/DnaFragAdaptor.pm#L223

PR #190, opened 2020-05-04
Add '-wait_for' instructions in lastz pipeline to fix flow bugs

CristiGuijarro:
## Description

The LASTZ pipeline is running things out of order. This is because the semaphores and `wait_for` s in the LASTZ pipeline are not set up correctly. Here I add a few more `wait_for` s to solve the issue.

**Related JIRA tickets:**
- ENSCOMPARASW-3471

#### Change 1
- `create_filter_duplicates_jobs` now waiting for both `create_pair_aligner_jobs` and `check_no_partial_gabs` and `lastz` 

#### Change 2
- `create_alignment_nets_jobs` now waiting for both `create_alignment_chains_jobs` and `remove_inconsistencies_after_chain`

## Testing
_[Pipeline here.](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-7&port=4617&dbname=cristig_lastz_wait_for_tests)_

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/PairAligner_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/PairAligner_conf.pm
@@ -353,7 +353,7 @@ sub core_pipeline_analyses {
  	    },
  	    {  -logic_name => 'create_filter_duplicates_jobs', #factory
  	       -module     => 'Bio::EnsEMBL::Compara::RunnableDB::PairAligner::CreateFilterDuplicatesJobs',
- 	       -wait_for =>  [ 'update_max_alignment_length_before_FD' ],
+ 	       -wait_for =>  [ 'update_max_alignment_length_before_FD', 'check_no_partial_gabs', 'create_pair_aligner_jobs', $self->o('pair_aligner_logic_name') ],
 	        -flow_into => {
 			       2 => { 'filter_duplicates' => INPUT_PLUS() },
 			     },
@@ -469,7 +469,7 @@ sub core_pipeline_analyses {
 			       1 => [ 'remove_inconsistencies_after_net' ],
 			       2 => [ 'alignment_nets' ],
 			      },
-            -wait_for => [ 'update_max_alignment_length_after_chain' ],
+            -wait_for => [ 'update_max_alignment_length_after_chain', 'create_alignment_chains_jobs', 'remove_inconsistencies_after_chain' ],
 	       -rc_name => '1Gb_job',
  	    },
  	    {  -logic_name => 'alignment_nets',
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/PairAligner_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/PairAligner_conf.pm
@@ -469,7 +469,7 @@ sub core_pipeline_analyses {
 			       1 => [ 'remove_inconsistencies_after_net' ],
 			       2 => [ 'alignment_nets' ],
 			      },
-            -wait_for => [ 'update_max_alignment_length_after_chain', 'create_alignment_chains_jobs', 'remove_inconsistencies_after_chain' ],
+            -wait_for => [ 'update_max_alignment_length_after_chain', 'create_alignment_chains_jobs', 'remove_inconsistencies_after_chain', 'alignment_chains' ],
 	       -rc_name => '1Gb_job',
  	    },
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/Compara/PipeConfig/PairAligner_conf.pm
```diff
@@ -353,7 +353,7 @@ sub core_pipeline_analyses {
  	    },
  	    {  -logic_name => 'create_filter_duplicates_jobs', #factory
  	       -module     => 'Bio::EnsEMBL::Compara::RunnableDB::PairAligner::CreateFilterDuplicatesJobs',
- 	       -wait_for =>  [ 'update_max_alignment_length_before_FD' ],
+ 	       -wait_for =>  [ 'update_max_alignment_length_before_FD', 'check_no_partial_gabs', $self->o('pair_aligner_logic_name') ],
```
JAlvarezJarreta, line 356: What you propose could release the blocked jobs if a job in `LastZ` fails due to MEMLIMIT (it will be marked as done in `LastZ`). You need to either add `LastZ_himem` as well or, alternatively, make it wait for the factory, as no jobs in `check_no_partial_gabs` will be marked as DONE before all `LastZ` and `LastZ_himem` have completed anyway:
```suggestion
 	       -wait_for =>  [ 'update_max_alignment_length_before_FD', 'check_no_partial_gabs', 'create_pair_aligner_jobs' ],
```

CristiGuijarro: That didn't work. 
I ran the pipeline 3 times. The first time with:
`-wait_for =>  [ 'update_max_alignment_length_before_FD', 'check_no_partial_gabs', 'create_pair_aligner_jobs' ],`
the second time with:
`-wait_for =>  [ 'update_max_alignment_length_before_FD', 'check_no_partial_gabs', 'create_pair_aligner_jobs', $self->o('pair_aligner_logic_name') ],`
and the third with:
`-wait_for =>  [ 'update_max_alignment_length_before_FD', 'check_no_partial_gabs', $self->o('pair_aligner_logic_name') ],`
The last two worked. The first didn't.
I can revert back to:
`-wait_for =>  [ 'update_max_alignment_length_before_FD', 'check_no_partial_gabs', 'create_pair_aligner_jobs', $self->o('pair_aligner_logic_name') ],` if @carlacummins agrees, because this worked.

carlacummins: The case you're outlining is almost impossible to occur since `create_pair_aligner_jobs` spends the majority of its runtime in `WRITE_OUTPUT`, producing new `LastZ` jobs. But if you really want to account for this extreme edge case, then sure - go with the final option outlined.

carlacummins: In fact, the suggested change from @JAlvarezJarreta would actually be a step back IMO, since it's removing the `LastZ` blocker. The case where there are only completed jobs in `check_no_partial_gabs` and `create_pair_aligner_jobs` is a much more likely event and is one that I personally have seen already in e101

JAlvarezJarreta: So I wonder why the first solution doesn't work, and the second does, when, unless I'm missing something, they are exactly the same factory-fan-funnel scenario but we are applying two different `wait for` solutions (my proposal for the first one is exactly the same as the second one done).

modules/Bio/EnsEMBL/Compara/PipeConfig/PairAligner_conf.pm
```diff
@@ -469,7 +469,7 @@ sub core_pipeline_analyses {
 			       1 => [ 'remove_inconsistencies_after_net' ],
 			       2 => [ 'alignment_nets' ],
 			      },
-            -wait_for => [ 'update_max_alignment_length_after_chain' ],
+            -wait_for => [ 'update_max_alignment_length_after_chain', 'create_alignment_chains_jobs', 'remove_inconsistencies_after_chain' ],
```
JAlvarezJarreta, line 472: Since this solution hasn't worked in the first case, I think it would be safer to do the same here (just in case):
```suggestion
            -wait_for => [ 'update_max_alignment_length_after_chain', 'create_alignment_chains_jobs', 'remove_inconsistencies_after_chain' , 'alignment_chains'],
```

PR #191, opened 2020-05-12
Feature/seq loc attrib

CristiGuijarro:
## Description

MT and PT sequences are now to be identified through a sequence attribute named sequence_location instead of name / synonyms.

**Related JIRA tickets:**
- ENSCOMPARASW-3468

## Overview of changes
Instead of mapping `seq_region_name` to either `NUC`, `MT` or `PT` and `synonym` if that doesn't match, we check the `sequence_location` attrib `value`.

#### Change 1
- Map the possible enum `cellular_component` values to the `sequence_location` attribute. If an attrib `value` for `sequence_location` is not defined, then the default is to assign `NUC`.

#### Change 2
-  Currently with only `MT`, `PT` and `NUC` in our schema, as a possible `cellular_component` , we need to add `OTHER` as an enum value to the column for the schema, until we know what we will do with the other organelles from the other divisions in the future. Whilst technically `PT` could account for all plastids, in this case it only maps to `chloroplast_chromosome` and other plastids will be assigned `OTHER`.

## Testing
Unit tests for dnafrag.pm and MasterDatabase.pm prior to Travis, then Travis

## Notes
We may want to include all possible organelles as unique `cellular_components` in the future and swap out `OTHER`. `OTHER` is the quick fix for the uncertain future.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/DnaFrag.pm
+++ b/modules/Bio/EnsEMBL/Compara/DnaFrag.pm
@@ -223,7 +223,6 @@ sub new_from_Slice {
         'genome_db_id' => $genome_db->dbID,
         '_codon_table_id' => $codon_table_id || 1,
         '_cellular_component' => $cellular_component,
-        '_sequence_location' => $sequence_location,
     } );
 }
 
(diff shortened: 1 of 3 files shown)


modules/Bio/EnsEMBL/Compara/DnaFrag.pm
```diff
@@ -199,17 +199,19 @@ sub new_from_Slice {
...
-            if (exists $name_to_cellular_component{uc $synonym->name}) {
-                $cellular_component = $name_to_cellular_component{uc $synonym->name};
-                last;
-            }
+
+    if ($sequence_location) {
+        if (exists $seq_loc_to_cell_component{$sequence_location}) {
+            $cellular_component = $seq_loc_to_cell_component{$sequence_location};
+        }
+        else {
```
JAlvarezJarreta, line 213:
```suggestion
        } else {
```

PR #195, opened 2020-06-03
Remove genomes from MSAs (part 1)

JAlvarezJarreta:
## Description

This PR introduces changes on certain runnables and pipelines that will be used later on in the new pipeline.

**Related JIRA tickets:**
- ENSCOMPARASW-3065

#### Change 1
- The ancestral db creation runnables used in the EPO pipelines has been moved to its own "Parts" now, so it can be directly reused in the pipeline I'm working on.
- I have removed the resource class of the last 2 runnables as for e101 they have required less than 200Mb for any EPO.

#### Change 2
- Added one new flag to `populate_new_database.pl` script to avoid copying the GERPs (it is safer/easier to recompute these than copying and updating them).
- Added this new flag and a preexisting one to `PopulateNewDatabase.pm` runnable so they can be used in pipelines.

#### Change 3
- Made `Gerp` runnable to use only the genomes of the MLSS's species set, instead of all the genomes in the database.

## Testing
I'm using this code in the new pipeline and everything is working as expected.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/Parts/EPOAlignment.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/Parts/EPOAlignment.pm
@@ -40,6 +40,7 @@ use warnings;
 
 use Bio::EnsEMBL::Hive::PipeConfig::HiveGeneric_conf;           # Allow this particular config to use conditional dataflow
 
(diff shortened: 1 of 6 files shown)


JAlvarezJarreta:
I have just rebased the branch, but nothing has changed in this PR.

muffato (approved):

PR #196, opened 2020-06-05
Embed the WGA Orthology QC and HighConfidence pipelines into the gene-tree pipelines

CristiGuijarro:
## Description

_To further streamline the inference of ortholog quality and confidence metrics._

**Related JIRA tickets:**
- ENSCOMPARASW-3257

## Overview of changes
Parameters needed to be migrated to the protein trees and ncrna trees pipelines and the thresholds for the division specific pipelines in order to incorporate the wga and high confidence pipelines. The High confidence orthologs pipeline only needed to be included in both ncrna and protein tree pipelines. A new `Pipeconfig/Parts/` config file was written to adapt the OrthologQMAlignment pipeline. 

#### Change 1
- New pipeconfig Part: OrthologQMAlignment to be used in both ncrna and protein tree pipelines. This now populates new parameter when complete to prevent HighConfidenceOrthologs from starting if semaphore is released before all pairwise and multiple alignments have completed.

#### Change 2
- All parameters and thresholds for both HighConfidenceOrthologs and OrthologQMAlignment included into the protein trees and ncrna pipeline config files. New parameter to turn OrthologQMAlignment on or off when all pairwise and multiple alignments have finished. 

#### Change 3
- New dummy runnables included to check for whether to run the wga analyses or high confidence analyses.

#### Change 4
- Create table command for the `ortholog_quality` tables in both ncrna and protein tree pipelines.

#### Change 5
- Ignore table parameters for mergedb pipeline to ignore `ortholog_quality` table.

#### Change 6
- Make change to standalone version of WGA pipeline so that it uses new `/Parts/OrthologQMAlignment.pm`

## Testing
Full pipeline runs with both protein trees and ncrna trees:
-- `mysql://ensro@mysql-ens-compara-prod-4:4401/cristig_vertebrates_nctrees_wga_hc_test`
-- `mysql://ensro@mysql-ens-compara-prod-5:4615/cristig_plants_ptrees_wga_hc_test`
-- `mysql://ensro@mysql-ens-compara-prod-5:4615/cristig_wga_standalone_test`
(ncrna trees was used for production e101)

PR #198, opened 2020-06-09
Feature/dependency graph

JAlvarezJarreta:
## Description

With the latest update from @CristiGuijarro (see [PR#196](https://github.com/Ensembl/ensembl-compara/pull/196)) and the release freezing for the time being ([link](https://docs.google.com/document/d/1OgPsEOE53htFMQX19XAcf3iA9eFVUsW7rcwqe1OrjbE/edit#heading=h.2qenazl2jq5f)), our dependency graph needed a big update. And here it is!

**Related JIRA tickets:**
- ENSCOMPARASW-3328

#### Change 1
- `*-trees WGA Orthology QC` and `*-trees High Confidence Orthologs` pipelines have been merged to their corresponding `*-trees`. The dependencies have been updated accordingly.

#### Change 2
- `Families`, `Mercator-Pecan`, `Ancestral Alleles` and `Patches against other species` pipelines won't be run during the release freezing.

#### Change 3
- From e102 we will have a single RelCo, so to ease their work, all tickets have been assigned to that person.

#### Change 4
- I have changed the direction of the graph from top-down to left-right because I find it more comfortable with our current wide screens.

## Testing
This is how the dependency graph would have looked for e101: https://www.ebi.ac.uk/~jalvarez/graphs/merged_release_graph.html?release=101

## Notes
_More changes were meant to be added (like the update of EPOs and Mercator-Pecan instead of rerunning them), but these tasks have been postponed to e103._

### DIFF
--- a/conf/pan/jira_recurrent_tickets.json
+++ b/conf/pan/jira_recurrent_tickets.json
@@ -43,17 +43,11 @@
       "component": "Relco tasks",
       "description": "https://www.ebi.ac.uk/seqdb/confluence/display/EnsCom/Intentions+for+release+<version>",
       "subtasks": [{
+            "assignee": "<RelCo>",
             "component": "Production tasks",
             "description": "[Bio::EnsEMBL::Compara::PipeConfig::Pan::ProteinTrees_conf|https://github.com/Ensembl/ensembl-compara/blob/release/<version>/modules/Bio/EnsEMBL/Compara/PipeConfig/Pan/ProteinTrees_conf.pm]",
             "summary": "Run the Protein-trees pipeline",
             "name_on_graph": "Protein-trees"
-         },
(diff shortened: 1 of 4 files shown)


JAlvarezJarreta:
The idea is that, at least for e102, we are moving to one RelCo (with @dthybert helping when he can), so I thought that it may be easier for the RelCo to handle all the tickets, and if he needs help, he can assign the ticket to someone else (or other people can reassign them to themselves). But if you prefer to leave it as it was, happy to revert that change.

CristiGuijarro:
All the production sub-tasks will be listed for the relco to see easily anyway under each main task, I don't think it is a necessary change for production sub-tasks, although I see the convenience for the main tickets.

PR #199, opened 2020-06-10
Update MLSS config XML file for vertebrates

JAlvarezJarreta:
## Description

Update MLSS config XML file for vertebrates to match the coming changes in our frozen production.

**Related JIRA tickets:**
- ENSCOMPARASW-3329
- ENSCOMPARASW-3071

#### Change 1
- Removed `Mercator-Pecan` and `Families` MLSSs.

#### Change 2
- Selected a single reference species per genome for LastZ (based on the best LastZ alignment chosen in each EPO Extended for e101).

#### Change 3
- Accept the species assembly too to define a specific genome version to be included in the EPOs and Mercator-Pecan (if updated, these genomes should be removed). Leave the references without assembly for the EPOs so they will be marked to be rerun automatically.

## Testing
`create_all_mlss.pl` script run successfully (in dry mode).

### DIFF
--- a/conf/vertebrates/mlss_conf.xml
+++ b/conf/vertebrates/mlss_conf.xml
@@ -77,7 +77,8 @@
     <!-- Sauropsids -->
     <one_vs_all method="LASTZ_NET" ref_genome="gallus_gallus">
       <species_set>
-        <taxonomic_group taxon_name="Sauropsida"/>
+        <taxonomic_group taxon_name="Sarcopterygii"/>
+        <taxonomic_group taxon_name="Amniota" exclude="1"/>
         <!-- Remove all the taxa that have a better reference than chicken -->
         <taxonomic_group taxon_name="Iguania" exclude="1"/>
       </species_set>
@@ -86,6 +87,15 @@
     <one_vs_all method="LASTZ_NET" ref_genome="anolis_carolinensis" against="Iguania"/>
 
     <!-- Fishes -->
+    <one_vs_all method="LASTZ_NET" ref_genome="danio_rerio">
+      <species_set>
+        <taxonomic_group taxon_name="Chordata"/>
+        <taxonomic_group taxon_name="Euteleostomi" exclude="1"/>
+        <taxonomic_group taxon_name="Cyclostoma" exclude="1"/>
+        <taxonomic_group taxon_name="Ciona" exclude="1"/>
+      </species_set>
+    </one_vs_all>
+
     <one_vs_all method="LASTZ_NET" ref_genome="oryzias_latipes">
       <species_set>
         <taxonomic_group taxon_name="Actinopterygii"/>
--- a/conf/vertebrates/mlss_conf.xml
+++ b/conf/vertebrates/mlss_conf.xml
@@ -78,7 +78,7 @@
     <one_vs_all method="LASTZ_NET" ref_genome="gallus_gallus">
       <species_set>
         <taxonomic_group taxon_name="Sarcopterygii"/>
-        <taxonomic_group taxon_name="Amniota" exclude="1"/>
+        <taxonomic_group taxon_name="Mammalia" exclude="1"/>
         <!-- Remove all the taxa that have a better reference than chicken -->
         <taxonomic_group taxon_name="Iguania" exclude="1"/>
       </species_set>
--- a/conf/vertebrates/mlss_conf.xml
+++ b/conf/vertebrates/mlss_conf.xml
@@ -4,11 +4,11 @@
   <!-- Collections are species-sets that are needed to define several mlsss -->
   <collections>
 
-    <!-- All chordates except the mouse strains, and the three outgroups -->
+    <!-- All chordates except the mouse strains, pig breeds and the three outgroups -->
     <!-- NB: we have several genomes of the same species, e.g. CHO, NMR -->
     <collection name="default">
       <taxonomic_group taxon_name="Chordata">
-        <!-- But exclude everything below mus_musculus -->
+        <!-- But exclude everything below mus_musculus and sus_scrofa -->
         <ref_for_taxon name="mus_musculus"/>
         <ref_for_taxon name="sus_scrofa"/>
       </taxonomic_group>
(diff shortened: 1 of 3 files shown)


conf/vertebrates/mlss_conf.xml
```diff
@@ -34,69 +34,69 @@
...
 
-    <one_vs_all method="LASTZ_NET" ref_genome="bos_taurus" ref_amongst="Eutheria">
+    <one_vs_all method="LASTZ_NET" ref_genome="bos_taurus">
       <species_set in_collection="default">
-        <taxonomic_group taxon_name="Cetartiodactyla"/>
+        <taxonomic_group taxon_name="Artiodactyla"/>
       </species_set>
     </one_vs_all>
 
-    <one_vs_all method="LASTZ_NET" ref_genome="ovis_aries" against="Caprinae" ref_amongst="Cetartiodactyla"/>
```
CristiGuijarro, line 76: I don't think agricultural biologists would be impressed with this removal

conf/vertebrates/mlss_conf.xml
```diff
@@ -34,69 +34,69 @@
...
-    <one_vs_all method="LASTZ_NET" ref_genome="ornithorhynchus_anatinus" against="Prototheria" ref_amongst="Euteleostomi"/>
-    <one_vs_all method="LASTZ_NET" ref_genome="latimeria_chalumnae" ref_amongst="Euteleostomi">
+    <one_vs_all method="LASTZ_NET" ref_genome="anolis_carolinensis" against="Iguania"/>
+
+    <!-- Fishes -->
+    <one_vs_all method="LASTZ_NET" ref_genome="oryzias_latipes">
       <species_set>
-        <!--Coelacanths and lungfishes -->
-        <taxonomic_group taxon_name="Sarcopterygii"/>
+        <!-- Include urochordates, cephalochordates, cyclostomes, cartilaginous fishes, ray-finned fishes, coelacanths and lungfishes -->
```
CristiGuijarro, line 86: This will also include amphibians, do we need the only amphibian we have to be aligned to both human and a fish?

conf/vertebrates/mlss_conf.xml
```diff
@@ -34,69 +34,69 @@
...
+        <taxonomic_group taxon_name="Chordata"/>
         <taxonomic_group taxon_name="Tetrapoda" exclude="1"/>
       </species_set>
     </one_vs_all>
+
+    <!-- Early vertebrates and other chordates -->
+    <one_vs_all method="LASTZ_NET" ref_genome="ciona_intestinalis" against="Ciona" ref_amongst="Cyclostomata"/>
+    <one_vs_all method="LASTZ_NET" ref_genome="petromyzon_marinus" against="Cyclostomata" ref_amongst="Ciona"/>
```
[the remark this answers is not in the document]
muffato: If we're meant to still do 1 LastZ for the genomes that are not in an EPO-Extended, then you need to add rules for the holes left in the taxonomy.

conf/vertebrates/mlss_conf.xml
CristiGuijarro: This is certainly the **most** questionable species set.

JAlvarezJarreta:
> The biggest problem is that this XMl file does not state that EPO and Pecan alignments are meant to shrink (whenever a non-reference species is changed / added) and recomputed (when a reference species is changed). At the moment, if the orangutan gets an assembly update, a new EPO mlss_id will be created _with it_.
> The problem is that we don't have the infrastructure (fields in this file and code in `create_all_mlss`) to track specific assemblies of a species, and [ENSCOMPARASW-3071](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-3071) was dismissed.
> 
> Unless you've got a workaround in the new pipeline that update EPOs and Pecan ?

So what I have done to detect this "update" in the MSAs is to check if the latest MLSS id has the `first_release` field equal to the current release. If so, it means something has been updated, so I get the MLSS id of the previous release, transfer the information from the previous to the current one, and do the update. The update is based on the difference on GenomeDB ids from the species sets of both MLSSs (so new species and updated assemblies are detected correctly).

conf/vertebrates/mlss_conf.xml
```diff
@@ -141,14 +141,6 @@
...
       </species_set>
     </multiple_alignment>
-    
-
-    <!-- Amniotes, excl. mouse strains -->
-    <multiple_alignment method="PECAN" gerp="1">
-      <species_set name="amniotes" display_name="amniota vertebrates" in_collection="default">
-        <taxonomic_group taxon_name="Amniota" only_good_for_alignment="1"/>
-      </species_set>
-    </multiple_alignment>
```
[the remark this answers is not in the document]
JAlvarezJarreta: I'm thinking now that my new code would work as expected even if new genomes are introduced (as it only takes into account the species removed from the previous release), but not the naming/species set creation. I'll add a new flag instead: `frozen`.

conf/vertebrates/mlss_conf.xml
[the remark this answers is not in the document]
JAlvarezJarreta: I have asked @dthybert about this. Once I have the answer, I'll update the XML file accordingly.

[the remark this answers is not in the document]
JAlvarezJarreta: We *need* to provide one LastZ for every genome.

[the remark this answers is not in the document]
muffato: OK, that's clear. It also explains why we need to run a LastZ for the new tasmanian devil (which was not in this XML file, BTW)

conf/vertebrates/mlss_conf.xml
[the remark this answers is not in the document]
muffato: EPO and Pecan are to be handled the same way, and should have similar definitions in the XML file. If you remove Pecan here you should remove EPO as well. If you need to keep EPO, you should keep Pecanas well ?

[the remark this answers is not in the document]
JAlvarezJarreta: I have addressed these issues in my new commits (hopefully).

conf/vertebrates/mlss_conf.xml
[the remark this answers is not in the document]
JAlvarezJarreta: Also addressed in one of my last commits (have tried to include coverage for all species).

conf/vertebrates/mlss_conf.xml
```diff
@@ -90,6 +94,7 @@
     </one_vs_all>
 
     <!-- Early vertebrates and other chordates -->
+    <one_vs_all method="LASTZ_NET" ref_genome="xenopus_tropicalis" against="Amphibia"/>
```
CristiGuijarro, line 97: :+1: :tada: (except we still only have one amphibian)...

conf/vertebrates/mlss_conf.xml
```diff
@@ -90,6 +94,7 @@
...
     <one_vs_all method="LASTZ_NET" ref_genome="ciona_intestinalis" against="Ciona" ref_amongst="Cyclostomata"/>
```
CristiGuijarro, line 98: Is this alignment even necessary? _Ciona_ is an invertebrate, so isn't used in any EPOs. This has probably always been here, because we use _Ciona_ as an outgroup to vertebrates. But what do we use the alignment for? This feels more like a Pan thing if we did more than just Pan homology. Just a comment.

conf/vertebrates/mlss_conf.xml
```diff
@@ -98,34 +103,206 @@
   <multiple_alignments>
 
     <!-- Primates -->
-    <multiple_alignment method="EPO+EPO_EXTENDED">
+    <multiple_alignment method="EPO">
+      <species_set name="primates">
+        <genome id="60"/>
```
CristiGuijarro, line 108: I cannot check all of these `genome_db_id` so just assuming you have them all correct. If a grant asks for another genome to be included, it just gets stuck in here/swapped out with another one?

conf/vertebrates/mlss_conf.xml
```diff
@@ -98,34 +103,206 @@
...
+        <genome id="209"/>
+        <genome id="210"/>
+        <genome id="221"/>
+        <genome id="222"/>
+        <genome id="225"/>
+        <genome id="317"/>
+        <genome id="361"/>
+      </species_set>
+    </multiple_alignment>
+    <multiple_alignment method="EPO_EXTENDED">
```
CristiGuijarro, line 123: We don't `gerp` here? (Just checking)

scripts/pipeline/create_all_mlss.pl
```diff
@@ -179,10 +179,11 @@ =head2 BEHAVIOUR CONFIGURATION
...
 
 sub find_genome_from_xml_node_attribute {
-    my ($xml_node, $attribute_name) = @_;
+    my ($xml_node, $attribute_name, $allow_old) = @_;
+    $allow_old //= 0;
     my $species_name = $xml_node->getAttribute($attribute_name);
-    my $gdb = $genome_dba->fetch_by_name_assembly($species_name) || throw("Cannot find $species_name in the available list of GenomeDBs");
-    die "Cannot find any current genomes with species_name '$species_name'. Please check that this name is still correct" unless $gdb->is_current;
+    my $gdb = $genome_dba->fetch_by_name_assembly($species_name) || $genome_dba->fetch_by_dbID($species_name) || throw("Cannot find $species_name in the available list of GenomeDBs");
+    die "Cannot find any current genomes matching '$species_name'. Please check that this name is still correct" unless ($allow_old || $gdb->is_current);
```
CristiGuijarro, line 186: I may be misreading this, but I am not sure this will behave as you would expect, it may never die because `$allow_old` has been initialised? Fine if it has been tested and proven otherwise, will take your word for it!

conf/vertebrates/mlss_conf.xml
JAlvarezJarreta: If we as with plants, it should be swapped. But we will see.

conf/vertebrates/mlss_conf.xml
JAlvarezJarreta: It wasn't in the original file, so I guess we use the GERP from mammals instead (or maybe it was a bug? @muffato @carlacummins).

scripts/pipeline/create_all_mlss.pl
JAlvarezJarreta: If `$allow_old` is 0 (default), then `$gdb->is_current` is what decides if it dies or not, and when `$allow_old` is 1, it will never die.

conf/vertebrates/mlss_conf.xml
muffato: The line between the www and metazoa sites is indeed blurry. Cionas were initially put on the www site, and I think EG didn't exist at the time. From there, all chordates are meant to be on the www site, even though in reality we have a much stronger focus on Vertebrates.
I seem to remember that amphioxus was somehow still to go to Metazoa, but I can't see it on any site, so maybe it has been delayed

conf/vertebrates/mlss_conf.xml
muffato: Correct, because (I was told) the genomes are too similar and GERP finds too many things. There is the mammals-wise GERP instead, indeed.

conf/vertebrates/mlss_conf.xml
```diff
@@ -34,98 +34,275 @@
...
   </pairwise_alignments>
 
   <multiple_alignments>
+
     <!-- Primates -->
-    <multiple_alignment method="EPO+EPO_EXTENDED">
+    <multiple_alignment method="EPO">
+      <species_set name="primates">
+        <genome id="60"/>
+        <genome id="150"/>
```
muffato, line 109: If you make this one `<genome name="homo_sapiens"/>` the script will be able to include it in the new mlss_id when it gets an assembly update (indeed we would like to recompute EPO). Same for all reference genomes (human, mouse, others ? I don't even know the correct list...) in other EPO alignments too (but not Pecan, right ?).

This way, we wouldn't need to manually update this file throughout the scale down

conf/vertebrates/mlss_conf.xml
CristiGuijarro: I think we should have all the chordates then! :sweat_smile:

CristiGuijarro:
> I would prefer using name+assembly rather than id, I think it would be easier to maintain, but I won't request this change now as time is running out. If I have some time later, I might make the change. Unless Cristi thinks it should be done.

I think it is okay for now with a change for later :+1:

conf/vertebrates/mlss_conf.xml
```diff
@@ -34,98 +34,274 @@
...
-    <!-- Mammals -->
-    <one_vs_all method="LASTZ_NET" ref_genome="monodelphis_domestica" against="Metatheria" ref_amongst="Amniota"/>
-    <one_vs_all method="LASTZ_NET" ref_genome="canis_lupus_familiaris" against="Carnivora" ref_amongst="Eutheria"/>
-    <one_vs_all method="LASTZ_NET" ref_genome="felis_catus" against="Feliformia" ref_amongst="Carnivora"/>
+    <one_vs_all method="LASTZ_NET" ref_genome="anolis_carolinensis" against="Iguania"/>
 
-    <one_vs_all method="LASTZ_NET" ref_genome="bos_taurus" ref_amongst="Eutheria">
-      <species_set in_collection="default">
-        <taxonomic_group taxon_name="Cetartiodactyla"/>
+    <!-- Fishes -->
```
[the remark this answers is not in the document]
JAlvarezJarreta: I checked this one and it seems that the more accurate scientific term to refer to fish when you have several species is `fishes` :sweat_smile:

conf/vertebrates/mlss_conf.xml
```diff
@@ -34,98 +34,274 @@
...
+    <one_vs_all method="LASTZ_NET" ref_genome="oryzias_latipes">
+      <species_set>
+        <taxonomic_group taxon_name="Actinopterygii"/>
       </species_set>
     </one_vs_all>
 
-    <one_vs_all method="LASTZ_NET" ref_genome="ovis_aries" against="Caprinae" ref_amongst="Cetartiodactyla"/>
-
     <!-- Early vertebrates and other chordates -->
+    <one_vs_all method="LASTZ_NET" ref_genome="xenopus_tropicalis" against="Amphibia"/>
```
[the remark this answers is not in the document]
JAlvarezJarreta: There has been an issue with my unison as it hasn't synced the file correctly, so this change hasn't been pushed.

conf/vertebrates/mlss_conf.xml
```diff
@@ -34,98 +34,274 @@
...
         <taxonomic_group taxon_name="Rodentia"/>
       </species_set>
     </one_vs_all>
-    <one_vs_all method="LASTZ_NET" ref_genome="gallus_gallus" against="Sauropsida" ref_amongst="Chordata"/>
-    <one_vs_all method="LASTZ_NET" ref_genome="danio_rerio" ref_amongst="Chordata">
-      <species_set>
-        <!-- Include urochordates, cephalochordates, cyclostomes, cartilaginous fishes, ray-finned fishes, coelacanths and lungfishes -->
-        <taxonomic_group taxon_name="Chordata"/>
-        <taxonomic_group taxon_name="Tetrapoda" exclude="1"/>
+
```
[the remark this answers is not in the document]
JAlvarezJarreta: Same as above.

conf/vertebrates/mlss_conf.xml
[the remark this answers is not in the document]
CristiGuijarro: Only in the context of a sentence when you are referring to multiple fish species. In this context, as a group of animals and a standalone word, I believe it should still be referred to as `fish` :slightly_smiling_face:

muffato:
@dthybert . I don't understand what you mean re. zebrafish and medaka: which one do you want to use in EPO-Extended ?

muffato:
Re. sauropsids: you say "as it is" but only mention chicken. "As it is" we have the lizard too

muffato:
We don't need to worry about opossum / platypus since' they're not in any EPO-Extended

JAlvarezJarreta:
> I'm afraid you need to revert some of the lastet changes. They can't yet be made because they would require running some LastZs in e102:
> 
> * for Perissodactyla: leave horse for now, and add a note somewhere to switch to cow when there is a horse update
> * same for Feliformia: keep cat for now, and we'll switch to dog later

Alternatively, I could start those LastZs now as the species involved won't be updated in e102. I have run `create_all_mlss.pl` and there's only 2 LastZs to be run with the current version of the MLSS XML config file in this PR:
```
_Bos taurus_ vs _Equus asinus asinus_ (on Btau)
_Gallus gallus_ vs _Leptobrachium leishanense_ (on Ggal)
```

muffato:
* Ah right, dog is already done against all Carnivora, so nothing to rerun there
* Is that not what we decided: chicken against all Sarcopterygii (except mammals) ?

JAlvarezJarreta:
I left `<one_vs_all method="LASTZ_NET" ref_genome="xenopus_tropicalis" against="Amphibia"/>` (wasn't sure what to do with it), so I just wanted to confirm which one of the two is correct, and I will remove the other.

CristiGuijarro:
> I left `<one_vs_all method="LASTZ_NET" ref_genome="xenopus_tropicalis" against="Amphibia"/>` (wasn't sure what to do with it), so I just wanted to confirm which one of the two is correct, and I will remove the other.

I don't think this will actually run anything unless we get another amphibian (it's very lonely), so it's probably pointless to keep unless someone comes along and says we need to include a new amphibian genome in compara :disappointed_relieved:

dthybert:
> Re. sauropsids: you say "as it is" but only mention chicken. "As it is" we have the lizard too

Sorry I also meant as it is with the lizard.

dthybert:
> d require running some

dthybert:
> I'm afraid you need to revert some of the lastet changes. They can't yet be made because they would require running some LastZs in e102:
> 
> * for Perissodactyla: leave horse for now, and add a note somewhere to switch to cow when there is a horse update
> * same for Feliformia: keep cat for now, and we'll switch to dog later

I  agree with this and in general we should minimize any lastz run. So if a new rule on a clade  involve running new lastz while there is no assembly update/new assembly affecting the clade, we should not implement the rule yet.

conf/vertebrates/mlss_conf.xml
```diff
@@ -34,98 +34,275 @@
...
+        <!-- Include urochordates, cephalochordates, cyclostomes, cartilaginous fishes, ray-finned fishes, coelacanths and lungfishes -->
+        <taxonomic_group taxon_name="Chordata"/>
         <taxonomic_group taxon_name="Tetrapoda" exclude="1"/>
       </species_set>
     </one_vs_all>
+
+    <!-- Early vertebrates and other chordates -->
+    <one_vs_all method="LASTZ_NET" ref_genome="xenopus_tropicalis" against="Amphibia"/>
+    <one_vs_all method="LASTZ_NET" ref_genome="ciona_intestinalis" against="Ciona" ref_amongst="Cyclostomata"/>
+    <one_vs_all method="LASTZ_NET" ref_genome="petromyzon_marinus" against="Cyclostomata" ref_amongst="Ciona"/>
```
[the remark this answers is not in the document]
dthybert: > The best lamprey alignment is against hagfish (then zebrafish) but the best hagfish alignment is against zebrafish. Elephant shark alignment has a slightly better alignment with human, but zebrafish is OK. Reedfish (which is an EPO) has a better alignment with zebrafish than human.
> The best coelacanth alignments are against chicken and platypus. All other alignments are equivalent. It's kinda the same for xenopus.
> 
> Overall, if we need to choose one, I would say:
> 
> * chicken against all Sarcopterygii except Amniota (Amniota = Sauropsida + Mammalia, which are both already addressed). That will capture coelacanth and xenopus
> * zebrafish against all Chordata, except Euteleostomi (Euteleostomi = Sarcopterygii (see the previous rule) + Actinopterygii (medaka)) and Cyclostoma and Ciona (they both have alignments against each other). This complex rule only matches the elephant shark at the moment, but it's there as a catch-all in case we get other basal vertebrates / chordates, given the fact that zebrafish has OK alignments with the species we have


> @dthybert . I don't understand what you mean re. zebrafish and medaka: which one do you want to use in EPO-Extended ?

For EPO extended in fish the rule you wrote above is fine (I missed it yesterday when I was reviewing). Zebra fish is such a big reference for the fish community that anyway if its genome is updated it will have to be treated like the mouse and the human.

JAlvarezJarreta:
I have excluded Amphibia for chicken and kept horse as reference for Perissodactyla, so no LastZs are required to be run for e102 (besides the species that will be updated).

conf/vertebrates/mlss_conf.xml
```diff
@@ -34,98 +34,285 @@
...
       </species_set>
     </one_vs_all>
-    <one_vs_all method="LASTZ_NET" ref_genome="gallus_gallus" against="Sauropsida" ref_amongst="Chordata"/>
-    <one_vs_all method="LASTZ_NET" ref_genome="danio_rerio" ref_amongst="Chordata">
-      <species_set>
-        <!-- Include urochordates, cephalochordates, cyclostomes, cartilaginous fishes, ray-finned fishes, coelacanths and lungfishes -->
-        <taxonomic_group taxon_name="Chordata"/>
-        <taxonomic_group taxon_name="Tetrapoda" exclude="1"/>
+
+    <one_vs_all method="LASTZ_NET" ref_genome="canis_lupus_familiaris" against="Carnivora"/>
```
muffato, line 59: Following https://github.com/Ensembl/ensembl-compara/pull/199#issuecomment-644611467 let's keep the e101 rules and edit them if cat / dog changes
```suggestion
    <one_vs_all method="LASTZ_NET" ref_genome="canis_lupus_familiaris" against="Caniformia"/>
    <one_vs_all method="LASTZ_NET" ref_genome="felis_catus" against="Feliformia"/>
```

conf/vertebrates/mlss_conf.xml
```diff
@@ -34,98 +34,285 @@
...
 
-    <!-- First, our top four species, which are references for all chordates -->
+    <!-- Mammals -->
+    <!-- Use of "default" collection excludes the mouse strains and the pig breeds -->
+    <!-- NB: we have a CACTUS alignment for mouse strains and an EPO Extended MSA for pig breeds -->
     <one_vs_all method="LASTZ_NET" ref_genome="homo_sapiens">
-      <!-- "default" collection means that the mouse strains are excluded (because only the reference mouse deserves to be aligned to human) -->
       <species_set in_collection="default">
-        <taxonomic_group taxon_name="Chordata"/>
+        <taxonomic_group taxon_name="Mammalia"/>
```
muffato, line 42: To match the definition of the EPOs
```suggestion
        <taxonomic_group taxon_name="Eutheria"/>
```

conf/vertebrates/mlss_conf.xml
```diff
@@ -34,98 +34,285 @@
...
+        <!-- Remove all the taxa that have a better reference than human -->
+        <taxonomic_group taxon_name="Rodentia" exclude="1"/>
+        <taxonomic_group taxon_name="Carnivora" exclude="1"/>
+        <taxonomic_group taxon_name="Perissodactyla" exclude="1"/>
+        <taxonomic_group taxon_name="Artiodactyla" exclude="1"/>
+        <taxonomic_group taxon_name="Metatheria" exclude="1"/>
+        <taxonomic_group taxon_name="Prototheria" exclude="1"/>
```
muffato, line 49: Not needed if we use Eutheria above
```suggestion
```

conf/vertebrates/mlss_conf.xml
muffato: I had a call with @JAlvarezJarreta to talk about this one. The principle is to move to the new rules that don't necessitate any new LastZ, and keep the other ones for now. We already have have dog vs all carnivores, so we can remove cat vs all feliformes

conf/vertebrates/mlss_conf.xml
```diff
@@ -34,98 +34,283 @@
...
-    <one_vs_all method="LASTZ_NET" ref_genome="monodelphis_domestica" against="Metatheria" ref_amongst="Amniota"/>
-    <one_vs_all method="LASTZ_NET" ref_genome="canis_lupus_familiaris" against="Carnivora" ref_amongst="Eutheria"/>
-    <one_vs_all method="LASTZ_NET" ref_genome="felis_catus" against="Feliformia" ref_amongst="Carnivora"/>
+    <!-- Sauropsids -->
+    <one_vs_all method="LASTZ_NET" ref_genome="gallus_gallus">
+      <species_set>
+        <taxonomic_group taxon_name="Sarcopterygii"/>
+        <taxonomic_group taxon_name="Mammalia" exclude="1"/>
+        <!-- Remove all the taxa that have a better reference than chicken -->
+        <taxonomic_group taxon_name="Iguania" exclude="1"/>
```
muffato, line 81: For that reason, we can remove the lizard vs Iguania rule
```suggestion
```

conf/vertebrates/mlss_conf.xml
```diff
@@ -34,98 +34,283 @@
...
+        <taxonomic_group taxon_name="Amphibia" exclude="1"/>
+      </species_set>
+    </one_vs_all>
 
-    <one_vs_all method="LASTZ_NET" ref_genome="bos_taurus" ref_amongst="Eutheria">
-      <species_set in_collection="default">
-        <taxonomic_group taxon_name="Cetartiodactyla"/>
+    <one_vs_all method="LASTZ_NET" ref_genome="anolis_carolinensis" against="Iguania"/>
```
muffato, line 86:
```suggestion
```

conf/vertebrates/mlss_conf.xml
JAlvarezJarreta: My understanding is that if the change of the rules (i.e. reduce the number of references) doesn't generate new LastZs, then they should be applied now.

PR #205, opened 2020-06-17
Removed 'become' commands from our pipelines

JAlvarezJarreta:
## Description

Since we have two virtual users now, and one virtual user cannot become the other, but we would like to start running pipelines with the virtual users, the `become` instructions had to be removed.

**Related JIRA tickets:**
- ENSCOMPARASW-3067

#### Change 1
- Removed all options to become the shared user as with the ACL rules they are no longer required.

#### Change 2
- Removed runnable `CopyFilesAsUser.pm` as it is no longer used or required.

## Testing
Not tested. Will do during production.

## Notes
`PrepareMasterDatabaseForRelease` pipeline is the only one where the `become` cannot be removed as it access the warehouse and there are no ACL rules there. I will update the instructions in confluence to warn about this restriction once the PR is approved.


JAlvarezJarreta:
I have removed the option because as soon as you use it in one pipeline, it cannot be created with a shared user (which is part of what we aim to achieve from this release on).

carlacummins:
I'm with @CristiGuijarro on this - I don't think we should completely remove support for `become` from the runnables. The pipelines shouldn't use the option any more, but I don't see the benefit in removing the feature entirely.

modules/Bio/EnsEMBL/Compara/PipeConfig/CreateSpeciesTree_conf.pm
```diff
@@ -280,17 +279,13 @@ sub pipeline_analyses {
               1 => ['copy_files_to_sketch_dir'],
           },
         },
-
+
```
carlacummins, line 282:
```suggestion

```

modules/Bio/EnsEMBL/Compara/PipeConfig/CreateSpeciesTree_conf.pm
```diff
@@ -280,17 +279,13 @@ sub pipeline_analyses {
...
+          -module     => 'Bio::EnsEMBL::Hive::RunnableDB::SystemCmd',
           -parameters => {
               'destination_dir' => $self->o('sketch_dir'),
-              'become_user'     => $self->o('write_access_user'),
               'source_dir'      => $self->o('output_dir'),
-              'file_list' => [
-                '#source_dir#/*.msh',
-                '#source_dir#/#collection#',
-              ],
+              'cmd'             => 'cp #source_dir#/*.msh #destination_dir# && cp #source_dir#/#collection# #destination_dir#',
```
carlacummins, line 288:
```suggestion
              'cmd' => 'cp #source_dir#/*.msh #destination_dir# && cp #source_dir#/#collection# #destination_dir#',
```

JAlvarezJarreta: This is to align with the previous two parameters, so I'm going to leave it if you don't mind :)

PR #206, opened 2020-06-18
Datacheck analysis added to LoadMembers

CristiGuijarro:
## Description

_Sometimes the cores are handed over without adequate checks, so when loaded we can end up with empty enum fields when we should not._

**Related JIRA tickets:**
- ENSCOMPARASW-3543

## Overview of changes
_A new analysis was included to run at the end of the pipeline and check for empty fields using the BlankEnums datacheck._

#### Change #1
- Added semaphore in pipeline to flow to new `datachecks` from `hc_members_globally`.

#### Change #2
- Plugged in compara `RunDataChecks.pm` runnable with necessary parameters.

## Testing
CITest pipeline [here](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-7&port=4617&dbname=cristig_citest_load_members_101) (test pass, then manual intervention in db to cause failure and correct failure reported)

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/LoadMembers_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/LoadMembers_conf.pm
@@ -514,9 +514,24 @@ sub core_pipeline_analyses {
                 mode            => 'members_globally',
             },
             %hc_analysis_params,
-            -flow_into          => WHEN(
-                            '#load_uniprot_members#' => 'save_uniprot_release_date',
-                            ),
+            -flow_into          => {
+                '2->A' => WHEN( '#load_uniprot_members#' => 'save_uniprot_release_date' ),
+                'A->1' => [ 'dc_members_enums' ],
+            },
+        },
+
+        {   -logic_name      => 'dc_members_enums',
+            -module          => 'Bio::EnsEMBL::Compara::RunnableDB::RunDataChecks',
+            -parameters      => {
+                'datacheck_names'  => ['BlankEnums'],
+                'work_dir'         => $self->o('work_dir'),
+                'history_file'     => '#work_dir#/datacheck.compara_load_members.history.json',
+                'output_file'      => '#work_dir#/datacheck.compara_load_members.tap.txt',
+                'failures_fatal'   => 1,
+                'pdbname'          => $self->o('pipeline_name'),
+                'dbtype'           => 'compara',
+            },
+            -max_retry_count => 0,
         },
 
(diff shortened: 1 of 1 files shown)


muffato:
I know the Jira ticket is only about this particular DC, but can you check if the entire set of Compara critical DCs could be used ? Otherwise, are there any other DCs that could be added here ?

CristiGuijarro:
@muffato I have checked for any other DCs that would be handy in this pipeline and I could find only 2 others possibilities: `ForeignKeysCompara` and `CheckSequenceTable`. I can see what happens when I plug those in the test pipeline and we can decide whether it is worth including them?

modules/Bio/EnsEMBL/Compara/PipeConfig/LoadMembers_conf.pm
```diff
@@ -516,14 +516,14 @@ sub core_pipeline_analyses {
             %hc_analysis_params,
             -flow_into          => {
                 '2->A' => WHEN( '#load_uniprot_members#' => 'save_uniprot_release_date' ),
-                'A->1' => [ 'dc_members_enums' ],
+                'A->1' => [ 'dc_sequences_and_enums' ],
```
JAlvarezJarreta, line 519: Since we have the list of DCs in the analysis, what do you think about calling the analysis in a more generic way?
```suggestion
                'A->1' => [ 'datachecks' ],
```
I think adding what was being checked in the analysis name was done because the HCs didn't always have a comprehensive name/task.

modules/Bio/EnsEMBL/Compara/PipeConfig/LoadMembers_conf.pm
```diff
@@ -516,14 +516,14 @@ sub core_pipeline_analyses {
...
             },
         },
 
-        {   -logic_name      => 'dc_members_enums',
+        {   -logic_name      => 'dc_sequences_and_enums',
```
JAlvarezJarreta, line 523: Same here.

muffato:
```suggestion
        {   -logic_name      => 'datachecks',
```

muffato:
It is quite scary that the ForeignKeys DC doesn't apply to this database ! But I've checked it and it seems there is only one failure I think the DC itself should be fixed - will move this discussion to Slack.

I agree with @JAlvarezJarreta : `datachecks` is more generic and future-proof.

PR #207, opened 2020-06-19
Use flatfiles for homology inference and loading

carlacummins:
## Description

Homologies are inferred from gene trees using the OrthoTree runnable. Traditionally, homologies were handled as follows:
- `ortho_tree` analysis writes homologies directly to the `homology`/`homology_member` tables
- these tables were read and data dumped to flatfile for the post-tree stages of the pipeline (one file per homology mlss_id)
- GOC, WGA and HighConf data written to flatfiles
- final step of the HighConfidenceOrthologs pipeline uses `UPDATE` statements to write `goc_score`, `wga_coverage` and `is_high_confidence` attributes

This PR seeks to drastically change how these steps are performed, completely avoiding the database tables until the end of the process. The new process is:
- `ortho_tree` analysis writes homologies to flatfiles (one file per gene tree)
- orthotree files are converted to per-mlss files in preparation for post-tree stage of pipeline (this can be performed with much greater concurrency)
- GOC, WGA and HighConf data written to flatfiles
- final step of the HighConfidenceOrthologs pipeline combines per-mlss homology dump file with attribute files (`goc_score`, `wga_coverage`, `is_high_confidence`) and uses `mysqlimport` to import the data in full to the appropriate database tables

**Related JIRA tickets:**
- ENSCOMPARASW-3023 (inference to flatfiles)
- ENSCOMPARASW-3484 (load table from flatfiles)

#### OrthoTree writes to flatfiles
- allow runnable to create Homology objects as usual, but instead of writing to db, write to `$self->param('output_file')`
- `full_string` method added to Homology object - this is what will be written to file
- healthchecking of the output file embedded directly into this runnable
- since many downstream analyses rely heavily on `homology_id`, these will now be generated by concatenating sorted seq_member_ids instead of relying on MySQL's `AUTO INCREMENT`. this still ensures uniqueness as there can only be 1 homology between any given pair of `seq_members`

#### Split OrthoTree output for post-tree
- new runnable `Bio::EnsEMBL::Compara::RunnableDB::GeneTrees::SplitOrthoTreeOutput` takes a list of mlss_ids and a directory. it traverses the directory and reads each file it finds, checking each line for an mlss_id matching the inputs. 1 file created per-mlss, mirroring the directory structure we currently use (reverse hashed mlss_id).
- this step worked best when mlss_ids are batched in ~100. added batching option to `MLSSIDFactory`
- unit test and test data (`modules/t/homology_flatfiles/mysqlimport_test/`) included

#### Inclusion of new process and rewiring of pipelines
- a new group of analyses have been added to `Bio::EnsEMBL::Compara::PipeConfig::Parts::DumpHomologiesForPosttree` :  `pipeline_analyses_split_homologies_posttree`
- include these analyses in ProteinTrees and ncRNATrees
- include `orthotree_dir` param in both pipelines
- remove `hc_tree_homologies` as hc is now embedded directly in orthotree runnable - replace with Dummy analysis (`final_tree_steps`) to keep the structure the same

#### Updated other runnables and unit tests with new field names
- the file headers / field names are slightly different now than they were before, so several runnables that use the homology dumps needed to be updated
- mostly the changes include: 
    - `hom_*` -> `homology_*` (e.g. `hom_gene_member_id` -> `homology_gene_member_id`)
    - `identity`/`hom_identity` -> `perc_id`/`homology_perc_id`
    - `seq_member_stable_id` -> `stable_id`
- unit tests and input files were also updated with these changes

#### ComputeStatistics moved in ProteinTrees
- the ComputeStatistics runnable extensively uses MySQL queries to fetch stats
- I attempted to make it use the flatfiles, but it was both slower and producing incorrect values, so decided to move it to the end of the pipeline (after homology data was loaded back into the db)
- also needed to remove the homology_id threshold as homology_ids will no longer fall into distinct ranges (see note above on `homology_ids`)

#### Move UpdateTableFromFile runnable
- move runnable to new `Bio/EnsEMBL/Compara/RunnableDB/Flatfiles/` directory
- update unit test

#### FindHomologyIDRange runnable
- in order to minimise disruption, the homology_ids should be stored into the ranges that have been defined for many years. to achieve that without impacting concurrency too much, I've created a runnable to act as a rudimentary AUTO_INCREMENT
- using a `pipeline_wide_parameter`, it simply assigns each mlss_id (jobs are split on mlss_id) a range before handing off to `mysqlimport`
- `MySQLImportHomologies.pm` now discards the homology_id from the file and uses this range + a counter to generate new ones

## Testing
- Unit tests have been included for new runnables
- Testing happened within a (now mangled) database
- Testing of final db structures can be found here:
    - ptrees: [mysql://ensro@mysql-ens-compara-prod-2:4522/carlac_ptrees_orthotree_test](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-2&port=4522&dbname=carlac_ptrees_orthotree_test)
    - nctrees: [mysql://ensro@mysql-ens-compara-prod-3:4523/carlac_nctrees_orthotree_test](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-3&port=4523&dbname=carlac_nctrees_orthotree_test)

## Notes
I am sorry - this got out of control 😅


JAlvarezJarreta:
I can see some commits that I feel they shouldn't be there.

carlacummins:
> I can see some commits that I feel they shouldn't be there (like `Merge branch 'release/102'`).

Why shouldn't they be there? I had started this work a long time ago and needed to merge all `release/102` changes into my branch before creating the PR so that it has no merging conflicts

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/SplitOrthoTreeOutput.pm
```diff
@@ -0,0 +1,159 @@
...
+
+sub orthotree_files {
+    my $self = shift;
+
+    return $self->param('orthotree_files') if $self->param('orthotree_files');
+
+    my @files;
+    my $orthotree_dir = $self->param('orthotree_dir');
+    find(sub {
+        push @files,$File::Find::name if /\.orthotree.tsv$/;
```
JAlvarezJarreta, line 152:
```suggestion
        push @files, $File::Find::name if /\.orthotree.tsv$/;
```

modules/t/MySQLImportHomologies.t
```diff
@@ -0,0 +1,110 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+use strict;
+use warnings;
+
+use Data::Dumper;
```
JAlvarezJarreta, line 20: Unused:
```suggestion
```

modules/Bio/EnsEMBL/Compara/PipeConfig/ncRNAtrees_conf.pm
```diff
@@ -177,7 +178,7 @@ sub default_options {
             #Parameters for HighConfidenceOrthologs
             'threshold_levels'            => [ ],          # division specific
             'high_confidence_capacity'    => 500,          # how many mlss_ids can be processed in parallel
-            'update_homologies_capacity'  => 30,           # how many homology mlss_ids can be updated in parallel
+            'import_homologies_capacity'  => 50,           # how many homology mlss_ids can be mysqlimported in parallel
```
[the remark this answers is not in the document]
carlacummins:
```suggestion
            'import_homologies_capacity'  => 50,           # how many homology mlss_ids can be imported in parallel (via mysqlimport)
```

modules/Bio/EnsEMBL/Compara/Homology.pm
```diff
@@ -403,6 +405,85 @@ sub toString {
...
+                a minimised amount of data on the given homology.
+                USE ONLY FOR DEBUGGING not for data output since the
+                format of this output may change as need dictates.
+  Returntype  : String
+  Exceptions  : none
+  Caller      : general
+
+=cut
+
+sub toJSON {
```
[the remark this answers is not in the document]
carlacummins: I just copied the Description from the `toString` method, which has the same warning. I see it more for external users. I think I will leave it as it is.

modules/Bio/EnsEMBL/Compara/RunnableDB/Flatfiles/MySQLImportHomologies.pm
```diff
@@ -0,0 +1,163 @@
...
+
+use base ('Bio::EnsEMBL::Compara::RunnableDB::BaseRunnable');
+
+sub fetch_input {
+    my $self = shift;
+
+    # handle any attrib files that may have been passed
+    $self->param('attribs', {});
+    my $attrib_files = $self->param('attrib_files');
+    return unless ( $attrib_files );
```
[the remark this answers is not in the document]
carlacummins: I don't think your line is achieving the same things as the lines it's replacing though.. 
I need `$self->param('attribs')` to be set before returning for later use. And I want to `return` (i.e. exit `fetch_input`) when there are no attribute files given. I could def write it in many different ways, but I'm not seeing any real benefit. Leaving it as is.

modules/Bio/EnsEMBL/Compara/PipeConfig/ncRNAtrees_conf.pm
```diff
@@ -1166,12 +1169,19 @@ sub core_pipeline_analyses {
...
+                'output_flatfile'     => '#orthotree_dir#/#hashed_gene_tree_id#/#gene_tree_id#.orthotree.tsv',
             },
-            -flow_into  => [ 'hc_tree_homologies' ],
+            -flow_into  => [ 'final_tree_steps' ],
             -rc_name    => '4Gb_job',
         },
 
+        {   -logic_name => 'final_tree_steps',
+            -module     => 'Bio::EnsEMBL::Hive::RunnableDB::Dummy',
+            -flow_into  => [ 'ktreedist', 'consensus_cigar_line_prep' ],
```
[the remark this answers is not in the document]
carlacummins: Yes, well spotted. This part of the pipeline is wired differently between ptrees and nctrees and I didn't notice.

modules/t/homology_flatfiles/ortho.test.tsv
```diff
@@ -1,51 +1,51 @@
...
+211339	ortholog_one2many	4084424	4084424	40133000	1331260	1789752	ENSMEUP00000013887	111	anolis_carolinensis	23.0425	902946280	904801451	C25A1.4.2	142	gallus_gallus	22.6872
+232255	ortholog_one2many	4084424	4084424	40133000	1314646	1773079	ENSMEUP00000004376	111	anolis_carolinensis	18.4573	902946280	904801451	C25A1.4.2	142	gallus_gallus	29.5154
+423301	ortholog_one2many	4084188	4084188	40133000	1315852	1774288	ENSMEUP00000010046	111	anolis_carolinensis	32.6754	902962376	904835472	Y18D10A.17.1	142	gallus_gallus	43.8235
+442981	ortholog_one2many	4084188	4084188	40133000	1313861	1772294	ENSMEUP00000005751	111	anolis_carolinensis	32.3529	902962376	904835472	Y18D10A.17.1	142	gallus_gallus	19.4118
+442999	ortholog_one2many	4084188	4084188	40133000	1315809	1774244	ENSMEUP00000007547	111	anolis_carolinensis	34.8837	902962376	904835472	Y18D10A.17.1	142	gallus_gallus	39.7059
+963386	ortholog_one2many	4082536	4082536	40133001	1298060	1735142	gallus_gallus	ENSMEUP00000015111	anolis_carolinensis	111	anolis_carolinensis	11.236	903146987	905178502	Y67A10A.7.1	142	gallus_gallus	11.5942
```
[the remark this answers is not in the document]
carlacummins: Yep - dunno what happened, but definitely a mangled line. Have fixed with 92ac8ab

modules/Bio/EnsEMBL/Compara/RunnableDB/Flatfiles/MySQLImportHomologies.pm
```diff
@@ -0,0 +1,163 @@
...
+
+        # add the attribs
+        foreach my $attrib_name ( 'goc_score', 'wga_coverage', 'is_high_confidence' ) {
+            my $this_attrib = $attribs->{$row->{homology_id}}->{$attrib_name};
+            $row->{$attrib_name} = defined $this_attrib ? $this_attrib : '\N';
+        }
+
+        my ( $homology_row, $homology_member_rows ) = $self->split_row_for_homology_tables($row);
+        print $h_csv $homology_row;
+        print $hm_csv $homology_member_rows;
```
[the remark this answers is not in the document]
carlacummins: Should have looked more closely at this suggestion before incorporating it! 😅 This results in empty files being passed to `mysqlimport`, unless you run in debug mode! Need to revert

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthoTree.pm
```diff
@@ -566,5 +602,115 @@ sub check_homology_consistency {
...
 }
 
+sub _create_flatfile {
+    my $self = shift;
+
+    my $outfile = $self->param('output_flatfile');
+
+    # create directory
+    my $outdir = dirname($outfile);
+    $self->run_command("mkdir -p $outdir");
```
muffato, line 612: I know it is done this way in several other runnables, but `run_command` will disconnect from the database. Assuming the rest of `run_analysis` needs to fetch some data, that's going to be a very short disconnection cycle. I would use
```perl
use File::Path qw(make_path)

make_path($outdir);
```

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthoTree.pm
```diff
@@ -566,5 +602,115 @@ sub check_homology_consistency {
...
+
+sub _check_file_integrity {
+    my ( $self, $file ) = @_;
+
+    # check line counts
+    my $leaf_count = scalar @{ $self->param('gene_tree')->root->get_all_leaves };
+    my $exp_lines = (($leaf_count * ($leaf_count - 1)) / 2) + 1; # add 1 for header line
+    my @wc_output = split(/\s+/, $self->get_command_output("wc -l $file"));
+    my $got_line_count = $wc_output[0];
+    die "Expected $exp_lines lines, but got $got_line_count: $file" if $exp_lines != $got_line_count;
```
muffato, line 708: I'm surprised by this. We don't store between-species paralogues, so this `$exp_lines` formula is an upper bound, but the count itself. The Runnable keeps the count in `$self->param('n_stored_homologies')`

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthoTree.pm
```diff
@@ -566,5 +602,115 @@ sub check_homology_consistency {
...
+
+    # check column integrity
+    my $awk_output = $self->get_command_output("awk '{print NF}' $file | sort | uniq -c");
+    my @col_counts = split("\n", $awk_output);
+    die "Expected equal number of columns throughout the file. Got:\n$awk_output" if scalar @col_counts > 1;
```
muffato, line 713: That'd be useful in `Utils::FlatFile`

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthoTree.pm
```diff
@@ -566,5 +602,115 @@ sub check_homology_consistency {
...
+            );
+        }
+
+        # Checks that all the relevant fields are non-NULL or non-zero
+        unless (
+            defined $row->{homology_type} &&
+            defined $row->{seq_member_id} &&
+            defined $row->{homology_seq_member_id} &&
+            defined $row->{cigar_line} &&
+            length($row->{cigar_line}) > 0 &&
```
muffato, line 667: Since `map_row_to_header` splits on `\s+`, the lengths will always be `> 0`

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthoTree.pm
```diff
@@ -566,5 +602,115 @@ sub check_homology_consistency {
...
+    my @header_cols = split(/\s+/, $header_line);
+
+    my $c = 1;
+    my %seen_homologies;
+    while ( my $line = <$fh> ) {
+        my $row = map_row_to_header($line, \@header_cols);
+        my ( $st_id, $hom_st_id ) = ($row->{stable_id}, $row->{homology_stable_id});
+
+        # Each homology must be linked to exactly 2 members
+        unless (defined $st_id && defined $hom_st_id) {
```
muffato, line 643: At this stage, we have checked that all the lines in the file have the same number of columns (which  `map_row_to_header` does too). So by construction all the hash values of `$row` will be defined. The only way of getting an undefined value in `$st_id` is by `stable_id` not being a valid hash key / column header (same for `homology_stable_id` and all the `defined` below from line 663).
It seems convoluted. You can check that directly in the header line, but in fact you can check that in `fetch_input`

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthoTree.pm
```diff
@@ -566,5 +602,115 @@ sub check_homology_consistency {
...
+        # A pair of gene can only appear in 1 homology at most
+        $seen_homologies{"$st_id - $hom_st_id"}++;
+        $seen_homologies{"$hom_st_id - $st_id"}++;
+        if ($seen_homologies{"$st_id - $hom_st_id"} > 1 ||
+            $seen_homologies{"$hom_st_id - $st_id"} > 1 ) {
+            die sprintf(
+                "Pair (%s - %s) seen more than once (line %d)",
+                ( $st_id, $hom_st_id, $c )
+            );
+        }
```
muffato, line 659:
```suggestion
        if (my $prev_line = ($seen_homologies{"$st_id - $hom_st_id"} || $seen_homologies{"$hom_st_id - $st_id"}) ) {
            die sprintf(
                "Pair (%s - %s) seen more than once (lines %d and %d)",
                ( $st_id, $hom_st_id, $prev_line, $c )
            );
        }
        $seen_homologies{"$st_id - $hom_st_id"} = $c;
        $seen_homologies{"$hom_st_id - $st_id"} = $c;
```

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthoTree.pm
muffato, line 708: That'd be useful in `Utils::FlatFile` and could be used in RunnableDB/GeneTrees/DumpHomologiesTSV.pm

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthoTree.pm
```diff
@@ -566,5 +602,115 @@ sub check_homology_consistency {
...
+    my $self = shift;
+
+    my $outfile = $self->param('output_flatfile');
+
+    # create directory
+    my $outdir = dirname($outfile);
+    $self->run_command("mkdir -p $outdir");
+
+    # open file handle and print header
+    open( my $outfh, '>', $outfile ) or die "Cannot open $outfile for writing";
```
muffato, line 615: I can't find the matching `close`

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthoTree.pm
```diff
@@ -538,8 +567,15 @@ sub store_gene_link_as_homology {
         $self->param('orthotree_homology_counts')->{'gene_split'}++;
     }
   }
-  
-  $self->param('homologyDBA')->store($homology) unless $self->param('_readonly');
+
+    if ( $self->param('output_flatfile') ) {
```
muffato, line 571: It's strange to add `$output_fh` to so many function calls and in the end use `$self->param('output_flatfile')` to check whether to use it or not.
Either use `$output_fh` in the `if` here, or put `$output_fh` in `$self->param` when the file is opened and you don't need to modify the signatures of `tag_genepairlink` and `store_gene_link_as_homology`

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthoTree.pm
carlacummins: Hum... almost 50k jobs and this check passed on all of them.. That's a little concerning

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthoTree.pm
carlacummins: Yes, this is true. TBH, I just ported all the checks from the original SQL healthchecks without fully considering whether the check really needed to be performed. I will remove this one. You are right - it would be picked up in one of the other checks.

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthoTree.pm
```diff
@@ -538,8 +567,15 @@ sub store_gene_link_as_homology {
...
+        $homology->{_dba} = $self->compara_dba;
+        # create homology_id from seq_member_ids as this will be unique
+        $homology->dbID( join('', sort map {$_->seq_member_id} @{ $homology->get_all_Members }) );
```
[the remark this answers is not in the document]
carlacummins: Yes, the original plan was for this id to be the one that is written to the database. This hinged on the idea that the ids I generated here would be unique, which apparently we can't rely on (currently running a test on how often duplicate ids occurred in my tests). I would indeed need to remove all references to the id ranges (quite a bit of work).

I think I have an alternative solution where we retain the 'traditional' ids in the defined ranges. `MySQLImportHomologies.pm` can discard the homology_id from the file and generate one with a simple counter. I will create a runnable to run before `MySQLImportHomologies.pm` which will identify the homology_id range each homology mlss should occupy based on the number of lines in files preceding it. Using an analysis_capacity of 1 will prevent clashes. In this model, I can use an underscore as you've suggested to make the internal file-based ids. 

I can't think of any downsides to having the file-based and db-based ids mismatching ATM 🤔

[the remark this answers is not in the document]
carlacummins: I think commit c1cb4a6 addresses and fixes this issue. I am aware that the standalone version of `HighConfidenceOrthologs_conf` will not work as it stands. I am still trying to figure out the cleanest way to define the `dbID_range_index` param there. Any ideas?

modules/Bio/EnsEMBL/Compara/RunnableDB/Flatfiles/MySQLImportHomologies.pm
```diff
@@ -0,0 +1,164 @@
...
+    # open homology file
+    open(my $hom_fh, '<', $filename) or die "Cannot open $filename for reading";
+    my $header_line = <$hom_fh>;
+    my @header_cols = split(/\s+/, $header_line);
+
+    # open output files - one per table in tmp space
+    my $homology_csv = $self->worker_temp_directory . '/homology.txt';
+    open(my $h_csv, '>', $homology_csv) or die "Cannot open $homology_csv for writing";
+    my $homology_member_csv = $self->worker_temp_directory . '/homology_member.txt';
+    open(my $hm_csv, '>', $homology_member_csv) or die "Cannot open $homology_member_csv for writing";
```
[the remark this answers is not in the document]
carlacummins: Yes, me too. We may end up being plagued by `/scratch` space not being big enough, etc. 

My hope is that it is a problem in a small proportion of files and those that are repeat offenders would have to be run with the `--worker_base_temp_dir` pointing to a `/hps` space instead. In the future, we could modify the pipeline to allow a certain number of tries with `/scratch` and then flow to another instance of this analysis with a `/hps` tmp space instead. It would take a little noodling around with hive. I'm not sure if the `0` branch would be able to do this?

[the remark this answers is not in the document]
carlacummins: The largest file generated in e101 is 592M (`/gpfs/nobackup/ensembl/carlac/default_vertebrates_protein_trees_101/dumps/homology_dumps/9/8/2/4/4/144289.protein.homologies.tsv`). This file is with the old schema, so is missing multiple fields. I still would not expect it to exceed ~1gb. Since this runnable is just splitting this data into 2 different files, I wouldn't expect the output to `/scratch` to be much bigger.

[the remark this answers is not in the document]
muffato: FYI the branch `0` wouldn't have helped. It only captures LSF kills, not errors thrown by Perl itself

modules/Bio/EnsEMBL/Compara/RunnableDB/Flatfiles/UpdateTableFromFile.pm
```diff
@@ -27,7 +27,7 @@ column names (`primary_key` is required and should be present in all files).
 
 =cut
 
-package Bio::EnsEMBL::Compara::RunnableDB::UpdateTableFromFile;
+package Bio::EnsEMBL::Compara::RunnableDB::Flatfiles::UpdateTableFromFile;
```
muffato, line 30: This module is not used any more. I imagine the changes you've made in it were for a previous iteration of your work, but without any context, I can't assess why these changes are required. I'll just approve it as it is

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthologyStats.pm
```diff
@@ -68,19 +68,9 @@ sub fetch_input {
...
-            foreach my $comp_gdb ( @$comp_gdbs ) {
-              $gdb_map->{$comp_gdb->dbID} = $gdb->dbID;
-            }
-        } else {
-            $gdb_map->{$gdb->dbID} = $gdb->dbID;
-        }
-    }
+
+    # create map between genome names and ids - excluding ids for component genomes
+    my %gdb_map = map { $_->name => $_->dbID } (grep {not $_->genome_component} @$genome_dbs);
```
muffato, line 73: Can you explain what has changed to require this ?

modules/t/MySQLImportHomologies.t
```diff
@@ -0,0 +1,110 @@
...
+            ],
+            'homology_id_start' => 11,
+        },
+    );
+
+    # check main homology data loading
+    my $hom_sth = $dba->dbc->prepare("SELECT 
+        homology_id, method_link_species_set_id, description, is_tree_compliant,
+        species_tree_node_id, gene_tree_node_id, gene_tree_root_id
+        FROM homology"
```
muffato, line 67: To make sure you get the rows in the right order
```suggestion
        FROM homology
        ORDER BY homology_id"
```

modules/t/MySQLImportHomologies.t
```diff
@@ -0,0 +1,110 @@
...
+                "$test_flatfile_dir/wga.tsv",
+                "$test_flatfile_dir/goc.tsv",
+                "$test_flatfile_dir/high_conf.tsv",
+            ],
+            'homology_id_start' => 11,
+        },
+    );
+
+    # check main homology data loading
+    my $hom_sth = $dba->dbc->prepare("SELECT
```
muffato, line 64: Comment: you could have uses `selectall_arrayref` or the `sql_helper->execute` here and below, then you don't need to call `execute` and `fetchall_arrayref`

modules/t/MySQLImportHomologies.t
```diff
@@ -0,0 +1,110 @@
...
+        ['11', '1234', 'ortholog_one2one',  1, '40133018', '14624903', '1377777'],
+        ['12', '1234', 'ortholog_one2many', 0, '40133016', '20937479', '2304777'],
+        ['13', '1234', 'ortholog_one2one',  1, '40133018', '15053143', '1631777'],
+    ];
+    is_deeply($homology_data, $exp_hom_data, 'homology data loaded correctly');
+
+    # check homology attribute data loading
+    my $attr_sth = $dba->dbc->prepare("SELECT 
+        goc_score, wga_coverage, is_high_confidence
+        FROM homology"
```
muffato, line 82: To make sure you get the rows in the right order
```suggestion
        FROM homology
        ORDER BY homology_id"
```

modules/t/MySQLImportHomologies.t
```diff
@@ -0,0 +1,110 @@
...
+
+    my $exp_attr_data = [
+        ['0',   undef,   1],
+        ['50',  '95.50', 1],
+        [undef, '0.00',  0],
+    ];
+    is_deeply($attrib_data, $exp_attr_data, 'homology attributes loaded correctly');
+
+    # check homology_member data
+    my $hom_mem_sth = $dba->dbc->prepare("SELECT * FROM homology_member");
```
muffato, line 95: To make sure you get the rows in the right order
```suggestion
    my $hom_mem_sth = $dba->dbc->prepare("SELECT * FROM homology_member ORDER BY homology_id, gene_member_id");
```

modules/t/MySQLImportHomologies.t
```diff
@@ -0,0 +1,110 @@
...
+    $hom_mem_sth->execute();
+    my $hom_mem_data = $hom_mem_sth->fetchall_arrayref;
+
+    my $exp_hom_mem_data = [
+        ['11', '607122', '759483',  '20M',     97, 83, 86],
+        ['11', '806596', '1021296', '10M5D5M', 99, 85, 87],
+        ['12', '694925', '875992',  '20M',     99, 84, 89],
+        ['12', '681661', '858054',  '10M5D5M', 99, 84, 89],
```
muffato, line 103: To match the `ORDER BY` clause
```suggestion
        ['12', '681661', '858054',  '10M5D5M', 99, 84, 89],
        ['12', '694925', '875992',  '20M',     99, 84, 89],
```

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthologyStats.pm
carlacummins: When I made this change, the file didn't have the `genome_db_id`, only the species name. It wasn't until a little later I discovered another analysis that required `genome_db_id` to be there too, so it was included then. Didn't seem worth changing this back as it's a lot more concise this way.

carlacummins:
> What a large piece of work !

Yes - apologies again for breaking my own rule with the PRs. I had planned to submit the orthotree & related as 1 PR, and the move to `mysqlimport` in another, but I ran out of time to create the intermediate state. Thanks for all of your comments - they were really helpful!

modules/Bio/EnsEMBL/Compara/RunnableDB/Flatfiles/MySQLImportHomologies.pm
```diff
@@ -0,0 +1,164 @@
...
+    my $user = $this_dbc->username;
+    my $pass = $this_dbc->password;
+    my $host = $this_dbc->host;
+    my $port = $this_dbc->port;
+    my $dbname = $this_dbc->dbname;
+    my $import_cmd = join(' ', 
+        "mysqlimport --host=$host --port=$port --user=$user --password=$pass",
+        "--local --lock-tables=0 --ignore --fields-terminated-by=','",
+        $dbname
+    );
```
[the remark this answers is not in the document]
JAlvarezJarreta: I didn't test the `bulk_insert` method as it stands. I think the ideal scenario, if all the data is already loaded in memory, would be to "pipe" it to a `LOAD DATA` (or `mysqlimport`) command, since dumping it into files and then reading them was about a 5% slower in my tests with `mysqldump ... > file.txt` + `mysqlimport ... file.txt`.

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthologyStats.pm
muffato: I see. I was concerned because the way the polyploids are handled differs. TBH I can't tell how the entire PR is working on polyploids, but I guess we can see things live in production

modules/Bio/EnsEMBL/Compara/RunnableDB/Flatfiles/MySQLImportHomologies.pm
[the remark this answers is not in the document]
muffato: The pipe into `LOAD DATA` is a very good idea. Can I create a Jira ticket or do you want to do it ?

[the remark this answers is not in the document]
muffato: Well, actually, this trick could be used here ! Instead of writing a file and running `mysqlimport`, we could pipe into `LOAD DATA`

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthoTree.pm
```diff
@@ -566,5 +590,83 @@ sub check_homology_consistency {
...
+            die sprintf(
+                "%s is not a canonical member (line %d)",
+                ($st_id, $c)
+            );
+        } elsif ( $hom_gene_member->canonical_member_id != $row->{homology_seq_member_id} ) {
+            die sprintf(
+                "%s is not a canonical member (line %d)",
+                ($hom_st_id, $c)
+            );
+        }
```
muffato, line 654: I know this is a copy of the previous SQLHealthchecks, but I think it should now be done differently.

The first problem is performance. There will be two `fetch_by_dbID` for each homology row, so about n² in total. Independent small queries are a killer. That's why many runnables have switch to using `Utils::Preloader` and `fetch_all_by_dbID_list`. I think it's a pity to replace a n² bottleneck with another one.

The second is usefulness / redundancy. There are still some SqlHealtchecks to check we only have canonical seq_members in the database (`only_canonical` is set to 1) and the trees are only made of canonical members. It is unlikely that the `ortho_tree` will come up with some non-canonical members.

The third problem is that this Runnable doesn't really care whether the members are canonical or not. It infers homologies from a reconciled gene tree, and is expected to list homologies that match the trees, that's all.

There is still value in checking that the file is sane and member_ids are valid. I would suggest to get the list of seq_members of the gene tree, and check that all the seq_members listed in the file are part of those. This commit should work: https://github.com/Ensembl/ensembl-compara/commit/111063e4e1836544d402a48d53bbe61fe143f904

modules/Bio/EnsEMBL/Compara/RunnableDB/Flatfiles/MySQLImportHomologies.pm
```diff
@@ -0,0 +1,164 @@
...
+
+    # open homology file
+    open(my $hom_fh, '<', $filename) or die "Cannot open $filename for reading";
+    my $header_line = <$hom_fh>;
+    my @header_cols = split(/\s+/, $header_line);
+
+    # open output files - one per table in tmp space
+    my $homology_csv = $self->worker_temp_directory . '/homology.txt';
+    open(my $h_csv, '>', $homology_csv) or die "Cannot open $homology_csv for writing";
+    my $homology_member_csv = $self->worker_temp_directory . '/homology_member.txt';
```
JAlvarezJarreta, line 89: Just in case (since this has happened in the past and I cannot answer the question myself), is there a chance 2 jobs for this runnable will end up in the same node/worker and there will be a race condition in these files?

muffato: `$self->worker_temp_directory` is unique for each worker, so I would say we're good

modules/t/test-genome-DBs/mysqlimport_test/compara/table.sql
```diff
@@ -0,0 +1,692 @@
...
+  `description` text,
+  `dnafrag_id` bigint(20) unsigned DEFAULT NULL,
+  `dnafrag_start` int(10) DEFAULT NULL,
+  `dnafrag_end` int(10) DEFAULT NULL,
+  `dnafrag_strand` tinyint(4) DEFAULT NULL,
+  `display_label` varchar(128) DEFAULT NULL,
+  PRIMARY KEY (`gene_member_id`),
+  UNIQUE KEY `stable_id` (`stable_id`),
+  KEY `taxon_id` (`taxon_id`),
+  KEY `source_name` (`source_name`),
```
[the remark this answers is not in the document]
muffato: Yes the schemas should be identical. I hope this could be checked on Travis, cf ENSCOMPARASW-3568

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OrthoTree.pm
carlacummins: Yeah, I'm happy to incorporate that commit. Not sure how to go about it? At this point, we kinda just need this PR to be merged so we can start production 😅

modules/Bio/EnsEMBL/Compara/RunnableDB/Flatfiles/FindHomologyIDRange.pm
```diff
@@ -0,0 +1,85 @@
...
+
+sub fetch_input {
+    my $self = shift;
+
+    # grab range start from pipeline_wide_parameters
+    my $range_start = $self->param('homology_id_range_start');
+    return if defined $range_start;
+
+    # initialise the range if it doesn't exist yet (i.e. first job)
+    $range_start = $self->param_required('offset');
```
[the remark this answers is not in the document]
carlacummins: Ah yes, in fact, the entire `fetch_input` sub doesn't do anything. I meant to remove it. Thanks.

muffato:
`git merge --ff-only <commit_id>`

modules/Bio/EnsEMBL/Compara/RunnableDB/Flatfiles/MySQLImportHomologies.pm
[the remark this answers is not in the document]
carlacummins: Can this be an extra ticket later? I think there's enough happening in this PR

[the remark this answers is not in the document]
carlacummins: Also, I'm not sure if it would handle the 2 different tables at the same time

carlacummins:
> `git merge --ff-only <commit_id>`

This didn't work (`fatal: Not possible to fast-forward, aborting.`). Used `git cherry-pick` instead.

modules/Bio/EnsEMBL/Compara/RunnableDB/Flatfiles/MySQLImportHomologies.pm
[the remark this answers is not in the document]
muffato: Yes. the ticket is meant for a later development, not this PR :)

PR #208, opened 2020-06-22
update the test REST server script to take into account EPO alignment in plant

dthybert:
## Description
The REST testing script was not working on the plant EPO test of the phyloxml output because the access to a species node was not made generic. This has been corrected by making this part generic to any species/divisions.

**Related JIRA tickets:**
- ENSCOMPARASW-3507

## Overview of changes
 The issue was caused by a hard coding of the access to the phyloxml output corresponding to  vertebrates species used for the test.  To solve this problem I have modified and renamed a function that search for the existence of a species production name in the output of a phyloxml. The modified function return now the phyloxml node associated to the species and it is now used used in the part testing EPO REST endpoints making it completely generic to any species/divisions. In addition to this, the EPO phyloxml output has been extended with the with the "property" node containing the production name.

#### Change #1:  Verify_Compara_REST_Endpoints.pl
     #1.1_ rename of the function by fetch_species_node to better describe  its new role: 
     #1.2 make the function fetch_species_node to recursively search for the species node having a given production name
     #1.3 Update the gene tree testing part as it was using the rename function (l346) and add a test for "defined" to test for the existence of the given species 
    #1.4  corrected a small bug in the creation of the REST url for non vertebrate divisions.
    #1.5 Add a call to fetch_species_node in the part of the script testing the EPO phyloxml output. 
  
#### Change #2: GenomicAlignTreePhyloXMLWriter.pm
   #1.1 add the "property" node to the alignment phyloxml output.

## Testing
- test of the extension of the phyloxml output with the property node:
    --> call to the EPO endpoints and look for the existence of the "property" node in the species node: 
/alignment/region/taeniopygia_guttata/3:106040329-106040379?content-type=application/json;species_set_group=sauropsids

- test the rest script:
  --> run the rest test script on all endpoint and and the test was successful if the rest script had no failures. 
   This tested the generic function and the non-regression on the gene tree part of the testing script..

## Notes
_Optional extra information._

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/Graph/GenomicAlignTreePhyloXMLWriter.pm
+++ b/modules/Bio/EnsEMBL/Compara/Graph/GenomicAlignTreePhyloXMLWriter.pm
@@ -192,6 +192,12 @@ sub _genomicaligntree_body {
     if ($all_genomic_aligns->[0]->genome_db->name ne "ancestral_sequences") {
       my $gdb = $all_genomic_aligns->[0]->genome_db;
       $self->_write_genome_db($gdb);
+      #add property element to provide the production name
+      $w->dataElement('property', $gdb->name,
+        'datatype' => 'xsd:string',
+        'ref' => 'Compara:genome_db_name',
+        'applies_to' => 'clade'
+      );
     }
 
     if ($compact_alignments) {
--- a/scripts/production/Verify_Compara_REST_Endpoints.pl
+++ b/scripts/production/Verify_Compara_REST_Endpoints.pl
@@ -96,15 +96,17 @@ elsif($division eq "plants"){
     $gene_member_id           = "AT3G52430";
     $gene_tree_id             = "EPlGT00140000000744";
     $alignment_region         = "1:8001-18000:1";
-    $lastz_alignment_region   = $alignment_region;
+    $lastz_alignment_region   = "1:12928-15180";
 
-    $species_1                = "arabidopsis_thaliana";
+    $species_1                = "oryza_sativa";
     $species_2                = "vitis_vinifera";
-    $species_3                = "oryza_barthii";
+    $species_3                = "arabidopsis_thaliana";
 
-    $taxon_1                  = 3702;#arabidopsis_thaliana
+    $taxon_1                  = 39947;#oryza_sativa
     $taxon_2                  = 29760;#vitis_vinifera
-    $taxon_3                  = 65489;#oryza_barthii
+    $taxon_3                  = 3702;#arabidopsis_thaliana
+
+    $species_set_group        = "rice";
 
     $gene_symbol              = "PAD4";
     $homology_type            = 'orthologues';
@@ -112,7 +114,6 @@ elsif($division eq "plants"){
     
     $extra_params             = 'compara=plants';
     $skip_families            = 1;
-    $skip_epo                 = 1;
     $skip_cactus              = 1;
 }
 elsif($division eq 'pan' or $division eq 'pan_homology'){
@@ -490,7 +491,7 @@ try{
         print "\nTesting GET EPO alignment region\/\:species\/\:region \n\n";
     
         my $ext = "/alignment/region/$species_1/$alignment_region?species_set_group=$species_set_group";
-        $ext .= "?$extra_params" if $extra_params;
+        $ext .= ";$extra_params" if $extra_params;
 
         $responseIDGet = $browser->get($server.$ext, { headers => { 'Content-type' => 'application/json' } } );
         ok($responseIDGet->{success}, "Check json validity");
@@ -502,8 +503,8 @@ try{
         $phyloXml = process_phyloXml_get($server.$ext.';content-type=text/x-phyloxml;aligned=0'.($extra_params ? ";$extra_params" : ''));
         #print Dumper $phyloXml;
         ok($phyloXml->{phylogeny}->{clade}->{sequence}->{mol_seq}->{is_aligned} == 0, "Check get alignment region and unaligned sequences");
-    
-        $jsontxt = process_json_get($server."/alignment/region/$species_1/$lastz_alignment_region?content-type=application/json;display_species_set=$species_1".($extra_params ? ";$extra_params" : ''));
+
+        $jsontxt = process_json_get($server."/alignment/region/$species_1/$lastz_alignment_region?content-type=application/json;display_species_set=$species_1;species_set_group=$species_set_group".($extra_params ? ";$extra_params" : ''));
         ok($jsontxt->[0]->{alignments}[0]->{species} eq $species_1, "Check alignment region display_species_set option validity");
     }
     
--- a/scripts/production/Verify_Compara_REST_Endpoints.pl
+++ b/scripts/production/Verify_Compara_REST_Endpoints.pl
@@ -264,25 +264,34 @@ sub fetch_leaf_hash_from_json {
     return $input_json;
 }
 
-sub verify_xml_leaf {
+
+# Fetch a species node using the production name node from a phyloxml input
+# if the searched species is not present it returns undef
+# so the presence of species can be tested by defined.
+sub fetch_species_node {
     my ($this_node, $species_name) = @_;
     my $nodes = ref($this_node) eq 'ARRAY' ? $this_node : [$this_node];
-
     foreach my $node ( @$nodes ) {
         if ( exists $node->{property} ) {
-            return 1 if ($node->{property}->{content} // '') eq $species_name;
+            return $node if (($node->{property}->{content} // '') eq $species_name);
         }
-
         if ( exists $node->{clade} ) {
-            my $verify_recursive = verify_xml_leaf($node->{clade}, $species_name);
-            return 1 if $verify_recursive;
+            my $species_node = fetch_species_node($node->{clade}, $species_name);
+            return $species_node if defined $species_node;
         } else {
-            foreach my $value ( values %$node ) {
-                return 1 if ($value->{property}->{content} // '') eq $species_name;
+            # in case we are on an ancestral or species node
+            foreach my $chd_node ( values %$node ){
+                if (exists $chd_node->{property}){
+                    return $chd_node if (($chd_node->{property}->{content} // '') eq $species_name);
+                }
+                if (exists $chd_node->{clade}){
+                    my $species_node = fetch_species_node($chd_node->{clade}, $species_name);
+                    return $species_node if defined $species_node;
+                }
             }
         }
     }
-    return 0;
+    return undef;
 }
 
 #Compara currently have no POST requests. For future purposes.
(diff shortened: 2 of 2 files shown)


CristiGuijarro:
This branch needs to be either rebased to `release/100` or compared against the `release/101`/`master` branch in the PR

CristiGuijarro:
Would it be possible to kindly replace the (admittedly annoying) trailing whitespace lines? There are a few reasons why it would be good to do so, one of the most important is the "git blame" - you can actually see this when viewing the file and selecting the "blame" tab. It would also make it easier to view the changes on screen. I think the changes are okay, and easy to understand in `GenomicAlignTreePhyloXMLWriter.pm`, but at the moment I cannot fully determine that my very human eyes are not missing something in `Verify_Compara_REST_Endpoints.pl` :slightly_smiling_face:

muffato:
There are some major problems:
1. The goal of the ticket was to test the EPO alignment. You **have not** done that. The script still doesn't check for EPO, see https://github.com/Ensembl/ensembl-compara/pull/208/files#diff-4e4e5131c3ef86830ae04f5ee9a6422cR115 `skip_epo` is stil set to 1
2. You have broken the "check prune species validity" test of the `genetree/id/:id` endpoint. The e101 script works on the live server, your script doesn't. (this endpoint doesn't use GenomicAlignTreePhyloXMLWriter)

muffato:
@dthybert. Can you please clean the whitespace changes like we asked ?

scripts/production/Verify_Compara_REST_Endpoints.pl
```diff
@@ -284,6 +285,35 @@ sub verify_xml_leaf {
     return 0;
 }
 
+#Fetch a species node using the productnion name node from a phyloxml input
```
CristiGuijarro, line 288:
```suggestion
# Fetch a species node using the production name node from a phyloxml input
```

muffato:
I still have a problem with this change. I have downloaded the script and the updated module, and the script fails
```
Testing GET EPO alignment region/:species/:region 

ok 31 - Check json validity
ok 32 - Check phyloXml validity
http://127.0.0.1:3002//alignment/region/oryza_sativa/3:1060403-1060503?species_set_group=rice;compara=plants;content-type=text/x-phyloxml;aligned=0
ok 33 - Check get alignment region and unaligned sequences
caught error: Can't use string ("0") as an ARRAY ref while "strict refs" in use at ./Verify_Compara_REST_Endpoints.pl line 538.
1..33
```

Test 33 is OK, so the problem is test 34: _Check alignment region display_species_set option validity_

In fact, if I run the script _without_ the new `fetch_species_node` function and the updated `GenomicAlignTreePhyloXMLWriter` (i.e. just setting the sample region), I get the same output: test 33 passes, but test 34 fails.

dthybert:
> I still have a problem with this change. I have downloaded the script and the updated module, and the script fails
> 
> ```
> Testing GET EPO alignment region/:species/:region 
> 
> ok 31 - Check json validity
> ok 32 - Check phyloXml validity
> http://127.0.0.1:3002//alignment/region/oryza_sativa/3:1060403-1060503?species_set_group=rice;compara=plants;content-type=text/x-phyloxml;aligned=0
> ok 33 - Check get alignment region and unaligned sequences
> caught error: Can't use string ("0") as an ARRAY ref while "strict refs" in use at ./Verify_Compara_REST_Endpoints.pl line 538.
> 1..33
> ```
> 
> Test 33 is OK, so the problem is test 34: _Check alignment region display_species_set option validity_
> 
> In fact, if I run the script _without_ the new `fetch_species_node` function and the updated `GenomicAlignTreePhyloXMLWriter` (i.e. just setting the sample region), I get the same output: test 33 passes, but test 34 fails.

This is bug in the REST end point which is notr related ot the script. if you use the url response: 
 https://rest.ensembl.org//alignment/region/oryza_sativa/3:1060403-1060503?content-type=application/json;display_species_set=oryza_sativa;compara=plants 

you will have the following answer:
{"error":"No method_link_specices_set found for method EPO and species_set_group mammals "}

I created a ticket for this:
https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-3729

CristiGuijarro:
Is that url correct? [https://rest.ensembl.org//alignment/region/oryza_sativa/3:1060403-1060503?content-type=application/json;species_set_group=rice;compara=plants](https://rest.ensembl.org//alignment/region/oryza_sativa/3:1060403-1060503?content-type=application/json;species_set_group=rice;compara=plants) works. The `display_species_set` is supposed to be a subset of species, but you have provided only one in the url, so it cannot form an alignment?

muffato:
As I said in the Jira ticket, it's not a bug. To get the rice alignment, you need to set the species-set to `rice`

scripts/production/Verify_Compara_REST_Endpoints.pl
```diff
@@ -501,9 +531,10 @@ sub verify_xml_leaf {
         print $server.$ext.';content-type=text/x-phyloxml;aligned=0' . "\n";
         $phyloXml = process_phyloXml_get($server.$ext.';content-type=text/x-phyloxml;aligned=0'.($extra_params ? ";$extra_params" : ''));
         #print Dumper $phyloXml;
-        ok($phyloXml->{phylogeny}->{clade}->{sequence}->{mol_seq}->{is_aligned} == 0, "Check get alignment region and unaligned sequences");
+        my $species_node = fetch_species_node($phyloXml->{phylogeny}, $species_1);
+        ok($species_node->{sequence}->{mol_seq}->{is_aligned} == 0, "Check get alignment region and unaligned sequences");
     
-        $jsontxt = process_json_get($server."/alignment/region/$species_1/$lastz_alignment_region?content-type=application/json;display_species_set=$species_1".($extra_params ? ";$extra_params" : ''));
+        $jsontxt = process_json_get($server."/alignment/region/$species_1/$lastz_alignment_region?content-type=application/json;display_species_set=$species_1;species_set_group=$species_set_group".($extra_params ? ";$extra_params" : ''));
```
muffato, line 537: Can you check it still works for Vertebrates ? For Vertebrates the endpoint used to query the mammals alignment (default species set), but now it is querying the primates one

dthybert: I did the test on both vertebrates and plants to make sure I did not introduced more regression before pushing and it did work

dthybert:
> The change in `GenomicAlignTreePhyloXMLWriter` and the new `fetch_species_node` function are actually not needed. The script works fine just by adding the sample information, changing the `?` to `;` and adding the `species_set_group` parameter.

I am a bit confused, we discussed about it and  when you suggested me to make these changes (enrichment of phyloxml + the recursive function), I have even told you that this may be a bit overkilling doing this in this way, especially for a testing script. You convinced me because you told me it would be beneficial for our user to enrich the phyloXML output (I also thought you suggested this for training purposes to better understand compara infrastructure, so I did it).  I feel we had a miscommunication here...

If I summarise the benefit of this development : 1- extend the phyloxml output for the alignment with additional information which is good for our users 2- the test is more thorough as it test now the presence of a species requested in the output . So now that this is implemented I think we should keep it, unless you think this can break part of compara.

muffato:
> I am a bit confused, we discussed about it and when you suggested me to make these changes (enrichment of phyloxml + the recursive function), I have even told you that this may be a bit overkilling doing this in this way, especially for a testing script. You convinced me because you told me it would be beneficial for our user to enrich the phyloXML output (I also thought you suggested this for training purposes to better understand compara infrastructure, so I did it). I feel we had a miscommunication here...

You should remember that I suggested this solution because _you_ told me that the phyloXML test didn't pass. But actually it passed, so these two changes are not needed. 
Yes there's clearly been so much misunderstanding on this work. That's why I later created ENSCOMPARASW-3725 to make it clear what needs to be done. The other thing you've done, ENSCOMPARASW-3507, is not characterised correctly: it was sold as a bugfix, but it's actually a new feature.

muffato:
Really, please *do not* change the indentation / blank lines / whitespace. This makes it harder to review

scripts/production/Verify_Compara_REST_Endpoints.pl
```diff
@@ -263,25 +264,34 @@ sub fetch_leaf_hash_from_json {
...
         if ( exists $node->{clade} ) {
-            my $verify_recursive = verify_xml_leaf($node->{clade}, $species_name);
-            return 1 if $verify_recursive;
-        } else {
-            foreach my $value ( values %$node ) {
-                return 1 if ($value->{property}->{content} // '') eq $species_name;
+            my $species_node = fetch_species_node($node->{clade}, $species_name);
+            return $species_node if defined $species_node;
+        }
+        else{# in case we are on an ancestral or species node
```
muffato, line 282: For example, here it would have been clearer to leave the line with the `else` as it is, and add a comment on the next line
```suggestion
        } else {
            # in case we are on an ancestral or species node
```

modules/Bio/EnsEMBL/Compara/Graph/GenomicAlignTreePhyloXMLWriter.pm
```diff
@@ -192,6 +192,13 @@ sub _genomicaligntree_body {
     if ($all_genomic_aligns->[0]->genome_db->name ne "ancestral_sequences") {
       my $gdb = $all_genomic_aligns->[0]->genome_db;
       $self->_write_genome_db($gdb);
+
```
muffato, line 195: Don't forget to remove this one
```suggestion

```

dthybert:
@muffato,  what do you mean by redoing a commit. Is it erasing the last ones and starting again with only one having the modification cleaned up (I will update the white spaces to be removed).
@CristiGuijarro is it the same commends you suggested me a few days ago? : 
    - rebase to the commit cb15407   (git rebase -i HEAD~4) and adding drop to each commit after cb15407
    - then a git push origin dthybert --force (I do not know why we did this actually)
    - then add and commit the file with the new changes (cleaned)

CristiGuijarro:
> @CristiGuijarro is it the same commends you suggested me a few days ago? :
> - rebase to the commit [cb15407](https://github.com/Ensembl/ensembl-compara/commit/cb1540797121936fdd428cf93142e32404fb0428) (git rebase -i HEAD~4) and adding drop to each commit after [cb15407](https://github.com/Ensembl/ensembl-compara/commit/cb1540797121936fdd428cf93142e32404fb0428)
> - then a git push origin dthybert --force (I do not know why we did this actually)
> - then add and commit the file with the new changes (cleaned)

Yes, you could do it this way. You could skip the in between force push, and just do it at the end though, once you have removed the `verify_xml_leaf` code again.

muffato:
> @muffato, what do you mean by redoing a commit. Is it erasing the last ones and starting again with only one having the modification cleaned up (I will update the white spaces to be removed).

Yes start again, but you can choose how many commits you make, t doesn't have to be only 1

PR #210, opened 2020-06-24
Improve Utils/CopyData.pm and populate_new_database.pl

JAlvarezJarreta:
## Description

Our script `populate_new_database.pl` didn't work as expected during e101, with several `server has gone away` errors midway through the copy process. Furthermore, this script is quite slow and any improvements in its performance would be much appreciated.

**Related JIRA tickets:**
- ENSCOMPARASW-3542

## Overview of changes

Performance improvements have been added based on the destination table's engine in `populate_new_database.pl` script and `copy_data` method.

#### Change 1
- Regarding the disconnection issues in `populate_new_database.pl`, `disconnect_if_idle` methods have been added to the `copy_data` method.

#### Change 2
- `LOAD DATA` has been proven to be much faster than an INSERT of multiple values in both MyISAM and InnoDB tables. Furthermore, I have found a data transfer method via `mysql` that works for both binary and non-binary columns. Thus, all the `copy_data_*` methods have been removed and only `copy_data` has remained as a global approach.

#### Change 3
- `copy_data` does not return the number of rows inserted any more.

#### Change 4
- `copy_table` is now based on `copy_data`, as the latter allows more configuration than the former, so the code can be reused.

## Testing
Regarding performance, my preliminary tests have shown that the query + `LOAD DATA` approach provides a reduction of at least 35% on the time cost. For instance, copying the `constrained_element` table (e100) to a new empty table took 45 minutes with the old code, less than 18 minutes with the proposed update, and for `gene_tree_node_tag`, from 16 minutes to 9 minutes.

However, I have tested thoroughly that both `copy_data` and `copy_table` work, but not the rest of the changes.

## Notes
- All the runnables and pipeline configs that use `Utils/CopyData.pm` have been updated accordingly.
- This solution only addresses part of the problem mentioned in the JIRA ticket. More work should be done in the future.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/EpoLowCoverage/ImportAlignment.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/EpoLowCoverage/ImportAlignment.pm
@@ -160,7 +160,7 @@ sub importAlignment {
 
     copy_data($self->param('from_dbc'), $self->compara_dba->dbc,
               "genomic_align_tree",
-              $gat_sql, 1, 'skip_disable_vars', $self->debug);
+              $gat_sql, 1, 0, $self->debug);
     print "\n\n" if $self->debug;
 
 
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/EpoLowCoverage/ImportAlignment.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/EpoLowCoverage/ImportAlignment.pm
@@ -113,23 +113,23 @@ sub importAlignment {
     }
 
     #Copy the species_set_header
-    my $ssh_rows = copy_data($self->param('from_dbc'), $self->compara_dba->dbc,
+    copy_data($self->param('from_dbc'), $self->compara_dba->dbc,
 	      "species_set_header",
 	      "SELECT species_set_header.* FROM species_set_header JOIN method_link_species_set USING (species_set_id) WHERE method_link_species_set_id = $mlss_id",
-          0, 0, 0, $self->debug);
+          0, 0, $self->debug);
     print "\n\n" if $self->debug;
 
     #Copy the method_link_species_set
-    my $mlss_rows = copy_table($self->param('from_dbc'), $self->compara_dba->dbc,
+    copy_table($self->param('from_dbc'), $self->compara_dba->dbc,
 	      "method_link_species_set",
 	      "method_link_species_set_id = $mlss_id");
     print "\n\n" if $self->debug;
 
     #Copy the species_set
-    my $ss_rows = copy_data($self->param('from_dbc'), $self->compara_dba->dbc,
+    copy_data($self->param('from_dbc'), $self->compara_dba->dbc,
 	      "species_set",
 	      "SELECT species_set.* FROM species_set JOIN method_link_species_set USING (species_set_id) WHERE method_link_species_set_id = $mlss_id",
-          0, 0, 0, $self->debug);
+          0, 0, $self->debug);
     print "\n\n" if $self->debug;
 
     #copy genomic_align_block table
@@ -139,9 +139,9 @@ sub importAlignment {
     } else {
         $gab_sql = "SELECT * FROM genomic_align_block WHERE method_link_species_set_id = $mlss_id";
     }
-    my $gab_rows = copy_data($self->param('from_dbc'), $self->compara_dba->dbc,
+    copy_data($self->param('from_dbc'), $self->compara_dba->dbc,
               "genomic_align_block",
-              $gab_sql, 0, 0, 0, $self->debug);
+              $gab_sql, 0, 0, $self->debug);
     print "\n\n" if $self->debug;
 
     #copy genomic_align_tree table
@@ -158,9 +158,9 @@ sub importAlignment {
                     "ORDER BY node_id DESC";
     }
 
-    my $gat_rows = copy_data($self->param('from_dbc'), $self->compara_dba->dbc,
+    copy_data($self->param('from_dbc'), $self->compara_dba->dbc,
               "genomic_align_tree",
-              $gat_sql, 1, 0, 'ignore_fks', $self->debug);
+              $gat_sql, 1, 'skip_disable_vars', $self->debug);
     print "\n\n" if $self->debug;
 
 
@@ -176,9 +176,9 @@ sub importAlignment {
                     " FROM genomic_align JOIN dnafrag USING (dnafrag_id)".
                     " WHERE method_link_species_set_id = $mlss_id $ancestral_dbID_constraint";
     }
-    my $ga_rows = copy_data($self->param('from_dbc'), $self->compara_dba->dbc,
+    copy_data($self->param('from_dbc'), $self->compara_dba->dbc,
               "genomic_align",
-              $ga_sql, 1, 0, 0, $self->debug);
+              $ga_sql, 1, 0, $self->debug);
 }
 
 
--- a/modules/Bio/EnsEMBL/Compara/Utils/CopyData.pm
+++ b/modules/Bio/EnsEMBL/Compara/Utils/CopyData.pm
@@ -328,7 +328,7 @@ sub copy_data {
     }
 
     # Get table's engine to optimise the copy process
-    my $table_engine;
+    my ($table_engine, $init_command);
     unless ($skip_disable_vars) {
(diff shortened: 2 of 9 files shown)


modules/Bio/EnsEMBL/Compara/Utils/CopyData.pm
```diff
@@ -270,207 +270,106 @@ sub clear_copy_data_cache {
...
+    unless ($skip_disable_vars) {
+        $table_engine = $to_dbc->db_handle->selectrow_hashref("SHOW TABLE STATUS WHERE Name = '$table_name'")->{Engine};
+        # Speed up writing of data by disabling certain variables, write the data, then enable them back
+        print "DISABLE VARIABLES\n" if $debug;
+        if ($table_engine eq 'MyISAM') {
+            $to_dbc->do("ALTER TABLE `$table_name` DISABLE KEYS");
+        } else {
+            $to_dbc->do("SET AUTOCOMMIT = 0");
+            $to_dbc->do("SET FOREIGN_KEY_CHECKS = 0");
+            $to_dbc->do("SET UNIQUE_CHECKS = 0");
```
muffato, line 348: I think these are per-session variables, so won't apply to the new session created below (`mysql --host=$to_host --port=$to_port --user=$to_user`)

JAlvarezJarreta: I keep forgetting about the life span of these variables. Solved in the next commit.

muffato:
Why did you remove `copy_data_pp` ? It was there to provide a copy_data method that uses the given DBConnections, i.e. will fit in an existing transaction, can be rolled-back, etc.

JAlvarezJarreta:
> Why did you remove `copy_data_pp` ? It was there to provide a copy_data method that uses the given DBConnections, i.e. will fit in an existing transaction, can be rolled-back, etc.

I did it because its performance was worse than the new method (and those features you mention are not "useful" with MyISAM tables). Besides, it wasn't being used anywhere in our codebase. But happy to put it back if you want :)

muffato:
> > Why did you remove `copy_data_pp` ? It was there to provide a copy_data method that uses the given DBConnections, i.e. will fit in an existing transaction, can be rolled-back, etc.
> 
> I did it because its performance was worse than the new method (and those features you mention don't work with MyISAM tables). Besides, it wasn't being used anywhere in our codebase. But happy to put it back if you want :)

Yes please, it's not a matter of performance. It's the only way of doing copy_data within an existing transaction. It was actually used (indirectly, via a `copy_data` call with the `ignore_fks` flag turned on) since @carlacummins adding some foreign keys in the species_tree / genomic_align_tree tables

JAlvarezJarreta:
I have found only one case of `ignore_fks` turned on via `copy_data`: [ImportAlignment.pm#L161](https://github.com/Ensembl/ensembl-compara/blob/release/102/modules/Bio/EnsEMBL/Compara/RunnableDB/EpoLowCoverage/ImportAlignment.pm#L161) (but since it is not inside a transaction I have used the new `copy_data` instead).

scripts/pipeline/populate_new_database.pl
```diff
@@ -807,18 +811,20 @@ sub copy_synteny_data {
   my ($old_dba, $new_dba, $method_link_species_sets) = @_;
 
   # Keys are disabled / enabled only once for the whole loop
-  $new_dba->dbc->do("ALTER TABLE `synteny_region` DISABLE KEYS");
-  $new_dba->dbc->do("ALTER TABLE `dnafrag_region` DISABLE KEYS");
+  my $sr_engine = $new_dba->dbc->db_handle->selectrow_hashref("SHOW TABLE STATUS WHERE Name = 'synteny_region'")->{Engine};
```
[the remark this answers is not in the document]
muffato: I would move `_get_table_engine` to our DBAdaptor class, but that's a minor comment. Doesn't have to be done in this PR, I might do it later as well :)

modules/Bio/EnsEMBL/Compara/RunnableDB/EpoLowCoverage/ImportAlignment.pm
```diff
@@ -158,9 +158,9 @@ sub importAlignment {
                     "ORDER BY node_id DESC";
     }
 
-    my $gat_rows = copy_data($self->param('from_dbc'), $self->compara_dba->dbc,
+    copy_data($self->param('from_dbc'), $self->compara_dba->dbc,
               "genomic_align_tree",
-              $gat_sql, 1, 0, 'ignore_fks', $self->debug);
+              $gat_sql, 1, 'skip_disable_vars', $self->debug);
```
muffato, line 163: I think there is a problem here. This is the `copy_data` call that was using `copy_data_pp` + `SET FOREIGN_KEY_CHECKS = 0` (`ignore_fks` was set) because the internal foreign keys of the `genomic_align_tree` table are otherwise a nightmare to sort out. With `skip_disable_vars` set, the new `copy_data` *won't* do `SET FOREIGN_KEY_CHECKS = 0`, so I think it shouldn't be set ?
```suggestion
              $gat_sql, 1, 0, $self->debug);
```

muffato:
1. Can you explain why we shouldn't do ENABLE / DISABLE when loading in an InnoDB table ?
2. (minor) The _This process might be very slow_ warnings in`copy_data.pl` can be removed

JAlvarezJarreta:
> 1. Can you explain why we shouldn't do ENABLE / DISABLE when loading in an InnoDB table ?
ENABLE and DISABLE KEYS is not implemented in the InnoDB Storage Engine. Citing the [source](https://dev.mysql.com/doc/refman/5.6/en/alter-table.html):
>For MyISAM tables, key updating can be controlled explicitly. Use ALTER TABLE ... DISABLE KEYS to tell MySQL to stop updating nonunique indexes. Then use ALTER TABLE ... ENABLE KEYS to re-create missing indexes. MyISAM does this with a special algorithm that is much faster than inserting keys one by one, so disabling keys before performing bulk insert operations should give a considerable speedup. Using ALTER TABLE ... DISABLE KEYS requires the INDEX privilege in addition to the privileges mentioned earlier.

> 2. (minor) The _This process might be very slow_ warnings in`copy_data.pl` can be removed
:+1:

modules/Bio/EnsEMBL/Compara/RunnableDB/EpoLowCoverage/ImportAlignment.pm
muffato: Why did you dismiss this change ?

modules/Bio/EnsEMBL/Compara/Utils/CopyData.pm
```diff
@@ -270,207 +263,100 @@ sub clear_copy_data_cache {
...
+    # Get table's engine to optimise the copy process
+    my ($table_engine, $init_command);
+    unless ($skip_disable_vars) {
+        $table_engine = $to_dbc->db_handle->selectrow_hashref("SHOW TABLE STATUS WHERE Name = '$table_name'")->{Engine};
+        # Speed up writing of data by disabling certain variables, write the data, then enable them back
+        print "DISABLE VARIABLES\n" if $debug;
+        if ($table_engine eq 'MyISAM') {
+            $to_dbc->do("ALTER TABLE `$table_name` DISABLE KEYS");
+        } else {
+            $init_command = "SET AUTOCOMMIT = 0; SET FOREIGN_KEY_CHECKS = 0; SET UNIQUE_CHECKS = 0;";
```
muffato, line 339: In https://support.tigertech.net/mysql-large-inserts they say we need a `COMMIT` after inserting the data. It's not needed here because `LOAD DATA` automatically does it ?

modules/Bio/EnsEMBL/Compara/RunnableDB/EpoLowCoverage/ImportAlignment.pm
JAlvarezJarreta: Because I misread the meaning of `'ignore_fks'`.

modules/Bio/EnsEMBL/Compara/Utils/CopyData.pm
JAlvarezJarreta: Must confess I did miss this bit, but it was working. I have found the reason [here](https://dev.mysql.com/doc/refman/5.6/en/implicit-commit.html):
> `SET autocommit = 1` (if the value is not already 1) implicitly ends any transaction active in the current session, as if you had done a COMMIT before executing the statement.

PR #213, opened 2020-07-06
Fix the GOC scores

JAlvarezJarreta:
## Description

GOC scores distributions were missing for e101. Furthermore, a bug in how polyploids are handled to calculate the GOC scores was generating only partial results. Both these issues are addressed in this PR.

**Related JIRA tickets:**
- ENSCOMPARASW-3611
- ENSCOMPARASW-3612

#### Change 1
- The GOC score distributions weren't generated correctly due to an error in the tag `n_goc_<score>`: the hash passed to the method was `goc_score => array_of_hom_mlss_ids` and it should `hom_mlss_id => goc_score` in `GeneOrderConservation.pm`.

#### Change 2
- The parallelisation of polyploids was generating a race condition in `output_file` as well as an overwriting of the `n_goc_<score>` tags. Since this analysis is quite fast (less than 2 minutes on average, with a maximum of less than 4 minutes for plants in e101), I have opted to remove the parallelisation and process all polyploids in a single job. This decision was also the least disruptive one in the current pipeline's graph.

#### Change 3
- Updated `GeneOrderConservation.t` and added/updated `HOMOEOLOGUES` test files.

## Testing
`GeneOrderConservation.t` passes successfully and the tags seem to be correctly there now for each MLSS.

## Notes
- Most of the code has just been "indented" inside a `foreach` loop, but GitHub doesn't seem to be able to detect it.


modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/GeneOrderConservation.pm
```diff
@@ -177,7 +177,11 @@ sub write_output {
     my $mlss        = $self->param('mlss');
 
     print "Writing n_goc_score to the database\n" if $self->debug;
-    $self->write_n_tag($mlss, 'goc', $goc_scores);
+    foreach my $score ( keys %$goc_scores ) {
+        my $floor_score = int($score / 25) * 25;
+        my $tag = sprintf('n_goc_%s', $floor_score // 'null');
```
muffato, line 182: Scores are all defined by construction

modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/GeneOrderConservation.pm
```diff
@@ -177,7 +177,11 @@ sub write_output {
     my $mlss        = $self->param('mlss');
 
     print "Writing n_goc_score to the database\n" if $self->debug;
-    $self->write_n_tag($mlss, 'goc', $goc_scores);
+    foreach my $score ( keys %$goc_scores ) {
+        my $floor_score = int($score / 25) * 25;
```
muffato, line 181: GOC scores are already multiples of 25

PR #214, opened 2020-07-07
Flat file dumps of the gene_member table

muffato:
## Description

Now that the homologies are stored in files, the next bottleneck faced by the GOC analysis is fetching gene_members from the database. 

## Overview of changes

With these commits, I make the pipeline dump the `gene_member` table into flat files (TSV) and the Runnables use them.

#### Change #1
- Added the dumping analysis. It runs in parallel of preparing the per-mlss_id homology files

#### Change #2
- Use the files in the GOC analysis

#### Change #3
- Use the files in the `get_long_short_orth_genes` analysis

#### Change #4
- Use the files in the WGA pipeline `calculate_wga_coverage`

## Testing

I have run the GOC analysis with an increased capacity (100 instead of 30). The analysis was overall 15x faster, and I haven't tried to optimise the capacity (100 is the only value I have tried)

## Notes

- I am not "hashing" the dumps per genome_db_id like we do for the homologies and the "genome dumps" to keep the change minimal in this PR. There are not so many files (_O(n)_) and there are not very big, so I don't think it will be a problem. In the long term I think they do, but I would do the change on `master`
- May clash with #213 . The only change I expect is to get the correct list of genome_db_ids to load gene_members of.


JAlvarezJarreta:
Since I'm a little bit busy with production it will take longer for me to finish, I'm happy to merge this PR first and then rebase mine.

modules/Bio/EnsEMBL/Compara/PipeConfig/ProteinTrees_conf.pm
```diff
@@ -3422,6 +3424,23 @@ sub core_pipeline_analyses {
...
+            -flow_into => {
+                2 => [ 'gene_dumps' ],
+            },
+        },
+
+        {   -logic_name     => 'gene_dumps',
+            -module         => 'Bio::EnsEMBL::Hive::RunnableDB::DbCmd',
+            -parameters     => {
+                'output_file'   => '#gene_dumps_dir#/gene_member.#genome_db_id#.tsv',
+                'append'        => ['-B', '-q'],
```
JAlvarezJarreta, line 3439:
```suggestion
                'append'        => ['--batch', '--quick'],
```
I find it more readable for people not that familiar with the `mysql` client.

modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/GeneOrderConservation.pm
```diff
@@ -82,16 +85,21 @@ sub fetch_input {
...
+        my @head_cols = split(/\s+/, $header);
+        while ( my $line = <$gene_fh> ) {
+            my $row = map_row_to_header( $line, \@head_cols );
+            if ($all_gene_member_ids->{$row->{'gene_member_id'}}) {
+                push @gene_members_ordered, [$row->{'gene_member_id'}, $row->{'dnafrag_id'}, $row->{'dnafrag_start'}, $row->{'dnafrag_strand'}];
+            }
+        }
+        close($gene_fh);
     }
-    my @gene_members_ordered = sort { $a->[1] <=> $b->[1] || $a->[2] <=> $b->[2] } @gene_members_unordered;
```
JAlvarezJarreta, line 94: I'm not sure if it matters, but the the previous `@gene_members_ordered` was sorted by `genome_db_id` and this new one is by `dnafrag_id` and `dnafrag_start` (based on the query that generates the files), so they may not be the same unless you are certain the files are read in order sorted by `genome_db_id`.

muffato: As far as I understand the previous `@gene_members_ordered` was sorted by `dnafrag_id` and then `dnafrag_start`, The new one is still sorted on these two fields, but within each genome_db: the `dnafrag_ids` are not globally ordered. It doesn't matter because the index that is created just below (`neighbourhood`) only needs the genes of each `dnafrag_id` to be ordered by `dnafrag_start` with no assumption how the `dnafrag_ids` themselves are ordered. They could even be interleaved, it would still work

modules/Bio/EnsEMBL/Compara/PipeConfig/ProteinTrees_conf.pm
```diff
@@ -85,6 +85,7 @@ sub default_options {
         'fasta_dir'             => $self->o('work_dir') . '/blast_db',  # affects 'dump_subset_create_blastdb' and 'blastp'
         'cluster_dir'           => $self->o('work_dir') . '/cluster',
         'dump_dir'              => $self->o('work_dir') . '/dumps',
+        'gene_dumps_dir'        => $self->o('dump_dir') . '/genes',
```
carlacummins, line 88: For all other types of file dump, I used the `dump_dir` as a base. It won't impact functionality, but I guess we should try to be consistent. 
We can either change `homology_dumps_dir` and `orthotree_dir` to use `work_dir` as a base or use `dump_dir` as a base here.

modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/GeneOrderConservation.pm
carlacummins: Why on earth did I ever sort them using perl instead of SQL?!! 🤯😅

muffato:
Of course it would benefit too ! I did a search but stupidly missed it

modules/Bio/EnsEMBL/Compara/PipeConfig/ProteinTrees_conf.pm
muffato: Sorry, I don't understand. Yes `homology_dumps_dir` and `orthotree_dir` use `dump_dir`:
```perl
'homology_dumps_dir'        => $self->o('dump_dir') . '/homology_dumps/',
'orthotree_dir'             => $self->o('dump_dir') . '/orthotree/',
```
and so does `gene_dumps_dir` :thinking: ?

muffato:
> since the `GeneOrderConservation.t` test is now using `gene_members` from file, could we empty this file: `modules/t/test-genome-DBs/orth_qm_goc/compara/gene_member.txt`

I think they are still needed to honour the foreign keys from the `homology_member` table, because we still test the "write to the database" mode of the Runnable. If we only test the "write to a file" mode, I guess we could empty a few more tables: `homology`, `homology_member`, `gene_member_hom_stats`, `sequence`, `seq_member`, `gene_member` and `dnafrag` ( `gene_member_hom_stats` and `sequence` could probably be dropped right now, actually).

muffato:
> And second, I think WGA could really benefit from this too. Perhaps it's outside of the scope of this PR, but a rewrite of this method could definitely be useful: https://github.com/Ensembl/ensembl-compara/blob/release/102/modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/CalculateWGACoverage.pm#L380-L398

I changed `RunnableDB/GeneSetQC/FindGeneFragments.pm`, so no reason not to change `RunnableDB/OrthologQM/CalculateWGACoverage.pm` too ;)

I think I've found out why I didn't turn this Runnable. I thought that each job was processing a batch, e.g. 500 homologies, and that it could be suboptimal to load the entire genomes in each batch. After looking into the pipeline, I realise this is not true: there is 1 job per mlss_id, that processes the entire genome (except the bit that could be reused, but I'm still happy with this).

I expect the impact to be modest, though, because the Runnable still needs to load the exon boundaries from the database (which includes a `JOIN` to the `gene_member` table), which could be dumped, but that I do consider to be outside of this PR :)
I guess the next step in the evolution of our pipelines would be to make the LoadMembers pipeline dump all its data to flat files ?

The change in `CalculateWGACoverage` is almost the same as in `FindGeneFragments`, but since we run WGA on ncRNAs too, I need to add the gene dumps in the ncRNA-tree pipeline as well, which makes this PR even broader.
Also, this PR was set for `release/102` but:
1. All the e102 pipelines have now started. Merging it now would mean running the dumps manually, and adding all the missing parameters in the relevant analyses.
2. The changes conflict with #213.

I think it'd be better to consider this PR for master (e103+) instead, and I'll rebase it / fix the conflicts once #213 is merged ?

I would also propose a slight refactoring of the WGA pipeline so that it doesn't pass so much data between jobs (at the moment, eHive has to store / retrieve the entire set of (considered) orthologies as job parameters). We could merge the `prepare_orthologs`, `calculate_wga_coverage`, `assign_wga_coverage_score` and `reuse_wga_score`, and gain a little bit of time.

modules/Bio/EnsEMBL/Compara/PipeConfig/ProteinTrees_conf.pm
carlacummins: Sorry, must have misread this. Really thought it was pointing at `$self->o('work_dir') . '/genes'` 🤦‍♀️

carlacummins (approved):

modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/CalculateWGACoverage.pm
```diff
@@ -380,18 +382,17 @@ Store it in a param 'member_info'
...
-        $gm_sth->execute($gm_id_2);
-        $member_info->{"gene_member_$gm_id_2"} = $gm_sth->fetchrow_hashref;
+    foreach my $gdb_id ($self->param_required('species1_id'), $self->param_required('species2_id')) {
+        my $gene_file = $self->param_required('gene_dumps_dir') . "/gene_member.${gdb_id}.tsv";
+        open( my $gene_fh, '<', $gene_file) or die "Cannot open $gene_file";
+        my $header = <$gene_fh>;
+        my @head_cols = split(/\s+/, $header);
+        while ( my $line = <$gene_fh> ) {
+            my $row = map_row_to_header( $line, \@head_cols );
+            $member_info->{'gene_member_'.$row->{'gene_member_id'}} = $row;
```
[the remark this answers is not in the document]
muffato: :+1: Especially as we may add more rows to the .tsv

PR #218, opened 2020-07-22
Split travis perl unit tests into internal and external

carlacummins:
## Description

Our TravisCI test suite is split into multiple parts (e.g. python, housekeeping, perl). Our Perl test suite has several tests that have external dependencies (e.g. ensembl-rest). When there is a problem in these external repos, their failures can mask any issues that we may introduce with commits or PRs.
Split the Perl test suite further into those that have external dependencies and those that are internal only in order to separate the types of failures we see.

**Related JIRA tickets:**
- ENSCOMPARASW-3544

## Overview of changes

- remove `ensembl` and `ensembl-rest` tests from `perl-unittest_harness.sh`
- create new `perl-external_unittest_harness.sh` to run those `ensembl` and `ensembl-rest` tests
- add the new harness to `.travis.yml` to trigger its run

## Testing
Will test using this PR


muffato:
I expect the code coverage to drop significantly because a large part of our coverage comes from the `ensembl-rest` tests

carlacummins:
I don't know why github isn't reporting the newest travis run, but it's here and it's passed: https://travis-ci.org/github/Ensembl/ensembl-compara/builds/710777948

Let me know if you'd like me to turn coverage on for the external tests (though I don't see a need to do so, personally)

muffato:
> Let me know if you'd like me to turn coverage on for the external tests (though I don't see a need to do so, personally)

I would expect consistency: either we want to know about the coverage of the test suite and we enable it on all runs, or we don't and we disable it everywhere. I find it awkward to partially enable it.

carlacummins:
Ok, since it's an all or nothing situation, it's best to go with all. Will reenable it now. Thanks 🙂

carlacummins:
> I have a feeling the `after_success` commands for Perl are overwriting each other

Thanks - yep, that was my feeling too. Was waiting to see what coverage would be reported. Think it's pretty clear now that it is indeed the case! 🙂  Will add the flag now.

muffato:
Coveralls is as always confusing, but it does see the tree jobs: "Jobs" section on https://coveralls.io/builds/32273402 . It should be able to combine them but that's not clear, cf "coverage remained the same at 36.902%" on the previous link and "coverage decreased (-4.9%) to 60.305%" on https://coveralls.io/builds/32273559

Codecov is clearer: https://codecov.io/gh/Ensembl/ensembl-compara/pull/218/commits . You can see that the coverage decreased and then increased back to what it was

carlacummins:
That's good because I hadn't figured out how to merge the coverage reports from 2 different harnesses 😅

PR #219, opened 2020-07-23
Runnable to load MLSS id(s)

JAlvarezJarreta:
## Description

In some of our pipelines we need to pass the MLSS id manually, which requires to go back to our master database for that division to get the number, making the whole process tedious and error-prone. With this new runnable all that work is not necessary any more, making production a little bit "safer".

I have tried to be as generic as possible, giving the option to flow the MLSS id(s) to the given branch code or to store them as pipeline-wide parameters.

#### Change 1
- New `LoadMLSSids.pm` runnable that will take to fetch and load the MLSS ids into the pipeline programatically, avoiding human errors.

#### Change 2
- Applied this runnable to the Protein trees pipelines as a "proof of concept".

## Testing
The unit tests are passing, and so are the tests I have done manually with the current vertebrates master whilst I was developing it.

## Notes
- Once this is approved, I will create a ticket to add this runnable to every pipeline that asks for the MLSS id in their `init_pipeline.pl` command (following the "proof of concept" proposed). For the EPO pipelines this is going to be a little bit trickier as the pipeline-wide parameters are not the same as in other pipelines, so they will need to be renamed everywhere.


modules/Bio/EnsEMBL/Compara/PipeConfig/ProteinTrees_conf.pm
```diff
@@ -752,7 +760,7 @@ sub core_pipeline_analyses {
             -module     => 'Bio::EnsEMBL::Compara::RunnableDB::GenomeDBFactory',
             -parameters => {
                 'compara_db'        => '#master_db#',   # that's where genome_db_ids come from
-                'mlss_id'           => $self->o('mlss_id'),
+                'mlss_id'           => '#mlss_id#',
```
muffato, line 763: Need to remove this line, otherwise the job will die
```
ParamError: substitution loop among {'mlss_id'} has been detected
```
```suggestion
```

modules/Bio/EnsEMBL/Compara/PipeConfig/ProteinTrees_conf.pm
```diff
@@ -628,7 +626,7 @@ sub core_pipeline_analyses {
             -module     => 'Bio::EnsEMBL::Compara::RunnableDB::AssertMatchingVersions',
             -input_ids  => [ { } ],
             -flow_into  => {
-                '1->A'  => [ 'copy_ncbi_tables_factory' ],
+                '1->A'  => [ 'load_mlss_id' ],
```
muffato, line 629: At the moment, the pipeline doesn't require a master database and has a switch after `check_member_db_is_same_version`. `LoadMLSSids` does require a master database, so by inserting it so early you're removing the ability to run the pipeline without a master database. You should insert it further down instead, for instance just before `load_genomedb_factory`.

modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMLSSids.pm
```diff
@@ -0,0 +1,178 @@
...
+    my $mlss_adaptor = $master_dba->get_MethodLinkSpeciesSetAdaptor;
+    # First, fetch all the MLSSs that match the given method type and species set name
+    my @mlsss = grep { $_->species_set->name eq $species_set_name }
+        @{ $mlss_adaptor->fetch_all_by_method_link_type($method_type) };
+    $self->throw("No MLSSs found for method '$method_type' and species set '$species_set_name'") unless @mlsss;
+
+    # Fetch the "current" MLSS id (for the given release number)
+    my @curr_mlss = grep { $_->is_in_release($release) } @mlsss;
+    $self->throw("No MLSS found for method '$method_type' and species set '$species_set_name' in release $release") unless @curr_mlss;
+    my $mlss = $curr_mlss[0];
```
muffato, line 127: What if `@curr_mlss` has more than 1 element ?

modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMLSSids.pm
```diff
@@ -0,0 +1,178 @@
...
+    $mlss_ids{mlss_id} = $mlss->dbID;
+
+    if ( $self->param('add_prev_mlss') ) {
+        # Fetch the previous MLSS id
+        my $prev_release = $mlss->first_release - 1;
+        my @prev_mlss = grep { $_->last_release == $prev_release } @mlsss;
+        $self->throw(sprintf("No previous MLSS found for MLSS '%s' (%s)", $mlss->name, $mlss->dbID)) unless @prev_mlss;
+        $mlss_ids{prev_mlss_id} = $prev_mlss[0]->dbID;
+    }
```
muffato, line 136: Do you have an example where this would be used ? Otherwise I can't tell whether this is the behaviour we need

modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMLSSids.pm
JAlvarezJarreta: I thought for a given method type and species set name we can only have 1 MLSS in the current release? I expected that something before would have complained about a malformed master database if not.

modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMLSSids.pm
JAlvarezJarreta: I need it for the new update MSA pipeline, as I need to load the data from the previous MLSS to update it. Not sure if other pipelines will use it, but it is a feature I definitely need for mine.

modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMLSSids.pm
muffato: There is no constraint in the code / schema to enforce that, it is just a convention. But you're right, there should be one at most in any given release.
I thought we had a HC/DC to check that in, but I can't find it. @CristiGuijarro : do you remember seeing one ? If not, we should create a Jira ticket to add this check

CristiGuijarro: There isn't a DC for that yet, but an appropriate place to put it might be in `CheckMSANames.pm`.
When I was working on the rice `species_sets` for the rice MSAs I was able to name every `species_set` "rice", and use those different `species_sets` with the **same** names for multiple attempts of the same `EPO` method. So this wasn't in a release db - but this sort of scenario didn't occur to me at the time as a new recruit.

muffato: We need this in release dbs as that's the assumption behind `fetch_by_method_link_type_species_set_name`. That's also advertised on the page of each multiple alignment, e.g. http://www.ensembl.org/info/genome/compara/mlss.html?mlss=183

In master dbs, it's probably generally the case for every release number ever, but there could be a couple of failures here and there. Also, like you said it's useful to break the rule when doing some tests. I also tested EPO with different `fish` species-sets end of last year. But I used my own copy of the master db for the duration of the tests (so did you, if I remember correctly), and I was not running HCs/DCs on it. So, for the master db, I guess it's up to us to decide if and how we want to enforce it :wink: but the same check as in the release db should work ?

modules/t/test-genome-DBs/test_master/compara/genome_db.txt
```diff
@@ -1,5 +1,8 @@
+1	9606	homo_sapiens	GRCh36	2009-07-Ensembl	0	0	\N	\N	Human	\N	79	\N
```
carlacummins, line 1: This test db is used by another test, which tests whether it can add/update/release genome dbs, including the ones you've added here. These changes may disrupt the other test (`MasterDatabase.t`).

modules/t/LoadMLSSids.t
```diff
@@ -0,0 +1,116 @@
...
+my $dba = $multi_db->get_DBAdaptor('compara');
+my $dbc = Bio::EnsEMBL::Hive::DBSQL::DBConnection->new(-dbconn => $dba->dbc);
+my $compara_db = $dbc->url;
+
+# Test that the MLSS id is returned correctly
+standaloneJob(
+	'Bio::EnsEMBL::Compara::RunnableDB::LoadMLSSids',  # module
+	{ # input param hash
+        'master_db'        => $compara_db,
+	    'method_type'      => 'PROTEIN_TREES',
```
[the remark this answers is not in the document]
JAlvarezJarreta: All the indentation issues come from a copy-paste I did of a previous file that was mixing tabs and spaces :man_facepalming:

modules/t/test-genome-DBs/test_master/compara/genome_db.txt
JAlvarezJarreta: I did check that the other tests using this database passed with this change, and Travis doesn't seem to complain either :smile:

modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMLSSids.pm
```diff
@@ -0,0 +1,191 @@
...
+    my $self = shift;
+    my $master_dba = $self->get_cached_compara_dba('master_db');
+    my $release = $self->param_required('release');
+    my $method_type = $self->param_required('method_type');
+    my $species_set_name = $self->param_required('species_set_name');
+
+    my %mlss_ids;
+    my $mlss_adaptor = $master_dba->get_MethodLinkSpeciesSetAdaptor;
+    # First, fetch all the MLSSs that match the given method type and species set name
+    my @mlsss = grep { $_->species_set->name eq $species_set_name }
```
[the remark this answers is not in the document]
JAlvarezJarreta: `fetch_by_method_link_type_species_set_name` returns **one** MLSS for that method link type and species set name, but a random one from all the MLSSs that match that criteria (including retired ones). So if we want the current and previous MLSSs, that's the best way I could think of.

modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMLSSids.pm
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        if ( $method_type =~ /^EPO_EXTENDED$/ ) {
+            # Fetch the EPO MLSS id
+            my @core_mlss = grep { ($_->species_set->name eq $species_set_name) && $_->is_in_release($release) }
+                @{ $mlss_adaptor->fetch_all_by_method_link_type('EPO') };
+            $self->throw(sprintf("No EPO MLSS found for MLSS '%s' (%s)", $mlss->name, $mlss->dbID)) unless @core_mlss;
+            $mlss_ids{ext_mlss_id} = $mlss_ids{mlss_id};
+            $mlss_ids{mlss_id} = $core_mlss[0]->dbID;
+        }
```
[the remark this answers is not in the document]
JAlvarezJarreta: The first three lines could, but the other two are different in each case, so I thought the code was more readable left this way. Will try a better, more compact approach.

modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMLSSids.pm
```diff
@@ -0,0 +1,186 @@
...
+            $mlss_ids{mlss_id} = $linked_mlss[0]->dbID unless $is_epo;
+            # The GERP MLSSs are linked to the EPO Extended MLSS, not the EPO MLSS
+            $mlss = $linked_mlss[0] if $is_epo;
+        }
+
+        if ( $method_type =~ /^(EPO|EPO_EXTENDED|PECAN)$/i ) {
+            # Fetch the GERP MLSS ids (constrained element and conservation score)
+            my $ce_mlss = $mlss->get_all_sister_mlss_by_class('ConstrainedElement.constrained_element');
+            $self->warning(sprintf("No Constrained Element MLSS found for MLSS '%s' (%s)\n", $mlss->name, $mlss->dbID)) unless @$ce_mlss;
+            $mlss_ids{ce_mlss_id} = $ce_mlss->[0]->dbID if @$ce_mlss;
```
muffato, line 161:
```suggestion
            if (@$ce_mlss) {
                $mlss_ids{ce_mlss_id} = $ce_mlss->[0]->dbID;
            } else {
                $self->warning(sprintf("No Constrained Element MLSS found for MLSS '%s' (%s)\n", $mlss->name, $mlss->dbID));
            }
```

modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMLSSids.pm
```diff
@@ -0,0 +1,186 @@
...
+            my $cs_mlss = $mlss->get_all_sister_mlss_by_class('ConservationScore.conservation_score');
+            $self->warning(sprintf("No Conservation Score MLSS found for MLSS '%s' (%s)\n", $mlss->name, $mlss->dbID)) unless @$cs_mlss;
+            $mlss_ids{cs_mlss_id} = $cs_mlss->[0]->dbID if @$cs_mlss;
```
muffato, line 164:
```suggestion
            if (@$cs_mlss) {
                $mlss_ids{cs_mlss_id} = $cs_mlss->[0]->dbID;
            } else {
                $self->warning(sprintf("No Conservation Score MLSS found for MLSS '%s' (%s)\n", $mlss->name, $mlss->dbID));
            }
```

modules/Bio/EnsEMBL/Compara/RunnableDB/LoadMLSSids.pm
```diff
@@ -0,0 +1,186 @@
...
+            # Fetch the linked MLSS id
+            my $linked_method_type = $is_epo ? 'EPO_EXTENDED' : 'EPO';
+            my @linked_mlss = grep { ($_->species_set->name eq $species_set_name) && $_->is_in_release($release) }
+                @{ $mlss_adaptor->fetch_all_by_method_link_type($linked_method_type) };
+            $self->throw(sprintf("No %s MLSS found for MLSS '%s' (%s)", $linked_method_type, $mlss->name, $mlss->dbID)) unless @linked_mlss;
+            # Assign the EPO MLSS id to mlss_id and EPO Extended MLSS id to ext_mlss_id
+            $mlss_ids{ext_mlss_id} = $is_epo ? $linked_mlss[0]->dbID : $mlss_ids{mlss_id};
+            $mlss_ids{mlss_id} = $linked_mlss[0]->dbID unless $is_epo;
+            # The GERP MLSSs are linked to the EPO Extended MLSS, not the EPO MLSS
+            $mlss = $linked_mlss[0] if $is_epo;
```
muffato, line 154:
```suggestion
            if ($is_epo) {
                $mlss_ids{ext_mlss_id} = $linked_mlss[0]->dbID;
                # The GERP MLSSs are linked to the EPO Extended MLSS, not the EPO MLSS
                $mlss = $linked_mlss[0];
            } else {
                $mlss_ids{ext_mlss_id} = $mlss_ids{mlss_id};
                $mlss_ids{mlss_id}     = $linked_mlss[0]->dbID;
            }
```

muffato (approved):
Some minor suggestions to combine consecutive conditional statements that use the same condition

PR #223, opened 2020-08-18
metazoan for 102

vsitnik:
## Description

Added conf for the Synteny pipeline for metazoa division.
Added ENV.pm template used for 102(49) built

## Overview of changes
same as above

## Testing
Db was built for 102 (succefully handed over)


muffato:
What's the difference between your `ENV.pm` and the default one ? The dumps dirs, as far as I can tell. What else ?

vsitnik:
@muffato :
```
...
-        'shared_user'           => 'compara_ensembl',
+        #'shared_user'           => 'compara_ensembl',
+        'shared_user'           => '',
...
-        'shared_hps_dir'        => '/hps/nobackup2/production/ensembl/' . $self->o('shared_user'),
+        #'shared_hps_dir'        => '/hps/nobackup2/production/ensembl/' . $self->o('shared_user'),
+        'shared_hps_dir'        => '/hps/nobackup2/production/ensembl/compara_ensembl',
...
-        'genome_dumps_dir'      => $self->o('shared_hps_dir') . '/genome_dumps/'.$self->o('division').'/',
+        #'genome_dumps_dir'      => $self->o('shared_hps_dir') . '/genome_dumps/'.$self->o('division').'/',
+        'genome_dumps_dir'      => '/gpfs/nobackup/ensembl/vsitnik/compara.102/dumps',
...
-        'homology_dumps_shared_basedir' => $self->o('shared_hps_dir') . '/homology_dumps/'. $self->o('division'),
+        #'homology_dumps_shared_basedir' => $self->o('shared_hps_dir') . '/homology_dumps/'. $self->o('division'),
+        'homology_dumps_shared_basedir' => '/gpfs/nobackup/ensembl/vsitnik/compara.102/homology_dumps',
...
```

muffato:
Did you have to change `warehouse_dir` too ? Where did your _PrepareMasterDatabaseForRelease_ pipeline copy the backup of the master database ?

vsitnik:
the problem with `warehouse_dir` is that it points to HMMs and stuff. Thus, I'm patching it before running PrepareMaster and revert it afterwards.

vsitnik:
another option, delete ENV 'template' from the PR. it's in the conf dir, anyway

muffato:
I'm thinking of different ways of doing it. Having a separate ENV file is the best way to get the configs out of sync. For `genome_dumps_dir` and `homology_dumps_shared_basedir' I would recommend having a symlink to your own directory from our shared location, so that the default config works.

Then, the `shared_user` parameter is only used in the PrepareMasterDatabaseForRelease pipeline to copy those .sql dumps. I would suggest to make the shared user an option (the Runnable would only add `become` if needed) that is set to _undef_ in the Metazoa version of the pipeline

vsitnik:
I'll remove the conf/metazoa/ENV.pm.102_49_template then, if you don't mind.
I hope, we'll address this issue next time.

vsitnik:
ENV template deleted

modules/Bio/EnsEMBL/Compara/PipeConfig/Metazoa/Synteny_conf.pm
```diff
@@ -0,0 +1,63 @@
...
+
+use base ('Bio::EnsEMBL::Compara::PipeConfig::Synteny_conf');  # All Hive databases configuration files should inherit from HiveGeneric, directly or indirectly
+
+sub default_options {
+    my ($self) = @_;
+    return {
+        %{$self->SUPER::default_options},   # inherit the generic ones
+        'division'     => 'metazoa',
+        
+        'alignment_db'    => 'compara_curr',
```
CristiGuijarro, line 58: These db parameters are present as default in the `Bio::EnsEMBL::Compara::PipeConfig::Synteny_conf` for `compara_curr`. The only non-default value provided here is `'division'     => 'metazoa'` which could be provided on the command line:
`init_pipeline.pl Bio::EnsEMBL::Compara::PipeConfig::Synteny_conf -host mysql-ens-compara-prod-X -port XXXX         -division metazoa`
I am not sure that this PR is a very necessary addition :slightly_smiling_face:

vsitnik: I assume, we can drop  PipeConfig/Metazoa/Lastz_conf.pm,  as well, as most of the PipeConfig/Plants, for homogenity.

CristiGuijarro: As far as I can tell `Bio::EnsEMBL::Compara::PipeConfig::Metazoa::Lastz_conf` and many others of the division specific configs contain more than the default parameters in `Bio::EnsEMBL::Compara::PipeConfig::*_conf`, and analysis tweaks that are very much needed. We use environmental variables with the pipelines for division `$COMPARA_DIV` in the `init_pipeline` command line, so for our side, we wouldn't need a conf just for the division parameter. I wouldn't say we have to drop this division specific config, but at the moment, the two db parameters at least are not needed in it. You could keep this config, with the future implications that other parameters may change or analysis tweaks are likely to be added?

vsitnik: I see. Thanks a lot, Cristi. Actually, I prefer specifying init options. And, yes, config is not that deverged from the base one. So, if we don't need this config as a marker that Synteny should be run for metazoan compara, I'll close this PR.

vsitnik: discussed with @CristiGuijarro . devision specific PipeConfig dirs are not used as any kind of a marker. thus closing this PR. thanks.

vsitnik:
closing, see resolved comment.

PR #224, opened 2020-08-20
JIRA Authentication

carlacummins:
## Description

Interactions with JIRA using our `Utils::JIRA` adaptor usually requires the user to interactively input their password at runtime. This is not practical when embedding JIRA-related tasks into pipelines. Neither does it feel like a good idea to store passwords in plaintext in environment variables. This PR edits the adaptor to allow us to use an encoded authentication string, providing a slightly more safe method.

https://developer.atlassian.com/server/jira/platform/basic-authentication/

**Related JIRA tickets:**
- https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-3781

#### Change #1
- Updates to `Utils::JIRA` to read authentication token from `$ENV{'JIRA_AUTH_TOKEN'}`. The password is only requested if this is not present.
- `_http_request` updated to include token in header where given, or fallback to `basic_authentication` if not

#### Change #2
- updated `CreateDCJiraTickets.pm` to check for `$JIRA_AUTH_TOKEN`, rather than a plaintext password

## Testing
Test run of the `create_datacheck_tickets.pl` script

## Notes
Updated [Production Environment confluence document](https://www.ebi.ac.uk/seqdb/confluence/display/EnsCom/Production+Environment) to include information about setting the `$JIRA_AUTH_TOKEN` variable.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/Utils/JIRA.pm
+++ b/modules/Bio/EnsEMBL/Compara/Utils/JIRA.pm
@@ -77,8 +77,17 @@ sub new {
     $self->{_user} = $user || $ENV{'USER'};
     $self->{_relco} = $relco || $self->{_user};
     $self->{_project} = $project || 'ENSCOMPARASW';
-    # Initialise user's password
-    $self->{_password} = defined $ENV{'USERPW'} ? $ENV{'USERPW'} : $self->_request_password();
+
+    if ( $ENV{'JIRA_AUTH_TOKEN'} ) {
+        # this token should be your 'user:pass' string encoded to base64
+        # export JIRA_AUTH_TOKEN=$(echo -n 'user:pass' | openssl base64)
+        # https://developer.atlassian.com/server/jira/platform/basic-authentication/
+        $self->{_auth_token} = $ENV{'JIRA_AUTH_TOKEN'};
+    } else {
+        # initialise user's password interactively
+        $self->{_password} = $self->_request_password();
+    }
+
     # If any of the following parameters are missing, get them from Compara
     # production environment
     # (https://www.ebi.ac.uk/seqdb/confluence/display/EnsCom/Production+Environment)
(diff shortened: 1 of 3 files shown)


modules/Bio/EnsEMBL/Compara/Utils/JIRA.pm
```diff
@@ -644,10 +653,17 @@ sub _http_request {
     $content_data //= {};
     # Create the HTTP request and LWP objects to get the response for the given arguments
     $self->{_logger}->debug("$method Request on $url\n");
-    my $request = HTTP::Request->new($method, $url);
-    $request->authorization_basic($self->{_user}, $self->{_password});
-    # The content data will be sent in JSON format
-    $request->header('Content-Type' => 'application/json');
+    my $request;
+    if ( $self->{_auth_token} ) {
```
JAlvarezJarreta, line 657:
```suggestion
    if ( exists $self->{_auth_token} ) {
```

modules/Bio/EnsEMBL/Compara/Utils/JIRA.pm
```diff
@@ -77,8 +77,17 @@ sub new {
     $self->{_user} = $user || $ENV{'USER'};
     $self->{_relco} = $relco || $self->{_user};
     $self->{_project} = $project || 'ENSCOMPARASW';
-    # Initialise user's password
-    $self->{_password} = defined $ENV{'USERPW'} ? $ENV{'USERPW'} : $self->_request_password();
+
+    if ( $ENV{'JIRA_AUTH_TOKEN'} ) {
```
JAlvarezJarreta, line 81:
```suggestion
    if ( exists $ENV{'JIRA_AUTH_TOKEN'} && defined $ENV{'JIRA_AUTH_TOKEN'} ) {
```
(or maybe you prefer to raise an error if the env variable is not defined)

CristiGuijarro (approved):

modules/Bio/EnsEMBL/Compara/Utils/JIRA.pm
carlacummins: I don't really understand what this is meant to catch.. Cases where `$self->{_auth_token}` is 0?

JAlvarezJarreta: Unless my understanding is off (which is possible), you not always have `$self->{_auth_token}` defined, so you need to check that `_auth_token` key exists in the hash before accessing it.

carlacummins: Nope, perl can handle that and will return `undef` in that case. Also, I'm sure I would have seen some failures in testing if this were the case. I tested both cases where I have the env set and don't have the env set and it worked fine.

modules/Bio/EnsEMBL/Compara/Utils/JIRA.pm
carlacummins: > (or maybe you prefer to raise an error if the env variable is not defined)

Yes, that's what this does 😉  Don't think we need this suggested change either.

PR #225, opened 2020-08-26
New SOP for out-of-the-release production work

JAlvarezJarreta:
## Description

New SOP to add genomes to the master database **before** the new production cycle starts to give us extra time to run all the required pipelines affected by the more disruptive species, mainly the fantastic 4.

**Related JIRA tickets:**
- ENSCOMPARASW-3648

#### Change 1
- Updated `PrepareMasterDBForRelease` pipeline and `UpdateGenomesFromRegFactory` runnable to handle the registry not only to create new master databases but also to update existing ones as well.

#### Change 2
- Created a new registry configuration file for this task, with the specific information of the species that need to be "pre-updated". I think it is better to keep it separated as its configuration is slightly different (and allows us a better track of the changes). So far this is a basic file, but may need to be extended as we will be using it to run all the production pipelines as well.

#### Change 3
- Added the new JSON file with the mouse-related JIRA tickets to be created and done before/during e103, taking advantage of this new SOP.

## Testing
I have created a copy of vertebrates master database from e101, run the first [Prepare pipeline](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-10&port=4648&dbname=jalvarez_prep_vertebrates_master_for_rel_102a) via the registry updating mouse, and then a [second time using the metadata](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-10&port=4648&dbname=jalvarez_prep_vertebrates_master_for_rel_102b) (adding to the report the mouse was updated as well, to be consistent with the previous changes). No conflicts were raised, just some MLSS and SS ids are "wasted" as they are generated but never actually used in between both runs. Both master databases (after the first run and final) are in `cp5` with names `jalvarez_compara_master_101a` and `jalvarez_compara_master_101`, respectively.

## Notes
As it happens with *Homo sapiens*, every species that has been previously updated via registry needs to have `force = 1` in their job so the analysis `add_species_into_master` doesn't fail.

### DIFF
--- a/conf/vertebrates/next_production_reg_conf.pl
+++ b/conf/vertebrates/next_production_reg_conf.pl
@@ -1,0 +1,103 @@
+#!/usr/bin/env perl
+# Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+# Copyright [2016-2020] EMBL-European Bioinformatics Institute
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Release Coordinator, please update this file before starting every release
+# and check the changes back into GIT for everyone's benefit.
+
+# Things that normally need updating are:
+#
+# 1. Release number
+# 2. Check the name prefix of all databases
+# 3. Possibly add entries for core databases that are still on genebuilders' servers
+
+use strict;
+use warnings;
+
+use Bio::EnsEMBL::Registry;
+use Bio::EnsEMBL::Compara::Utils::Registry;
+
+my $next_release = $ENV{'CURR_ENSEMBL_RELEASE'};
+my $curr_release = $next_release - 1;
+my $prev_release = $curr_release - 1;
+
+my $prev_db_suffix = Bio::EnsEMBL::Compara::Utils::Registry::PREVIOUS_DATABASE_SUFFIX;
+
+# ---------------------- SPECIES TO UPDATE OUT OF THE RELEASE ------------------
+
+my $updated_core_dbs = {
+    'mus_musculus' => [ 'mysql-ens-genebuild-prod-2', 'thibaut_mus_musculus_havana_39' ],
+};
+
+my $prev_core_dbs = {
+    "mus_musculus$prev_db_suffix" => [ 'mysql-ens-vertannot-staging', 'mus_musculus_core_102_38' ],
+};
+
+# ---------------------- CURRENT CORE DATABASES --------------------------------
+
+# All the core databases live on the Vertebrates staging server or our mirror
+# Bio::EnsEMBL::Registry->load_registry_from_url("mysql://ensro\@mysql-ens-sta-1:4519/$curr_release");
+Bio::EnsEMBL::Registry->load_registry_from_url("mysql://ensro\@mysql-ens-mirror-1:4240/$curr_release");
+# Remove species that will be updated $next_release
+Bio::EnsEMBL::Compara::Utils::Registry::remove_species( [ keys %$updated_core_dbs ] );
+# Add the correct core database for those species
+Bio::EnsEMBL::Compara::Utils::Registry::add_core_dbas( $updated_core_dbs );
+
+# ---------------------- PREVIOUS CORE DATABASES -------------------------------
+
+# Previous release core databases will be required by PrepareMasterDatabaseForRelease, LoadMembers and Mercator-Pecan
+*Bio::EnsEMBL::Compara::Utils::Registry::load_previous_core_databases = sub {
+    Bio::EnsEMBL::Registry->load_registry_from_db(
+        -host   => 'mysql-ens-mirror-1',
+        -port   => 4240,
+        -user   => 'ensro',
+        -pass   => '',
+        -db_version     => $prev_release,
+        -species_suffix => $prev_db_suffix,
+    );
+    # Do the same with the previous core database
+    Bio::EnsEMBL::Compara::Utils::Registry::remove_species( [ keys %$prev_core_dbs ] );
+    Bio::EnsEMBL::Compara::Utils::Registry::add_core_dbas( $prev_core_dbs );
+};
+
+# ---------------------- COMPARA DATABASE LOCATIONS ----------------------------
+
+# FORMAT: species/alias name => [ host, db_name ]
+my $compara_dbs = {
+    # general compara dbs
+    'compara_master' => [ 'mysql-ens-compara-prod-5', 'jalvarez_compara_master_101' ],
+    'compara_prev'   => [ 'mysql-ens-compara-prod-7', "jalvarez_compara_$curr_release" ],
+};
+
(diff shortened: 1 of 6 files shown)


modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromRegFactory.pm
```diff
@@ -15,39 +15,91 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 
 use warnings;
 use strict;
+
 use Bio::EnsEMBL::Registry;
+use Bio::EnsEMBL::Compara::Utils::CoreDBAdaptor;
 
 use base ('Bio::EnsEMBL::Hive::Process');
 
+
```
CristiGuijarro, line 45: I don't think we need this new newline
```suggestion
```

modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromRegFactory.pm
```diff
@@ -15,39 +15,91 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-    $self->param('species_to_update', $species_list);
+    my %core_dbas;
+    foreach my $species_name ( @$species_list ) {
+        my $dba = Bio::EnsEMBL::Registry->get_DBAdaptor($species_name, 'core');
+        $core_dbas{$species_name} = $dba;
+    }
+
+    # If provided, get the list of allowed species
+    my $allowed_species_file = $self->param('allowed_species_file');
+    if (defined $allowed_species_file && -e $allowed_species_file) {
```
CristiGuijarro, line 58:
```suggestion
    if (defined $allowed_species_file && -s $allowed_species_file) {
```
In case the file exists but is empty?

modules/Bio/EnsEMBL/Compara/PipeConfig/Parts/PrepareMasterDatabaseForRelease.pm
```diff
@@ -142,8 +142,13 @@ sub pipeline_analyses_prep_master_db_for_release {
 
         {   -logic_name => 'update_genome_from_registry_factory',
             -module     => 'Bio::EnsEMBL::Compara::RunnableDB::PrepareMaster::UpdateGenomesFromRegFactory',
+            -parameters => {
+                'master_db' => $self->o('master_db'),
```
CristiGuijarro, line 146:
```suggestion
                'compara_db' => $self->o('master_db'),
```
To go with runnable below

modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromRegFactory.pm
```diff
@@ -15,39 +15,91 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+        # Keep only the species included in the allowed list
+        my $allowed_species = { map { $_ => 1 } @{ decode_json($self->_slurp($allowed_species_file)) } };
+        my @excluded_species = grep { ! exists $allowed_species->{$_} } @{ keys %core_dbas };
+        delete @core_dbas{@excluded_species};
+    }
+
+    my $master_dba = Bio::EnsEMBL::Compara::DBSQL::DBAdaptor->go_figure_compara_dba( $self->param_required('master_db') );
```
CristiGuijarro, line 65:
```suggestion
    my $master_dba = $self->compara_dba();
```

modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromRegFactory.pm
```diff
@@ -15,39 +15,91 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-    $self->dataflow_output_id(\@new_genomes_dataflow, 2);
+    my @genomes_to_update = map { {species_name => $_, force => 0} } @{ $self->param('genomes_to_update') };
+    $self->dataflow_output_id(\@genomes_to_update, 2);
+    my @genomes_to_retire = map { {species_name => $_} } @{ $self->param('genomes_to_retire') };
+    $self->dataflow_output_id(\@genomes_to_retire, 3);
+    my @genomes_to_verify = map { {species_name => $_} } @{ $self->param('genomes_to_verify') };
+    $self->dataflow_output_id(\@genomes_to_verify, 5);
 }
 
+
```
CristiGuijarro, line 104:
```suggestion
```

CristiGuijarro (approved):
Comments are only suggestions and non essential :slightly_smiling_face:

conf/vertebrates/next_production_reg_conf.pl
```diff
@@ -0,0 +1,103 @@
...
+my $prev_db_suffix = Bio::EnsEMBL::Compara::Utils::Registry::PREVIOUS_DATABASE_SUFFIX;
+
+# ---------------------- SPECIES TO UPDATE OUT OF THE RELEASE ------------------
+
+my $updated_core_dbs = {
+    'mus_musculus' => [ 'mysql-ens-genebuild-prod-2', 'thibaut_mus_musculus_havana_39' ],
+};
+
+my $prev_core_dbs = {
+    "mus_musculus$prev_db_suffix" => [ 'mysql-ens-vertannot-staging', 'mus_musculus_core_102_38' ],
```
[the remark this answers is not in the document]
JAlvarezJarreta: The aliases are created that way by the registry (`load_registry_from_url()` and `load_registry_from_db()`), so that is the easiest way I could think of to define the new cores and be able to use both keys and hashes to replace them.
And yes, I think only the Fantastic 4 are meant to be involved in this out-of-the-release process.

modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromRegFactory.pm
JAlvarezJarreta: That file should never be empty, so if it exists, being empty should remove all species from the release (that's the expected behaviour). "Unfortunately", with the current implementation `decode_json()` will fail because it doesn't accept empty strings. I think either I fix it to work as I have said before, or to die with an appropriate message if the file exits and it is empty (I prefer this one). I'm not keen on just ignoring the file because it is empty. Opinions?

modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromRegFactory.pm
JAlvarezJarreta: It's a Python PEP8 thingy that has gotten stuck in my brain :sweat_smile:

modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromRegFactory.pm
JAlvarezJarreta: Well spotted!

conf/vertebrates/next_production_reg_conf.pl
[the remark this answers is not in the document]
JAlvarezJarreta: Those are the species that would have a higher impact in our production, yes, but I wasn't sure if all those should be run out of the release.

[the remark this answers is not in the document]
dthybert: if we can I think it is better to run them out of the release

PR #228, opened 2020-09-08
MIsc. WGA improvements

muffato:
## Description

Follow up of #196, made of several things I noticed while working on #229. I would have made the comments in #196 but I didn't get the chance to review it.

First of all, note that I have already pushed two bugfixes on `release/103`: ce71c75d918a5e206e07111816e3670da3c2dbcd and cf543b0ae5547682ba840dda388faeb027c29780

## Overview of changes

The commit messages are self-explanatory, but for clarity, here is a summary

#### Change n°1
- Don't need to define `compara_db` and `alt_homology_db` since by default the pipeline will default to the current database. Also having these defined means that the Runnable will open _another_ connection to the database, leaving the first one open and unused.

#### Change n°2
- Don't create the Compara tables in the standalone pipeline

#### Change n°3
- Moved parameters outside of the analysis, thus limiting what options pipelines that include WGA need to have

## Testing

I have run the (standalone) pipeline in #229


CristiGuijarro (approved):
:+1:

carlacummins (approved):

PR #229, opened 2020-09-08
WGA "All in one"

muffato:
## Description

The per-mlss_id work of WGA pipeline is split over four analyses, with a lot of data being transferred between them: homologies, coordinates, scores, etc. If you look at the runtime stats, `assign_wga_coverage_score.fetch_input` takes almost 40% of the total runtime, just fetching the scores. Some overhead take is not listed here, for instance, eHive fetching job parameters before the job starts.

```
+---------------------------+------------------------+----------------------+-------------------------+
| logic_name                | avg_input_msec_per_job | avg_run_msec_per_job | avg_output_msec_per_job |
+---------------------------+------------------------+----------------------+-------------------------+
| prepare_orthologs         |                 227613 |                  342 |                    2725 |
| calculate_wga_coverage    |                1770726 |               772603 |                   10888 |
| assign_wga_coverage_score |                1815250 |                    0 |                    2720 |
| reuse_wga_score           |                      0 |                    0 |                     514 |
+---------------------------+------------------------+----------------------+-------------------------+
```

My idea was to speed this up by combining those Runnables into 1.

## Overview of changes

I have created a new "Frankenstein" Runnable that calls all existing four Runnables. The advantage is that almost no changes are required in the existing Runnables, so the history will still be normally accessible.

I have benchmarked the new Runnable and the optimal capacity is 140, making it run in 7 hours in total (15 min per mlss_id) on e102 data. In the real database (`carlac_default_vertebrates_protein_trees_102`) it took 48 min per mlss_id the first time I looked at it (mid July) and now it seems to have taken 131 min per mlss_id, so maybe it was rerun since ? Anyway, the Frankenstein code is faster.

## Testing

No unit tests, but I have run the pipeline

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/CalculateWGACoverage.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/CalculateWGACoverage.pm
@@ -148,7 +148,7 @@ sub run {
 	my %aln_ranges = %{ $self->param('aln_ranges') };
     my %mlss_mapping = %{ $self->param('mlss_db_mapping') };
 
-	my (@qual_summary, @orth_ids);
+	my (@qual_summary, @orth_ids, %max_quality);
 
 	foreach my $orth ( @orth_info ) {
         my ($orth_dnafrags, $orth_ranges) = $self->_orth_dnafrags($orth);
@@ -157,6 +157,9 @@ sub run {
         my $exon_ranges    = $self->_get_exon_ranges_for_orth($orth);
 
 		push( @orth_ids, $homo_id );
+
+		# The WGA score defaults to 0
+		$max_quality{$homo_id} = 0;
 
 		next unless ( defined $this_aln_range ); 
 
@@ -167,6 +170,8 @@ sub run {
                 my $dba = Bio::EnsEMBL::Compara::DBSQL::DBAdaptor->go_figure_compara_dba($aln_db);
                 my $gdba = $dba->get_GenomeDBAdaptor;
 
+				my $sum_scores = 0;
+				my $n_scores   = 0;
 				foreach my $gdb_id ( sort {$a <=> $b} keys %{ $orth_ranges } ){
 					# Make sure that if this is a component gdb_id it refers back to the principal. The exon and ortholog data is on the components and not on the principal, but the aln_mlss_id is only on principal
 					my $gdb = $gdba->fetch_by_dbID($gdb_id);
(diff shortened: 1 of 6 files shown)


modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/CalculateWGACoverage.pm
```diff
@@ -184,16 +189,24 @@ sub run {
 						  intron_length            => $combined_coverage->{intron_len},
 						}
 					);
+					$sum_scores += $combined_coverage->{score};
+					$n_scores ++;
 				}
                 # disconnect from alignment_db
                 $dba->dbc->disconnect_if_idle();
+
+				my $avg_score = $sum_scores / $n_scores; # $n_scores should be 2
```
CristiGuijarro, line 198: If `$n_scores` should be 2, accumulated as it is on line 193, perhaps a handy warning to flag when it is not 2?

CristiGuijarro (approved):
I think it makes sense, it's more like Frankenstein's monster v2, but the changes are not too tricky to follow :+1:

modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/WGACoverage.pm
```diff
@@ -0,0 +1,108 @@
...
+    # IN: max_quality
+    # IN: member_type
+    # IN: mlss_db_mapping
+    # IN: aln_mlss_ids
+    # IN: output_file
+    # IN: reuse_file (for reading)
+    Bio::EnsEMBL::Compara::RunnableDB::OrthologQM::AssignQualityScore::write_output($self);
+
+    # If we want to record the scores in the ortholog_quality table
+    #$self->dataflow_output_id( $self->param('qual_summary'), 3 );
```
carlacummins, line 105: This table was being used for partial reuse (for example: only updating scores for one type of alignment, while keeping the other scores). I don't think is happening any more as it wasn't saving much time. It can prob be removed eventually.

carlacummins (approved):

modules/Bio/EnsEMBL/Compara/RunnableDB/OrthologQM/CalculateWGACoverage.pm
muffato: :+1: I have added a commit: https://github.com/Ensembl/ensembl-compara/commit/fdf464f43ed9202e4da9e155842fe9744fb0b8fc

PR #232, opened 2020-09-28
Expanded the tests to check that the indexes are the same too

muffato:
## Description

Follow-up of #227 / ENSCOMPARASW-3568. I have found that some test databases don't have the right set of indexes.

## Overview of changes

I have expanded `Utils::Test::get_schema_from_database` to also fetch the indexes, and updated the test databases so that they pass the check.

The trick is that the Compara test databases often have different indexes, depending on whether the real database they were created from has foreign keys or not, because MySQL automatically adds indexes in order to check the foreign key constraints.

I have made `travisci/sql-unittest/test_dbs_up_to_date.t` load the schema from two databases (with and without foreign keys) and consider all the extra indexes as optional.

## Testing

I have fixed all the test databases so that the test passes .

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/Utils/Test.pm
+++ b/modules/Bio/EnsEMBL/Compara/Utils/Test.pm
@@ -350,7 +350,7 @@ sub get_pipeline_tables_create_statements {
 =cut
 
 sub get_schema_from_database {
-    my $dbh = shift;
+    my ($dbh, $db_name) = @_;
     # I can't get column_info() to return all the columns of all the tables
     # of a given / the current database, so need to get the list of the
     # tables first, and then call column_info() on each of them.
@@ -362,6 +362,12 @@ sub get_schema_from_database {
         $schema{$t}->{'COLUMNS'} = $sth->fetchall_hashref('COLUMN_NAME');
         my @pk_columns = $dbh->primary_key_info(undef, undef, $t);
         $schema{$t}->{'PRIMARY_KEY'} = \@pk_columns;
+        $sth = $dbh->statistics_info(undef, $db_name, $t, undef, 'quick');
+        $schema{$t}->{'INDEXES'} = $sth->fetchall_hashref(['INDEX_NAME', 'ORDINAL_POSITION']);
+        foreach my $h (values %{$schema{$t}->{'INDEXES'}}) {
+            delete $_->{TABLE_SCHEM} for values %$h;
+            delete $_->{CARDINALITY} for values %$h;
+        }
     }
     return \%schema;
 }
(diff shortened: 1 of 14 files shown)


modules/Bio/EnsEMBL/Compara/Utils/Test.pm
```diff
@@ -362,6 +362,12 @@ sub get_schema_from_database {
         $schema{$t}->{'COLUMNS'} = $sth->fetchall_hashref('COLUMN_NAME');
         my @pk_columns = $dbh->primary_key_info(undef, undef, $t);
         $schema{$t}->{'PRIMARY_KEY'} = \@pk_columns;
+        $sth = $dbh->statistics_info(undef, $db_name, $t, undef, 'quick');
+        $schema{$t}->{'INDEXES'} = $sth->fetchall_hashref(['INDEX_NAME', 'ORDINAL_POSITION']);
+        foreach my $h (values %{$schema{$t}->{'INDEXES'}}) {
+            delete $_->{TABLE_SCHEM} for values %$h;
+            delete $_->{CARDINALITY} for values %$h;
+        }
```
JAlvarezJarreta, line 370: I think it would be better to understand this bit of code (its intention) with a short comment.

JAlvarezJarreta (approved):

modules/Bio/EnsEMBL/Compara/Utils/Test.pm
```diff
@@ -350,18 +350,34 @@ sub get_pipeline_tables_create_statements {
...
         $schema{$t}->{'PRIMARY_KEY'} = \@pk_columns;
+
+        # Fetch the indexes
+        # Note that unlike the two others, this call requires the database name
+        $sth = $dbh->statistics_info(undef, $db_name, $t, undef, 'quick');
+        $schema{$t}->{'INDEXES'} = $sth->fetchall_hashref(['INDEX_NAME', 'ORDINAL_POSITION']);
+        foreach my $h (values %{$schema{$t}->{'INDEXES'}}) {
+            # To make this identical across databases:
+            # 1. Remove the database name
+            delete $_->{TABLE_SCHEM} for values %$h;
```
CristiGuijarro, line 377: This is probably exactly as it is, but it looks like an unfinished word `TABLE_SCHEMA` :sweat_smile:

CristiGuijarro (approved):

modules/Bio/EnsEMBL/Compara/Utils/Test.pm
muffato: I know ... but it's correct ! cf https://metacpan.org/pod/DBI
> TABLE_SCHEM: The name of the schema containing the TABLE_NAME value. This field is NULL (undef) if not applicable to data source, and empty if not applicable to the table.

PR #233, opened 2020-09-28
More syntax checks

muffato:
## Description

This test is inspired by #230 (which I have used to test the new checks). When we update / get a pull request to update the newick tree, `mlss_conf.xml`,  `allowed_species.json`, etc, we always have to manually check whether they are correctly formatted, otherwise they may break in production,

## Overview of changes

Here I have added some more tests in the test-suite to check all those formats, plus a few more I could easily implement:
 - JSON
 - XML
 - Newick
 - Yaml
 - TSV
 - Shell scripts

#### Yaml

The syntax is checked with `yamllint`. I am using the default settings plus a few ones so that it doesn't complain unnecessarily. I have also added a few `# yamllint disable-line` pragmas in the `.yml` files to avoid making large changes in the files. If we need to modify those lines, we should consider bringing them up to the expected standard and remove  `# yamllint disable-line`.

#### Newick

I have fixed two bugs in our venerable NewickParser, which I have found by testing `newick.t` on custom files. NewickParser.pm should probably have its own unit tests, but I didn't feel like writing them now

#### Test scope

I have modified `Utils::Test::find_all_files` to check all directories (it was missing a few), meaning I had to add a few changes in `apache2.t` and `executability.t`

#### Shell scripts

We are now using `shellcheck` to check shell scripts. Beware it's quite picky ! But it found a bug in `perl-linter_harness.sh` :clap: 
It doesn't like this line because the syntax prevents from checking the return code of the `python` command
```
export PYTHONPATH=$PYTHONPATH:$(python -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])')
```
but do we care ? I have currently added a pragma to disable the test

## Testing

These are unit tests, and they all pass


modules/Bio/EnsEMBL/Compara/Graph/NewickParser.pm
```diff
@@ -222,6 +222,8 @@ sub parse_newick_into_tree
         throw("parse error: nothing expected after ;");
       }
   }
+  throw("no tree at all") unless defined $root;
+  throw("parse error: unbalanced ()\n") if($bracket_level ne 0);
```
JAlvarezJarreta, line 226: Let it breathe? ;)
```suggestion
  throw("parse error: unbalanced ()\n") if ($bracket_level ne 0);
```

travisci/all-housekeeping/executability.t
```diff
@@ -35,6 +35,10 @@ foreach my $f (@all_files) {
     # Exceptions
     $should_be_executable = '' if $f =~ /\/docs\/conf.py$/;
     $should_be_executable = '' if $f =~ /production.*reg_conf.*\.pl$/;
+    $should_be_executable = '' if $f =~ /docs\/conf\.py$/;
```
JAlvarezJarreta, line 38: Lines 36 and 38 are fairly similar. Do we need the distinction between them? I feel the latter one you have added should be enough to cover both scenarios.
```suggestion
    $should_be_executable = '' if $f =~ /docs\/conf\.py$/;
    $should_be_executable = '' if $f =~ /production.*reg_conf.*\.pl$/;
```

travisci/all-housekeeping/newick.t
```diff
@@ -0,0 +1,100 @@
...
+
+
+use File::Basename;
+use File::Spec;
+use JSON qw(decode_json);
+use Test::Exception;
+use Test::More;
+use XML::LibXML;
+
+use Bio::EnsEMBL::Utils::IO qw (slurp);
```
JAlvarezJarreta, line 27:
```suggestion
use Bio::EnsEMBL::Utils::IO qw(slurp);
```

JAlvarezJarreta (approved):

JAlvarezJarreta:
Quite useful set of tests :+1:

modules/Bio/EnsEMBL/Compara/Graph/NewickParser.pm
muffato: Sure :) !

travisci/all-housekeeping/executability.t
muffato: Ah, I didn't see it !

muffato: I'll fix the source commit

JAlvarezJarreta (approved):

CristiGuijarro (approved):
:+1:

PR #234, opened 2020-09-28
Overkill to be extra sure that trees to be deleted are deleted

CristiGuijarro:
## Description

Reindex of ncRNA trees for default vertebrates from release 100 to 101 was messy. I followed the problem sequences/trees through the reindex pipeline and all those that should flowed through correctly. The majority of the issues were actually caused by incorrect data in the release 100 trees being copied across. The only detectable issues caused by a bug in the reindex pipeline was the removal of the trees in which members had been lost between releases. These were dataflowed correctly, and picked up in the correct analyses later on, with `DONE` jobs apparently smooth. Somewhere along the line, I suspect that the transaction did not execute as smoothly as suggested and errors were missed. There does not seem to be edge case issues as assumed after all.

**Related JIRA tickets:**
- ENSCOMPARASW-3603

## Overview of changes
I have put in a check (overkill) following a retry with pause in the `delete_tree` transaction in `DeleteOneTree.pm` to ensure that the tree is removed. If the tree is not removed, the job will fail, preventing painful debugging later on.

## Testing
The runnables were checked in standalone mode with the known incorrect data to ensure the correct dataflows, and execution manually.

## Notes
I have no idea what caused the issues in the e100 ncRNA pipeline and why these were not picked up on in the original datachecking/healthchecking to pass on to the reindexed version. Perhaps worth running DCs on the previous pipelines before reindexing in the future.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/DeleteOneTree.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/DeleteOneTree.pm
@@ -47,8 +47,8 @@ sub write_output {
         $tree->preload;
         $gene_tree_adaptor->delete_tree($tree);
         $tree->release_tree;
-    }, 1, 2 );
-    $self->_check_clean_eradication(); # Make sure that the tree has been removed after waiting 2 seconds and retrying once - a messy reindex is a pointlss reindex
+    }, 1, 2 );  # Retry once and wait 2 seconds between retries
+    $self->_check_clean_eradication();  # Make sure that the tree has been removed - a messy reindex is a pointless reindex
 }
 
 sub _check_clean_eradication {
@@ -61,4 +61,3 @@ sub _check_clean_eradication {
 }
 
 1;
-
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/DeleteOneTree.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/DeleteOneTree.pm
@@ -47,7 +47,17 @@ sub write_output {
         $tree->preload;
         $gene_tree_adaptor->delete_tree($tree);
         $tree->release_tree;
-    } );
+    }, 1, 2 );
+    $self->_check_clean_eradication(); # Make sure that the tree has been removed after waiting 2 seconds and retrying once - a messy reindex is a pointlss reindex
+}
+
+sub _check_clean_eradication {
+    my $self = shift @_;
+
+    my $gene_tree_adaptor = $self->compara_dba->get_GeneTreeAdaptor;
+    if (my $tree = $gene_tree_adaptor->fetch_by_dbID($self->param_required('gene_tree_id'))) {
+        die 'Tree with root_id=' . $self->param('gene_tree_id') . ' removal has failed.';
+    }
 }
 
 1;


modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/DeleteOneTree.pm
```diff
@@ -47,7 +47,17 @@ sub write_output {
         $tree->preload;
         $gene_tree_adaptor->delete_tree($tree);
         $tree->release_tree;
-    } );
+    }, 1, 2 );
+    $self->_check_clean_eradication(); # Make sure that the tree has been removed after waiting 2 seconds and retrying once - a messy reindex is a pointlss reindex
```
JAlvarezJarreta, line 51:
```suggestion
    }, 1, 2 );  # Retry once and wait 2 seconds between retries
    $self->_check_clean_eradication();  # Make sure that the tree has been removed - a messy reindex is a pointless reindex
```

JAlvarezJarreta (approved):
Besides the typo, my reorganisation of the comment is just a suggestion :)

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/DeleteOneTree.pm
```diff
@@ -47,8 +47,17 @@ sub write_output {
         $tree->preload;
         $gene_tree_adaptor->delete_tree($tree);
         $tree->release_tree;
-    } );
+    }, 1, 2 );  # Retry once and wait 2 seconds between retries
```
muffato, line 50: If the `sub` fails, the transaction code will retry it if `$retry` is set, or throw an error otherwise. You're working on the premise that in e101, something went wrong but no error was raised. Based on that, I don't see how setting `$retry` to 1 will make it try again this time.

CristiGuijarro: I was basing it a little on something that happened to me when I was trying to copy/replace the trees from the previous release and delete the unstructured trees with `delete_tree()`. I wasn't sure what triggers the `$retry`. In my experiments, when the trees were very large `delete_tree()` struggled, and the processes timed out before they were reported as such, I wondered if it was a catch-up thing or something to do with rollback and perhaps pausing for a couple of seconds might help. What you say makes sense though. Remove this bit of possibly useless overkill?

muffato: Since we don't understand how the trees were not deleted, I shouldn't try to be logical :) Let's keep it and if we're lucky we won't face the issue again

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/DeleteOneTree.pm
```diff
@@ -47,8 +47,17 @@ sub write_output {
...
+    $self->_check_clean_eradication();  # Make sure that the tree has been removed - a messy reindex is a pointless reindex
 }
 
-1;
+sub _check_clean_eradication {
+    my $self = shift @_;
 
+    my $gene_tree_adaptor = $self->compara_dba->get_GeneTreeAdaptor;
+    if (my $tree = $gene_tree_adaptor->fetch_by_dbID($self->param_required('gene_tree_id'))) {
```
muffato, line 58: FYI this is fine for this adaptor, but other adaptors cache their data and further calls to `fetch_by_dbID` keep on returning the same object (unless the adaptor has a method to delete the object _and_ remove the entry from the cache). For instance, MLSS adaptor has a `delete` method that removes the entry from the table _and_ the cache, but not the GenomeDB adaptor, and the default `delete_by_dbID` method it inherits from BaseAdaptor is not cache-aware.

muffato (approved):

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/DeleteOneTree.pm
CristiGuijarro: I had not considered that at all, that's good to know, thanks!

PR #235, opened 2020-10-06
removed arachis_hypogaea & restored prunus_avium, ostreococcus_lucimarinus

brunocontrerasmoreira:
## Description

After handover of new plants during release 103/EG49 @james-monkeyshines realized that tetraploid arachis_hypogaea had some sequence discrepancies despite having passed critical core DCs. We decided to retire this species in the interest of time and restored two diploid species to keep the same number of species in Compara @CristiGuijarro @muffato @dthybert @gnaamati


CristiGuijarro (approved):
It doesn't make much if any difference, but the `species_tree.topology.nw` could be left as it was :+1:

weix-cshl:
Does that mean the following compara dump needs an update? 
http://ftp.ensemblgenomes.org/vol1/pub/.release-49/plants/mysql/ensembl_compara_plants_49_102/

Also shouldn't it be 50_103 if this is for the next build?

brunocontrerasmoreira:
Hi @weix-cshl , this affects release 1037EG50. Those compara dumps are unchanged.

PR #236, opened 2020-10-08
Feature/rm genome msas

JAlvarezJarreta:
## Description

With the new frozen release cycle we want to be able to remove non-reference species that are updated from EPO/Mercator-Pecan, and moved them to EPO Extended (in the former case), to avoid recalculating EPO/Mercator-Pecan MSAs again. This pipeline will take care of this automatically for a given method link and species set name.

**Related JIRA tickets:**
- ENSCOMPARASW-3065

#### Change 1
- This pipeline copies the MSA information from the previous release, updating the ancestral database if required, and updates the MLSS tags as well as the ancestral `dnafrag`s and all the ids that have the MLSS integrated in them. Afterwards, it will check the `GAB`s affected by the removed genome(s) and update them, updating also the `GA`s, `GAT`s (EPO only). The `GAB`s are treated in parallel and all the updates for each one are performed in a single transaction, to ensure consistency.

#### Change 2
- The EPO Extended pipeline has been integrated as @carlacummins said it wouldn't make a difference in terms of time cost to update the MSA adding the new genome(s) or to recompute it again from scratch, so I have decided to go for the latter option.

## Testing
I have run two "fake" production runs (from e101 to e102), one for [Mercator-Pecan](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-9&port=4647&dbname=jalvarez_amniotes_pecan_update_102) and another for [mammals EPO](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-9&port=4647&dbname=jalvarez_mammals_epo_update_102).

## Notes
- Since we are tight on time, I won't add unit tests for the new runnables in this PR. I will submit a new PR in the near future with those.
- The new dependency was already present in the farm, so that is why I used it.
- The latest commit (https://github.com/Ensembl/ensembl-compara/pull/236/commits/ac357e94670fc29ccc23fb255d9e89c4664c39ce) can be seen in [fish EPO](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-9&port=4647&dbname=jalvarez_fish_epo_update_102).

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMSA_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMSA_conf.pm
@@ -1,0 +1,439 @@
+=head1 LICENSE
+
+Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+Copyright [2016-2020] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=head1 NAME
+
+Bio::EnsEMBL::Compara::PipeConfig::UpdateMSA_conf
+
+=head1 SYNOPSYS
+
+    init_pipeline.pl Bio::EnsEMBL::Compara::PipeConfig::UpdateMSA_conf -host mysql-ens-compara-prod-X -port XXXX \
+        -division $COMPARA_DIV -method_type <lowercase_method_type> -species_set_name <species_set_name>
+
+=head1 DESCRIPTION
+
+The PipeConfig file for the pipeline that copies the data from the previous
+release MSA and updates it. This update involves:
+- Mercator-Pecan: remove the updated species and recompute GERP
+- EPO: copy the ancestral core db and update the dna names, remove the updated
+  species from the EPO MSA and update the ancestral dnafrag ids, and recompute
+  its EPO Extended MSA (GERP included), incorporating the updated species
+
+=cut
+
+package Bio::EnsEMBL::Compara::PipeConfig::UpdateMSA_conf;
+
+use strict;
+use warnings;
+
+use JSON;
+
+use Bio::EnsEMBL::Hive::PipeConfig::HiveGeneric_conf;  # for WHEN and INPUT_PLUS
+
+use Bio::EnsEMBL::Compara::PipeConfig::Parts::EPOAncestral;
+use Bio::EnsEMBL::Compara::PipeConfig::Parts::EpoLowCoverage;
+
+use base ('Bio::EnsEMBL::Compara::PipeConfig::ComparaGeneric_conf');
+
+
+sub default_options {
+    my ($self) = @_;
+    return {
+        %{$self->SUPER::default_options},  # inherit the generic ones
+
+        'pipeline_name' => $self->o('species_set_name') . '_' . $self->o('method_type') . '_update_' . $self->o('rel_with_suffix'),
+        'work_dir'      => $self->o('pipeline_dir'),
+        'output_dir'    => $self->o('work_dir') . '/feature_dumps/',
+        'bed_dir'       => $self->o('work_dir') . '/bed_dir/',
+
+        'master_db'         => 'compara_master',
+        'prev_db'           => 'compara_prev',
+        'reuse_db'          => $self->o('species_set_name') . '_' . $self->o('method_type') . '_prev',
+        'prev_ancestral_db' => $self->o('species_set_name') . '_ancestral',
+        
+        # EPOAncestral parameters
+        'ancestral_sequences_name'          => 'ancestral_sequences',
+        'ancestral_sequences_display_name'  => 'Ancestral sequences',
+        # Core ancestral database, created on the same server as the pipeline database
+        'ancestral_db'             => {
+            -driver  => $self->o('pipeline_db', '-driver'),
+            -host    => $self->o('pipeline_db', '-host'),
+            -port    => $self->o('pipeline_db', '-port'),
+            -user    => $self->o('pipeline_db', '-user'),
+            -pass    => $self->o('pipeline_db', '-pass'),
+            -species => $self->o('ancestral_sequences_name'),
+            -dbname  => $self->o('dbowner') . '_' . $self->o('species_set_name') . '_ancestral_core_' . $self->o('rel_with_suffix'),
+        },
+        
+        # EpoLowCoverage parameters
+        'epo_db'            => $self->pipeline_url(),
+        'low_epo_mlss_id'   => undef,  # required but unused
+        'base_epo_mlss_id'  => undef,  # required but unused
+        'max_block_size'    => 1000000,  # max size of alignment before splitting
+        'pairwise_location' => [ qw(compara_prev lastz_batch_*) ],  # default location for pairwise alignments (can be a string or an array-ref)
+        'lastz_complete'    => 0,  # set to 1 when all relevant LastZs have complete
+
+        # GERP default parameters
+        'run_gerp'          => 1,
+        'gerp_window_sizes' => [1, 10, 100, 500],
+
+        # Default statistics
+        'skip_multiplealigner_stats'    => 0,  # skip this module if set to 1
+
+        # Resource requirements
+        'gerp_capacity' => 500,
+    };
+}
+
+
+sub pipeline_checks_pre_init {
+    my ($self) = @_;
+    
+    die "The method type has to be in lowercase" if $self->o('method_type') =~ /^(EPO|PECAN)$/;
+    die "The method type '" . $self->o('method_type') . "' is not an updatable MSA" if $self->o('method_type') !~ /^(epo|pecan)$/;
+}
+
+
+sub pipeline_create_commands {
+    my ($self) = @_;
+
+    return [
+        @{$self->SUPER::pipeline_create_commands},  # inherit creation of DB, hive tables and compara tables
+
+        $self->pipeline_create_commands_rm_mkdir(['work_dir', 'output_dir', 'bed_dir']),
+    ];
+}
+
+
+sub pipeline_wide_parameters {
+    my ($self) = @_;
+    return {
+        %{$self->SUPER::pipeline_wide_parameters},  # inherit anything from the base class
+
+        'master_db'         => $self->o('master_db'),
+        'prev_db'           => $self->o('prev_db'),
+        'reuse_db'          => $self->o('reuse_db'),
+        'ancestral_db'      => $self->o('ancestral_db'),
+        'prev_ancestral_db' => $self->o('prev_ancestral_db'),
+
+        'method_type'       => $self->o('method_type'),
+        'species_set_name'  => $self->o('species_set_name'),
+        'curr_release'      => $self->o('ensembl_release'),
+
+        'genome_dumps_dir'  => $self->o('genome_dumps_dir'),
+
+        'lastz_complete'             => $self->o('lastz_complete'),
+        'run_gerp'                   => $self->o('run_gerp'),
+        'skip_multiplealigner_stats' => $self->o('skip_multiplealigner_stats'),
+    };
+}
+
+
+sub core_pipeline_analyses {
+    my $self = shift;
+
+    return [
+        {   -logic_name => 'load_mlss_ids',
+            -module     => 'Bio::EnsEMBL::Compara::RunnableDB::LoadMLSSids',
+            -input_ids  => [{}],
+            -parameters => {
+                'method_type'      => $self->o('method_type'),
+                'species_set_name' => $self->o('species_set_name'),
+                'release'          => '#curr_release#',
+                'add_prev_mlss'    => 1,
+                'add_sister_mlsss' => 1,
+                'branch_code'      => 2,
+            },
+            -rc_name    => '500Mb_job',
+            -flow_into  => {
+                '2->A'  => WHEN(
+                    '#method_type# eq "epo"' => { 'populate_new_database' => {
+                        'mlss_id_list' => [ '#mlss_id#', '#prev_mlss_id#', '#ext_mlss_id#', '#ce_mlss_id#', '#cs_mlss_id#' ],
+                    }},
+                    ELSE                        { 'populate_new_database' => {
+                        'mlss_id_list' => [ '#mlss_id#', '#prev_mlss_id#', '#ce_mlss_id#', '#cs_mlss_id#' ],
+                    }},
+                ),
+                'A->2'  => [ 'msa_update_factory' ],
+            },
+        },
+        {   -logic_name => 'populate_new_database',
+            -module     => 'Bio::EnsEMBL::Compara::RunnableDB::GenomicAlignBlock::PopulateNewDatabase',
+            -parameters => {
+                'program'           => $self->o('populate_new_database_exe'),
+                'reg_conf'          => $self->o('reg_conf'),
+                'old_compara_db'    => '#prev_db#',
+                'filter_by_mlss'    => 1,
+                'skip_gerp'         => 1,
+            },
+            -rc_name    => '500Mb_job',
+        },
+        # Update the MSA-related information in this and the ancestral core databases
+        {   -logic_name => 'msa_update_factory',
+            -module     => 'Bio::EnsEMBL::Hive::RunnableDB::Dummy',
+            -flow_into  => {
+                '1->A'  => WHEN(
+                    '#method_type# eq "epo"'    => {
+                        'set_gerp_mlss_tag' => INPUT_PLUS(),
+                        'set_internal_ids'  => INPUT_PLUS(),
+                        'mlss_factory'      => { 'inputlist' => [['#mlss_id#'], ['#ext_mlss_id#']] },
+                        'drop_ancestral_db' => { 'mlss_id' => '#mlss_id#', 'prev_mlss_id' => '#prev_mlss_id#' },
+                    },
+                    ELSE                           {
+                        'set_gerp_mlss_tag' => { 'ext_mlss_id' => '#mlss_id#', 'ce_mlss_id' => '#ce_mlss_id#', 'cs_mlss_id' => '#cs_mlss_id#' },
+                        'mlss_factory'      => { 'inputlist' => [['#mlss_id#']] },
+                    }
+                ),
+                'A->1'  => [ 'transfer_msa' ],
+            },
+        },
+        {   -logic_name => 'set_internal_ids',
+            -module     => 'Bio::EnsEMBL::Hive::RunnableDB::SqlCmd',
+            -parameters => {
+                'sql' => [
+                    'ALTER TABLE genomic_align AUTO_INCREMENT=#expr((#ext_mlss_id# * 10**10) + 1)expr#',
+                    'ALTER TABLE genomic_align_block AUTO_INCREMENT=#expr((#ext_mlss_id# * 10**10) + 1)expr#',
+                    'ALTER TABLE genomic_align_tree AUTO_INCREMENT=#expr((#ext_mlss_id# * 10**10) + 1)expr#'
+                ],
+            },
+        },
+        {   -logic_name => 'mlss_factory',
+            -module     => 'Bio::EnsEMBL::Hive::RunnableDB::JobFactory',
+            -parameters => {
+                'column_names'  => [ 'mlss_id' ],
+            },
+            -flow_into  => {
+                2 => [ 'make_species_tree' ],
+            },
+        },
+        {   -logic_name => 'make_species_tree',
+            -module     => 'Bio::EnsEMBL::Compara::RunnableDB::MakeSpeciesTree',
+            -parameters => {
+                'species_tree_input_file' => $self->o('binary_species_tree'),
+            },
+            -rc_name    => '500Mb_job',
+        },
+        # Copy the 4 tables with data in the ancestral core database and update the ancestor names with the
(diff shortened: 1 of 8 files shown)


CristiGuijarro (approved):
None of my comments are necessary, as it is mostly new code proven by several tests as a working pipeline :ok_hand:

modules/Bio/EnsEMBL/Compara/RunnableDB/UpdateMSA/TransferAlignment.pm
```diff
@@ -0,0 +1,108 @@
...
+        $dba->dbc->do("UPDATE genomic_align_block SET genomic_align_block_id = genomic_align_block_id + $id_diff,
+            method_link_species_set_id = $curr_mlss_id, group_id = group_id + $id_diff
+            WHERE method_link_species_set_id = $prev_mlss_id");
+        # Update all id columns in genomic_align_tree table
+        $dba->dbc->do("UPDATE genomic_align_tree SET node_id = node_id + $id_diff,
+            parent_id = parent_id + $id_diff, root_id = root_id + $id_diff,
+            left_node_id = left_node_id + $id_diff, right_node_id = right_node_id + $id_diff
+            WHERE node_id BETWEEN $min_id AND $max_id");
+        # If it is an EPO MSA, transfer the ancestral dnafrags to the new MLSS id
+        if ( $method_type =~ /^EPO$/i ) {
```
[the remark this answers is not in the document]
JAlvarezJarreta: `LoadMLSSids` uses lowercase, but our database (and other pipelines) use uppercase, so I left it case insensitive so that this runnable can be integrated in other places with less chances of breaking/needing updates. I had some troubles with "forcing" upper and lower cases with `uc()` and `lc()` in the pipeconfig, so I have tried to avoid them all around.

modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMSA_conf.pm
```diff
@@ -0,0 +1,442 @@
...
+        'pipeline_name' => $self->o('species_set_name') . '_' . $self->o('method_type') . '_update_' . $self->o('rel_with_suffix'),
+        'work_dir'      => $self->o('pipeline_dir'),
+        'output_dir'    => $self->o('work_dir') . '/feature_dumps/',
+        'bed_dir'       => $self->o('work_dir') . '/bed_dir/',
+
+        'master_db'         => 'compara_master',
+        'prev_db'           => 'compara_prev',
+        'reuse_db'          => $self->o('species_set_name') . '_' . $self->o('method_type') . '_prev',
+        'prev_ancestral_db' => $self->o('species_set_name') . '_ancestral',
+
```
carlacummins, line 67:
```suggestion

```

modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMSA_conf.pm
```diff
@@ -0,0 +1,442 @@
...
+        'ancestral_db'             => {
+            -driver  => $self->o('pipeline_db', '-driver'),
+            -host    => $self->o('pipeline_db', '-host'),
+            -port    => $self->o('pipeline_db', '-port'),
+            -user    => $self->o('pipeline_db', '-user'),
+            -pass    => $self->o('pipeline_db', '-pass'),
+            -species => $self->o('ancestral_sequences_name'),
+            -dbname  => $self->o('dbowner') . '_' . $self->o('species_set_name') . '_ancestral_core_' . $self->o('rel_with_suffix'),
+        },
+
```
carlacummins, line 81:
```suggestion

```

modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMSA_conf.pm
```diff
@@ -0,0 +1,442 @@
...
+        # Resource requirements
+        'gerp_capacity' => 500,
+    };
+}
+
+
+sub pipeline_checks_pre_init {
+    my ($self) = @_;
+
```
carlacummins, line 105:
```suggestion

```

modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMSA_conf.pm
```diff
@@ -0,0 +1,442 @@
...
+            -parameters => {
+                'inputquery'    => 'SELECT gat2.root_id, #mlss_id# FROM genomic_align_tree gat1 LEFT JOIN genomic_align ga USING (node_id) JOIN genomic_align_tree gat2 USING (root_id) WHERE gat2.parent_id IS NULL AND ga.method_link_species_set_id = #mlss_id# GROUP BY gat2.root_id',
+                'column_names'  => [ 'root_id', 'mlss_id' ],
+            },
+            -rc_name    => '2Gb_job',
+            -flow_into  => {
+                '2->A' => [ 'set_neighbour_nodes' ],
+                'A->1' => { 'update_max_alignment_length' => { 'method_link_species_set_id' => '#mlss_id#', 'method_type' => '#method_type#' } },
+            },
+        },
```
carlacummins, line 354:
```suggestion
        },
```

modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMSA_conf.pm
```diff
@@ -0,0 +1,442 @@
...
+            -rc_name    => '2Gb_job',
+            -flow_into  => {
+                '2->A' => [ 'set_neighbour_nodes' ],
+                'A->1' => { 'update_max_alignment_length' => { 'method_link_species_set_id' => '#mlss_id#', 'method_type' => '#method_type#' } },
+            },
+        },  
+        {   -logic_name    => 'set_neighbour_nodes',
+            -module        => 'Bio::EnsEMBL::Compara::RunnableDB::EpoLowCoverage::SetNeighbourNodes',
+            -rc_name       => '2Gb_job',
+            -batch_size    => 10,
```
carlacummins, line 358:
```suggestion
            -batch_size    => 10,
```

modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMSA_conf.pm
```diff
@@ -0,0 +1,442 @@
...
+            -flow_into  => {
+                '2->A' => [ 'set_neighbour_nodes' ],
+                'A->1' => { 'update_max_alignment_length' => { 'method_link_species_set_id' => '#mlss_id#', 'method_type' => '#method_type#' } },
+            },
+        },  
+        {   -logic_name    => 'set_neighbour_nodes',
+            -module        => 'Bio::EnsEMBL::Compara::RunnableDB::EpoLowCoverage::SetNeighbourNodes',
+            -rc_name       => '2Gb_job',
+            -batch_size    => 10, 
+            -hive_capacity => 20,
```
carlacummins, line 359:
```suggestion
            -hive_capacity => 20,
```

carlacummins (changes requested):
This is a very incomplete review, but github is acting strangely and I want to submit what I have before it gets lost. Will have more comments coming.

On a related note, I recently installed this atom package which highlights trailing whitespace in bright yellow. Has been super useful so far. https://atom.io/packages/trailing-spaces

modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMSA_conf.pm
```diff
@@ -0,0 +1,442 @@
...
+        # Core ancestral database, created on the same server as the pipeline database
+        'ancestral_db'             => {
+            -driver  => $self->o('pipeline_db', '-driver'),
+            -host    => $self->o('pipeline_db', '-host'),
+            -port    => $self->o('pipeline_db', '-port'),
+            -user    => $self->o('pipeline_db', '-user'),
+            -pass    => $self->o('pipeline_db', '-pass'),
+            -species => $self->o('ancestral_sequences_name'),
+            -dbname  => $self->o('dbowner') . '_' . $self->o('species_set_name') . '_ancestral_core_' . $self->o('rel_with_suffix'),
+        },
```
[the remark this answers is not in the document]
JAlvarezJarreta: This is a copy-paste from `EPOwith2x`: https://github.com/Ensembl/ensembl-compara/blob/release/103/modules/Bio/EnsEMBL/Compara/PipeConfig/EPOwith2x_conf.pm#L100-L108
I guess it is done this way so you don't need to configure the registry *before* the pipeline has finished, as we usually do with all the other pipelines.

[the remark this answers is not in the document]
muffato: (yes, that was the reason)

modules/Bio/EnsEMBL/Compara/RunnableDB/UpdateMSA/UpdateGAB.pm
```diff
@@ -0,0 +1,279 @@
...
+    # Get the new CIGAR lines, discarding any gap-only column produced by the removal of the marked GAs
+    my @prev_cigars = map { $_->cigar_line } @$genomic_aligns;
+    my @updated_cigars = Bio::EnsEMBL::Compara::Utils::Cigars::minimize_cigars(@prev_cigars);
+    my %gas_to_update;
+    while ( my ($index, $cigar) = each @updated_cigars ) {
+        next if $cigar eq $prev_cigars[$index];
+        $gas_to_update{$genomic_aligns->[$index]->dbID} = $cigar;
+    }
+    # Get the new alignment length
+    my $new_aln_length = Bio::EnsEMBL::Compara::Utils::Cigars::alignment_length_from_cigar($updated_cigars[0]);
```
carlacummins, line 266: I feel like removing gap only columns would be a useful method to have on the GenomicAlignBlock object itself (and it seems like the logical place to keep this sort of functionality)

modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMSA_conf.pm
```diff
@@ -0,0 +1,437 @@
...
+                'ancestral_db' => $self->o('ancestral_db'),
+            },
+            -flow_into  => [ 'remove_prev_mlss' ],
+        },
+        {   -logic_name => 'remove_prev_mlss',
+            -module     => 'Bio::EnsEMBL::Hive::RunnableDB::SqlCmd',
+            -parameters => {
+                'wrap_in_transaction' => 1,
+                'sql'                 => [
+                    'SET FOREIGN_KEY_CHECKS=0',
```
carlacummins, line 297: There's no order to do these `DELETE`s in that doesn't require FKs to be turned off?

modules/Bio/EnsEMBL/Compara/RunnableDB/UpdateMSA/SyncAncestralDB.pm
```diff
@@ -0,0 +1,78 @@
...
+    my $ancestral_db = $self->param_required('ancestral_db');
+    my $ancestor_names = $self->param_required('ancestor_names');
+
+    my $dbc = Bio::EnsEMBL::Hive::Utils::go_figure_dbc($ancestral_db);
+    # Speed up deleting data by disabling the tables keys
+    $dbc->do("ALTER TABLE `dna` DISABLE KEYS");
+    $dbc->do("ALTER TABLE `seq_region` DISABLE KEYS");
+    # Delete the sequence regions that have been removed from the MSA as well
+    my $nrows = $dbc->do("DELETE seq_region, dna FROM seq_region JOIN dna USING (seq_region_id)
+        WHERE name IN ('" . join("','", @$ancestor_names) . "')");
```
carlacummins, line 70: How large are you expecting `@$ancestor_names` to be? The `IN (...)` starts to break down after ~500, I believe. A loop might be a better bet here.

carlacummins (commented):
Overall seems fine. Couple of extra comments on top of what I've added inline with the code:
1. not sure how long the first part of the pipeline takes, but for EPOwith2x, we have the `check_for_lastz` analysis as a 'switch' to stop/start the pipeline at the `low_coverage_alignment` part - do we want to keep that switch?
2. a huge amount of this pipeline is running SQL commands directly on existing objects rather than using the API. I could find several cases where an already-existing function was reimplemented in this code. it's not ideal.
3. when updating the GABs, i would have greatly favoured creating a copy of objects that needed to updated, updating them with API methods and then saving the copy/mutated version of the object before deleting the original one. this would also remove the need to run `TransferAlignment` at all, since all that's doing is updating ids

modules/Bio/EnsEMBL/Compara/RunnableDB/UpdateMSA/SyncAncestralDB.pm
muffato: FYI, there is a method to split a list into 500-element chunks and call a subroutine on each: `$adaptor->split_and_callback`

modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMSA_conf.pm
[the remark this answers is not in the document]
muffato: Adding more views to the discussion ...

There is a difference though: currently we only edit the required _input_ databases in the registry file before starting a pipeline, and the location of the pipeline (i.e. its output) is set on the `init_pipeline.pl` command-line. This `ancestral_db` is the rare case of a pipeline having a second _output_ database which, as it is/was, draws its parameters from the same command-line options.
There is the case of `compara_curr`, which several pipelines used for (incremental) output, and is assumed to be there in the registry file ... **But** it's a _shared_ output database, as opposite to the ancestral db which entirely comes from this pipeline.

I also think that, when doing a test run of the pipeline, there is a higher risk of erasing an _existing_ ancestral database if it uses the registry file instead of creating its own, new database.

:shrug:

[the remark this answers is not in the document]
muffato: Two more things to consider

1. I seem to remember that loading the registry file will _fail_ if a database doesn't exist, so it can't be used to tell where to create a database
2. In the pipeline the location of the ancestral db is actually primarily stored in `genome_db.locator`. Once the genome_db has been created, the source of the information (pipeline param or registry file doesn't matter any more).  I see that as a risk in the sense that when doing a test run, I may forget about that, edit the registry file, and start my tests without realising the tests will modify the true ancestral database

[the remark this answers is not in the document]
carlacummins: Well I think that's the nail in the coffin for setting it in the registry - let's leave it as is.

modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMSA_conf.pm
```diff
@@ -0,0 +1,437 @@
...
+                    'UPDATE seq_region SET name = REPLACE(name, "_#prev_mlss_id#_", "_#mlss_id#_")',
+                ],
+            },
+        },
+        # Transfer MLSS, GA* and dnafrag ids from the previous MLSS id to the current one, update the MSA and
+        # ancestral core database, and finally remove the previous MSA information
+        {   -logic_name => 'transfer_msa',
+            -module     => 'Bio::EnsEMBL::Compara::RunnableDB::UpdateMSA::TransferAlignment',
+            -parameters => {
+                'method_type' => $self->o('method_type'),
```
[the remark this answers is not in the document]
JAlvarezJarreta: Since it is already provided as a pipeline parameter, I thought it was faster to just pass it than having to use the API to get it again.

modules/Bio/EnsEMBL/Compara/RunnableDB/UpdateMSA/UpdateGAB.pm
JAlvarezJarreta: To be added to the pipeline's 2.0 version ;)

modules/Bio/EnsEMBL/Compara/RunnableDB/UpdateMSA/UpdateGAB.pm
```diff
@@ -0,0 +1,279 @@
...
+            my $sth = $dba->dbc->prepare("UPDATE genomic_align_tree SET parent_id = ?, distance_to_parent = ? WHERE node_id = ?");
+            foreach my $node_id ( keys %$node_ids_to_update ) {
+                my $values = $node_ids_to_update->{$node_id};
+                if ( ! defined $values->{parent_id} ) {
+                    # This node is going to be the new root: update the root_id of the entire GAT
+                    $dba->dbc->do("UPDATE genomic_align_tree SET root_id = $node_id WHERE root_id = " . $values->{prev_root_id});
+                }
+                $sth->execute($values->{parent_id}, $values->{distance}, $node_id);
+            }
+        }
```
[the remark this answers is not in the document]
JAlvarezJarreta: I didn't try them all, but I tested a few of these methods and they didn't do what one would expect, and there wasn't time to implement them from scratch, so this was the "easiest" approach.

modules/Bio/EnsEMBL/Compara/RunnableDB/UpdateMSA/UpdateGAB.pm
```diff
@@ -0,0 +1,279 @@
...
+
+=cut
+
+package Bio::EnsEMBL::Compara::RunnableDB::UpdateMSA::UpdateGAB;
+
+use warnings;
+use strict;
+
+use Bio::EnsEMBL::Registry;
+use Bio::EnsEMBL::Hive::Utils;
```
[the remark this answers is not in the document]
JAlvarezJarreta: Default Runnable snippet, sorry.

modules/Bio/EnsEMBL/Compara/RunnableDB/UpdateMSA/UpdateGAB.pm
```diff
@@ -0,0 +1,279 @@
...
+        $root->build_leftright_indexing();
+        $gat_adaptor->update_subtree($root);
+    }
+
+    # Flow the ancestor dnafrag names to syncronise the ancestral database with the dnafrag table
+    $self->dataflow_output_id({ name => $_ }, 2) for @ancestor_names;
+}
+
+
+sub _remove_gab {
```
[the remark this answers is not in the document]
JAlvarezJarreta: What about `_prep_gab_to_rm` and `_prep_gab_to_update`? (not finding any better names atm)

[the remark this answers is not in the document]
carlacummins: Sure, or `_register_gab_for_rm`/`_register_gab_for_update`

carlacummins:
Just to mention: I think the code as it is will work and should be merged largely unchanged (just addressing the smaller issues now) since we need it for e103 and this version has been tested. As you say, a v2.0 complete with unit tests will be prepared in future and we can talk about the larger-scale restructure then. Thanks 🙂

modules/Bio/EnsEMBL/Compara/RunnableDB/UpdateMSA/UpdateGAB.pm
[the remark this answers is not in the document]
CristiGuijarro: Perhaps `_allocate_gab_to_rm` and `_allocate_gab_to_update` :grin:

modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMSA_conf.pm
JAlvarezJarreta: I have checked and I didn't know you can actually tell `DELETE` to do its job in a specific order (we need to remove `node_id` in decreasing order so it doesn't conflict with `parent_id` in `species_tree_node`).

modules/Bio/EnsEMBL/Compara/RunnableDB/UpdateMSA/SyncAncestralDB.pm
JAlvarezJarreta: I tested it with mammals EPO e102 and it seemed to work without issues (not sure about the length of this array). But I will apply @muffato's suggestion as a prevention.

JAlvarezJarreta:
> Overall seems fine. Couple of extra comments on top of what I've added inline with the code:
> 
> 1. not sure how long the first part of the pipeline takes, but for EPOwith2x, we have the `check_for_lastz` analysis as a 'switch' to stop/start the pipeline at the `low_coverage_alignment` part - do we want to keep that switch?

By default, the species being moved from EPO to EPO Extended will have an updated assembly, meaning we will need to compute new LastZs for them, so that is why the switch has remained. The ancestral db update is slow. The GAB update is quite fast overall, but dependant on the relevance of the species being removed.

> 2. a huge amount of this pipeline is running SQL commands directly on existing objects rather than using the API. I could find several cases where an already-existing function was reimplemented in this code. it's not ideal.

The reimplementation in many cases has been done on purpose. For GATs, several API methods I could find didn't do what was expected of them to do, and many others are just missing. For other cases, the problem is that the API is not always consistent: sometimes it will automatically trigger an update in the database, others you have to trigger it, making it quite hard to make the work in this pipeline robust to failures. But happy to give it another go for version 2.0.

> 3. when updating the GABs, i would have greatly favoured creating a copy of objects that needed to updated, updating them with API methods and then saving the copy/mutated version of the object before deleting the original one. this would also remove the need to run `TransferAlignment` at all, since all that's doing is updating ids

With the current functionality of the pipeline, what you propose wouldn't remove the need of `TransferAlignment` as not all the GABs are flowed to the analysis: only those affected by the "removal" of the species. Furthermore, I opted for this approach to avoid increasing the number of rows of these quite large tables (specially `genomic_align`) even more, as we are already reaching performance issues in some cases.

carlacummins:
> that is why the switch has remained

I actually thought you'd removed it - I didn't expect the name to be changed to `epo_extended_rib`

carlacummins:
> the API is not always consistent

I would see that as a reason to make it consistent, not work around it

carlacummins:
> I opted for this approach to avoid increasing the number of rows

What I've suggested is to create a copy of the API objects, mutate the copy, store the copy and immediately delete the original object. The duplication would last seconds at most.

carlacummins (approved):

PR #243, opened 2020-11-10
Move the meadow-specific parameters to a JSON config file

muffato:
## Description

I was thinking about how to fold back the parameters (paths, etc) of the `experimental/cloud` branch in line with `master` so that both branches don't diverge too much (the more we wait, higher the risk of a conflict). I got inspired by the eHive mechanism to detect the local environment and choose the parameters accordingly. See https://github.com/Ensembl/ensembl-hive/blob/version/2.5/hive_config.json#L17-L38. This is within a _Meadow_ section, and there are two layers: the first one is the "type" of the "meadow", e.g. `LSF`, `LOCAL`, `SLURM`, etc, the second one is the name of the cluster, e.g. `EBI` (for noah), `YODA`, etc.

What also motivated me is the upcoming farm change. I expect there will be some sort of transition period where we'll be discovering the new farm, but not yet able to run everything on it, so we would have to use noah at the same time. It's not practical to keep two versions of `ENV.pm`.

## Overview of changes

I'm not so fond about having everything for all known farms in one file, so here is a variant where there is one file per farm, named `<type>.<name>.json`, under `ensembl-compara/conf/software`.

All the farm-dependent paths (executables and directories) are now in the JSON files, and ComparaGeneric_conf has a method to detect the current farm and load the file. In the PR I include the JSON file of our main EBI farm (basically the same parameters as in the current `ENV.pm`), but also what I think would work in the cloud Slurm cluster (taken from `ENV.pm` of the `experimental/cloud` branch), and also a file that matches my own laptop setup.

## Testing

I have initialised pipelines on my laptop and the EBI farm, and checked that the parameters are correct.

## Notes

1. The main caveat is that all the parameters in the JSON file have the _lowest_ priority, i.e. they will be overridden by anything in the actual pipeline PipeConfig (because they're set in ComparaGeneric and the subclasses do first `$self->SUPER::default_options`). For instance, `treerecs_exe` was defined in ProteinTrees_conf directly. I had to remove it from the latter in order for the JSON source to be considered. Another example: to be able to define `panther_annotation_file` (which is currently in ProteinTrees_conf) I would have to remove it from the PipeConfig and define it in all JSON files, or set a meadow-agnostic default value in `ENV::shared_default_options()` and override it in the required JSON files. In other words, the various pipelines need to _allow_ a parameter to be defined in the JSON files, it's not something that's yet possible for all parameters.
2. The syntax I have used is admittedly weird. String values can do `$self->o()`-style parameter substitution with a pair of **double** hashes, e.g. `##linubrew_home##/bin`. Array values are expected to be of the form `["function_name", "arg1", "arg2", ...]`, where the function is one of the `check_exe_in_cellar` family. In both cases, I tried to find a way of accommodating the need. If you've got a better idea, the floor is yours :)

I see this change as a stepping stone. With such a mechanism in place, we could do some more serious refactoring, and things that we haven't yet really envisaged. For instance, if the future is a mix of a on-premises farm and cloud, and we end up running a pipeline across the farm and the cloud _at the same time_ (connecting to the same database) we could change all `$self->param('xxxx_exe')` calls to `$self->get_path_in_meadow('xxxx_exe')`, which would look in the JSON file. Those paths wouldn't have to be stored in the database, they would be loaded from the relevant JSON file at runtime.


carlacummins (approved):
Nice update!

conf/software/LOCAL.mattxps.json
```diff
@@ -0,0 +1,73 @@
+{
+    "linuxbrew_home"            : "/home/linuxbrew/.linuxbrew",
+    "hps_dir"                   : "/home/matthieu/tmp/hps",
+    "warehouse_dir"             : "/home/matthieu/tmp/warehouse/compara/",
```
JAlvarezJarreta, line 4: I understand where these come from, but I'm always a little bit against "proprietary" configurations (should we end up with one per Compara member?). So I'm opening a discussion here: should this file be converted to a LOCAL template, maybe renaming it `LOCAL.template.json` and replace `matthieu` by `$USER`, or is it ok to have one LOCAL JSON file per member/laptop?

conf/software/LOCAL.mattxps.json
```diff
@@ -0,0 +1,73 @@
...
+    "rapidnj_exe"               : ["check_exe_in_cellar", "rapidnj/2.3.2/bin/rapidnj"],
+    "raxml_exe_avx"             : ["check_exe_in_cellar", "raxml/8.2.8/bin/raxmlHPC-AVX"],
+    "raxml_exe_sse3"            : ["check_exe_in_cellar", "raxml/8.2.8/bin/raxmlHPC-SSE3"],
+    "raxml_pthread_exe_avx"     : ["check_exe_in_cellar", "raxml/8.2.8/bin/raxmlHPC-PTHREADS-AVX"],
+    "raxml_pthread_exe_sse3"    : ["check_exe_in_cellar", "raxml/8.2.8/bin/raxmlHPC-PTHREADS-SSE3"],
+    "samtools_exe"              : ["check_exe_in_cellar", "samtools/1.10/bin/samtools"],
+    "semphy_exe"                : ["check_exe_in_cellar", "semphy/2.0b3/bin/semphy"],
+    "server_exe"                : ["check_exe_in_cellar", "exonerate24/2.4.0/bin/exonerate-server"],
+    "treebest_exe"              : ["check_exe_in_cellar", "treebest/rc5/bin/treebest"],
+    "treerecs_exe"              : "/home/matthieu/src/gene_tree/Treerecs/bin/treerecs",
```
JAlvarezJarreta, line 70: Related to previous open discussion.

JAlvarezJarreta (approved):

conf/software/LOCAL.mattxps.json
muffato: The content itself of `LOCAL.mattxps.json` is actually very similar to the other files. I think both the LSF and SLURM files could be considered as "templates" for your own laptop, but you'd have to remember how to name the file.

JAlvarezJarreta: Then we can leave it as it is :+1:

PR #244, opened 2020-11-11
Minor changes to format of table.sql file

james-monkeyshines:
## Description
The `CompareSchema` datacheck has been updated to include compara dbs, but there are some small changes that can be made in the table.sql file to avoid additional complexity in the datacheck code. These changes are synonymous as far as MySQL is concerned.
Datacheck PR: https://github.com/Ensembl/ensembl-datacheck/pull/324

**Related JIRA tickets:**
https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-6233

## Testing
Datacheck was run successfully for all divisions.

### DIFF
--- a/sql/table.sql
+++ b/sql/table.sql
@@ -178,7 +178,7 @@ CREATE TABLE genome_db (
   FOREIGN KEY (taxon_id) REFERENCES ncbi_taxa_node(taxon_id),
 
   PRIMARY KEY (genome_db_id),
-  UNIQUE name (name,assembly,genome_component)
+  UNIQUE KEY name (name,assembly,genome_component)
 
 ) COLLATE=latin1_swedish_ci ENGINE=MyISAM;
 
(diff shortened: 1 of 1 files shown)


CristiGuijarro (approved):

sql/table.sql
```diff
@@ -1520,7 +1520,7 @@ CREATE TABLE gene_tree_node (
   root_id                         int(10) unsigned,
   left_index                      int(10) NOT NULL DEFAULT 0,
   right_index                     int(10) NOT NULL DEFAULT 0,
-  distance_to_parent              double default 1.0 NOT NULL,
+  distance_to_parent              double default 1 NOT NULL,
```
JAlvarezJarreta, line 1523: Just a comment to leave here: not sure if the float-point approximation will do its magic when the default is an integer instead of float. But let's hope for the best :+1:

JAlvarezJarreta (approved):

sql/table.sql
muffato: About this one, what was the problem here @james-monkeyshines ? Is it that the parser expects at least one non-zero digit after the decimal point ? or that it doesn't support decimal values at all ?

james-monkeyshines: It seems as though mysql ignores the decimal bit when creating the table; the datacheck is doing a `show create table` to get the definition, and that has `double default 1 NOT NULL`.

PR #251, opened 2020-12-09
plant config for release/104

brunocontrerasmoreira:
## Description

This PR addresses new and updated plant species announced in https://www.ebi.ac.uk/panda/jira/browse/ENSINT-639 and https://www.ebi.ac.uk/panda/jira/browse/ENSINT-638

@JAlvarezJarreta @carlacummins 

## Overview of changes

- We update the species tree but leave out the new wheat cultivars as these are not yet subject to Compara analyses @gnaamati 
- We update the list of allowed species, removing two low-usage species and adding two new ones


muffato:
@brunocontrerasmoreira . We have run two wheat LastZ alignments (against the _Stanley_ cultivar and against spelt) and we can release them in e104 if you want, but that would require adding these two genomes to `allowed_species.json`. I don't know whether that means removing two other genomes (@dthybert  ?)

brunocontrerasmoreira:
Thanks @muffato @dthybert , we would argue that the wheat work (part of the DFW grant) should not be under the same limitations of the other species. Same with rice, as the PanOryza grant just started. Let me know if you agree,
Bruno

JAlvarezJarreta (approved):
Looks fine to me. @marcoooo, I think you may be interested in knowing about these changes as well.

marcoooo:
> Looks fine to me. @marcoooo, I think you may be interested in knowing about these changes as well.

Yes I am, though our dispatch won't be installed live for this release, I am afraid.

dthybert:
> Thanks @muffato @dthybert , we would argue that the wheat work (part of the DFW grant) should not be under the same limitations of the other species. Same with rice, as the PanOryza grant just started. Let me know if you agree,
> Bruno

The reason why  we caped the number of species is to avoid having a gene tree pipeline become too big at some point. So we need to have the number of species in the plant protein tree should stay to 100 species. Saying this,  we could start thinking of having for the rice and the wheat the same approach that we have for mouse strain or pig with a rice trees/alignment and wheat trees/alignment. However if we go down that route, I would like to put in place a better planing of the species entering into compara with at least two releases in advance(i.e at 104 already planing for 106), the main reason is that we want to avoid having to run too many pipeline in each production and so we will not want to come in a case that rice + wheat + other plants are updated in the same release  and therefore  would trigger the three protein tree pipeline and the rice multiple alignment + lastz which would be too much for us (especially that we are importing additional divisions into compara). It may be too late for 104 to put this in place though.

brunocontrerasmoreira:
Thanks for your comments @dthybert . @gnaamati and me will try to let you know well in advance.
1) For wheat the calendar is not in our hands, as the people doing the annotation have their own pace.
2) For rice, the PanOryza project just started, and it is likely that the new cores will take a while to be handed over, definetely not in release 104.

dthybert:
> * For wheat the calendar is not in our hands, as the people doing the annotation have their own pace.
> * For rice, the PanOryza project just started, and it is likely that the new cores will take a while to be handed over, definetely not in release 104.

dthybert:
> @brunocontrerasmoreira . We have run two wheat LastZ alignments (against the _Stanley_ cultivar and against spelt) and we can release them in e104 if you want, but that would require adding these two genomes to `allowed_species.json`. I don't know whether that means removing two other genomes (@dthybert ?)

After discussion with Bruno we think it is better to postpone the inclusion of the pairwise alignment as we would remove to many species to be able to include the wheat. The alignment will be included when we will start to do a wheat specific protein tree.

dthybert (approved):

ipilizota (approved):

PR #259, opened 2021-03-09
Split Datacheck pipeline into integratable parts

CristiGuijarro:
## Description

The datacheck pipeline `DatachecksForRelease_conf` works well for the release databases, but is rigidly set in this way. We want to split the pipeline into parts so that the parallel running datachecks can be easily integrated at the end of other pipelines and run based on a specified `group` parameter.

**Related JIRA tickets:**
- ENSCOMPARASW-4345

## Overview of changes
From one file to three.

#### Change 1
- The non-parallel `run_datacheck` analyses have been removed altogether. We will never want to run the DCs linearly one after the other in release and we have a single runnable to incorporate into a pipeline if we do want to run a single specific datacheck.

#### Change 2
- The integratable parts have been split into 2 for flexible use cases. The first is the bigger `ParallelDataCheckFactory` which incorporates the whole package down to storing the results in the pipeline and emailing the results to the user, a better use for tagging on the end of pipelines.
- The second part is the smaller part that simply contains the 3 analyses: `datacheck_fan`, `datacheck_fan_high_mem` and `create_jira_tickets` - this smaller package may be better used midway through pipelines rather than at the end or when a hard-coded list of [`datacheck_names`] may be needed instead.

## Testing
The pipeline has been initialised and run and shown to work. An [example pipeline here.](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensadmin&host=mysql-ens-compara-prod-10&port=4648&dbname=cristig_homology_annot_dc_test&passwd=xxxxx)

## Notes
The `jira_ticket_creation` can be easily "switched off" by passing it `test_mode=1` parameter - this may be a better use during pipeline integrated DCs with no failure tolerance.
I made some white-space changes - but given I was the original author, and I don't think they affect PR readability too much - I think these should be allowed.
I have based this against the `feature/homology_annotation` branch with the intention of cherry picking the single commit over to master - there will be no file conflicts.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/DatachecksForRelease_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/DatachecksForRelease_conf.pm
@@ -36,8 +36,9 @@ package Bio::EnsEMBL::Compara::PipeConfig::DatachecksForRelease_conf;
 
 use strict;
 use warnings;
-use Bio::EnsEMBL::Hive::Version 2.5;
+use Bio::EnsEMBL::Hive::Version 2.4;
 use Bio::EnsEMBL::Hive::PipeConfig::HiveGeneric_conf; # for WHEN and INPUT_PLUS
+use Bio::EnsEMBL::Compara::PipeConfig::Parts::DataCheckFactory;
 
 use base ('Bio::EnsEMBL::Compara::PipeConfig::ComparaGeneric_conf');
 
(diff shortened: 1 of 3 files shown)


modules/Bio/EnsEMBL/Compara/PipeConfig/DatachecksForRelease_conf.pm
```diff
@@ -36,8 +36,9 @@ package Bio::EnsEMBL::Compara::PipeConfig::DatachecksForRelease_conf;
 
 use strict;
 use warnings;
-use Bio::EnsEMBL::Hive::Version 2.5;
+use Bio::EnsEMBL::Hive::Version 2.4;
```
JAlvarezJarreta, line 39: I always get confused myself in to which version "support" :sweat_smile:

modules/Bio/EnsEMBL/Compara/PipeConfig/Parts/ParallelDataCheckFactory.pm
```diff
@@ -0,0 +1,87 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=head1 NAME
+
+Bio::EnsEMBL::Compara::PipeConfig::Parts::NewPart
```
JAlvarezJarreta, line 20:
```suggestion
Bio::EnsEMBL::Compara::PipeConfig::Parts::DataCheckFactory
```
(or `Bio::EnsEMBL::Compara::PipeConfig::Parts::ParallelDataCheckFactory`)

modules/Bio/EnsEMBL/Compara/PipeConfig/Parts/ParallelDataCheckFactory.pm
```diff
@@ -0,0 +1,87 @@
...
+
+=head1 DESCRIPTION
+
+    This is a part of a the DataCheck pipe_config containing the datacheck analyses
```
JAlvarezJarreta, line 24:
```suggestion
    This is a part of the DataCheck pipeline containing the datacheck analyses.
```

JAlvarezJarreta (approved):
Like the idea of having them in two parts very much :+1: 
A small suggestion: since the sequential DC has been removed, why not naming `ParallelDataCheckFactory` just `DataCheckFactory`? (and `pipeline_analyses_parallel_datacheck_factory` just `pipeline_analyses_datacheck_factory`)

PR #260, opened 2021-03-10
Bugfixes and updates on the Reference pipeline

JAlvarezJarreta:
## Description

Several small fixes to make `UpdateReferenceDatabase` pipeline ready to load all references from every division (but `bacteria`).

**Related JIRA tickets:**
- ENSCOMPARASW-4360 (partial)
- ENSCOMPARASW-4260

#### Change 1
- Revert commit 7b6bf9d: the species was renamed in e104 and I mistakenly thought it was a typo.

#### Change 2
- `genebuild.last_geneset_update` is an optional `meta` tag that it is usually only present in vertebrates. Follow the same procedure as Production: if the tag is not present, use `genebuild.start_date` instead (mandatory tag).

#### Change 3
- Make `UpdateGenomesFromMetadataFactory.pm` runnable able to handle metadata information from all divisions. For this, `division` needs to be an empty string because it will use the pipeline-wide parameters instead if not provided.

#### Change 4
- `stable_id` still needs to be unique, but limit this uniqueness per `genome_db_id` to support multiple `stable_ids` in other assemblies of the same species.

## Testing
I've created a small references database with 6 vertebrates from e102 and then updated them from e104, and everything has worked without issues: [mysql://ensro@mysql-ens-compara-prod-2:4522/jalvarez_update_prev_references](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-2&port=4522&dbname=jalvarez_update_prev_references)
A first attempt to load all the references can be found here: [mysql://ensro@mysql-ens-compara-prod-2:4522/jalvarez_update_references_104](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-2&port=4522&dbname=jalvarez_update_references_104), but there are still some things that need to be addressed/fixed to make it fully work.

## Notes
- Couldn't think of a better approach for `stable_id`: the ideal would be to prevent duplicates in `stable_id` *for two different species* (as it was before) but allow it if the species are the same (but different assembly), but that is not possible with the current schema.
- Taken the opportunity to update some PODs

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateReferenceDatabase_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateReferenceDatabase_conf.pm
@@ -184,9 +184,10 @@ sub core_pipeline_analyses {
                 'meta_host'             => $self->o('meta_host'),
                 'allowed_species_file'  => $self->o('config_dir') . '/allowed_species.json',
                 'perc_threshold'        => $self->o('perc_threshold'),
-                'division'              => 'vertebrates',
+                'division'              => '',  # our references cover every division
                 'master_db'             => '#ref_db#',
             },
+            -rc_name    => '1Gb_job',
             -flow_into  => {
                 '2->A' => [ 'update_reference_genome' ],
                 '3->A' => [ 'retire_reference' ],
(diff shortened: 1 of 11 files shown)


CristiGuijarro (approved):

modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateReferenceDatabase_conf.pm
```diff
@@ -184,9 +184,10 @@ sub core_pipeline_analyses {
                 'meta_host'             => $self->o('meta_host'),
                 'allowed_species_file'  => $self->o('config_dir') . '/allowed_species.json',
                 'perc_threshold'        => $self->o('perc_threshold'),
-                'division'              => 'vertebrates',
+                'division'              => '',  # our references cover every division
```
carlacummins, line 187: Is it better set to `undef` then? It's seems to be what we usually do..
```suggestion
                'division'              => undef,  # our references cover every division
```

carlacummins (approved):
Looks great! It's quite elegant in the end 😄 

Would you mind renaming the PR to something more descriptive before we merge?

PR #262, opened 2021-03-15
Offset members by genome_db_id

JAlvarezJarreta:
## Description

Add a `genome_db_id`-based offset of members to avoid clashes between the reference and the homology annotation pipelines.

**Related JIRA tickets:**
- ENSCOMPARASW-4075

#### Change 1
- The `member_id` is done cleanly by two single queries thanks to the update in the schema to `ON UPDATE CASCADE`, and using a variable to do an offset autoincrement per `genome_db_id`.

#### Change 2
- Added `offset_id` flag to preserve previous behaviour as default.

#### Change 3
- Both `gene_member_id` and `seq_member_id` are now `BIGINT` to avoid having to change this in quite some time, allowing plenty of room for many genomes as well as members.

## Testing
I have done several tests but one that you can have a look at is [jalvarez_update_prev_references_104](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-2&port=4522&dbname=jalvarez_update_prev_references_104), in which I loaded a few e102 genomes first and then added some more references from e104 (updating the previous references when required). The reference database is in `cp3`: `jalvarez_reference_db_test_update`.

## Notes
- I have opted for a change in the schema only in the two affected tables for now. In the future, if required, we could extend this to any other tables we may want to update automatically as well.
- Bear in mind this change only works with `InnoDB` (which is the one we use internally as default).
- That specific test has been also used to make a decision regarding ENSCOMPARASW-4287.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateReferenceDatabase_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateReferenceDatabase_conf.pm
@@ -73,6 +73,7 @@ sub default_options {
         'store_others'                => 0,
         'store_exon_coordinates'      => 0,
         'store_related_pep_sequences' => 0, # do we want CDS sequences as well as protein sequences?
+        'offset_ids'                  => 1, # offset member ids by the genome_db_id?
 
         # member HC options
         'allow_ambiguity_codes'         => 1,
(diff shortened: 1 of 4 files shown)


CristiGuijarro (changes requested):
Seems simple enough :+1: 
Not sure about the `int -> bigint` changes - a little overkill? `int` should be enough I think - at least until we have literally thousands of references (which won't be for a very long time - someone else's problem).

CristiGuijarro (approved):

PR #265, opened 2021-03-16
Feature/homology annotation ref set

dthybert:
## Description
Refinement of reference sets using subtree depth, busco and N50 statistics. 
Addition o shared_vertebrate collection

**Related JIRA tickets:**
- ENSCOMPARASW-4363

## Overview of changes

1- in mlss_conf.xml:

Update of the following collection:
    - shared
    - default
    - vertebrata
    - mammalia
    - actinopterygii
    - sauropsida
Addition of the following collection
    - actinopterygii
    - sauropsida
    - shared_vertebrates 

2- update of the allowed_species.json

Use the script create_allowed_species.py to update the list of allowed species 
Note that the script is creating a new file each time so the name are not necessary int he same order than previously
This script should make sure that mlss_conf.xml and allowed_species.json have the same number of species and avoid risk of error caused by manual intervention

3- addition of create_allowed_species.py 
This script read the mlss_con.xml file and generate an allowed_species.json file with all the species present in mlss_con.xml

## Testing
create_allowed_species.py run and checked manually that all species were present

## Notes
NA


CristiGuijarro:
The reason for the failing checks here is that there are species included in the `mlss_conf.xml` that are not in the `allowed_species.json` - so I guess you need to put the missing species in the `allowed_species.json`.

conf/references/mlss_conf.xml
```diff
@@ -11,39 +11,44 @@
...
-      <genome name="bombyx_mori"/>
-      <genome name="capitella_teleta"/>
-      <genome name="nematostella_vectensis"/>
-      <genome name="adineta_vaga"/>
+      <genome name="daphnia_pulex"/>
+      <genome name="amphimedon_queenslandica"/>
+      <genome name="clytia_hemisphaerica"/>
+      <genome name="lottia_gigantea"/>
+      <genome name="trichoplax_adhaerens"/>
+      <genome name=" mnemiopsis_leidyi"/>
```
CristiGuijarro, line 44:
```suggestion
      <genome name="mnemiopsis_leidyi"/>
```

conf/references/mlss_conf.xml
```diff
@@ -11,39 +11,44 @@
...
-      <genome name="triticum_aestivum"/>
+      <genome name="oryza_sativa"/>
       <!-- Protists -->
       <genome name="plasmodium_falciparum"/>
     </collection>
 
+    <collection name ="shared_vertebrates">
+      <base_collection name="shared"/>
+      <genome name="mus_musculus"/>
+      <genome name="xenopus_tropicalis"/>
```
CristiGuijarro, line 26:
```suggestion
      <genome name="xenopus_tropicalis"/>
      <genome name="ciona_intestinalis"/>
```

conf/references/mlss_conf.xml
```diff
@@ -57,94 +62,194 @@
...
       <genome name="zymoseptoria_tritici"/>
       <genome name="erysiphe_necator_gca_000798715"/>
+      <genome name="schizosaccharomyces_pombe"/>
     </collection>
 
     <collection name="vertebrata">
-      <base_collection name="shared"/>
+      <base_collection name="shared_vertebrates"/>
+      <!-- outgroup -->
+      <genome name="ciona_intestinalis"/>
```
CristiGuijarro, line 71:
```suggestion
```
(Based on including it in the `shared_vertebrates` collection)

conf/references/mlss_conf.xml
```diff
@@ -57,94 +62,194 @@
...
+      <genome name="dipodomys_ordii"/>
+      <genome name="jaculus_jaculus"/>
+      <genome name="peromyscus_maniculatus_bairdii"/>
+      <genome name="Nannospalax_galili"/>
+    </collection>
+
+    <collection name="actinopterygii">
+      <base_collection name="shared_vertebrates"/>
+      <!-- Outgroups -->
+      <genome name="ciona_intestinalis"/>
```
CristiGuijarro, line 157:
```suggestion
```
As before

conf/references/mlss_conf.xml
```diff
@@ -57,94 +62,194 @@
...
+      <!-- 	Elopiformes -->
+      <genome name="megalops_cyprinoides"/>
+      <!-- Beloniformes -->
+      <genome name="oryzias_latipes"/>
+    </collection>
+
+    <collection name="sauropsida">
+      <base_collection name="shared_vertebrates"/>
+      <!-- Outgroup -->
+      <genome name="ciona_intestinalis"/>
```
CristiGuijarro, line 210:
```suggestion
```
As before

CristiGuijarro (commented):
As mentioned in comment, `allowed_species.json` needs to be amended to fit the reference species in these collections. Including _Ciona intestinalis_ in the `shared_vertebrates` is just a suggestion - because it seems an obvious constant outgroup to me that should be used in all vertebrates.

conf/references/mlss_conf.xml
dthybert: indeed, this is something I completely forgot to do. Thank you for reminding

conf/references/mlss_conf.xml
dthybert: ciona_intestinalis cannot be in the shared_vertebrates as it is not in the mammalia. The reason why it is not in the mammalia is because there is too many important species in the mammalia belonging to the a same subclade and so I preferred to remove a few outgroups rather than representative of some mammalian subclades.

conf/references/mlss_conf.xml
dthybert: cannot be removed because not in shared vertebrates

conf/references/mlss_conf.xml
dthybert: cannot be removed because not in shared vertebrates

conf/references/mlss_conf.xml
dthybert: cannot be removed because not in shared vertebrates

conf/references/mlss_conf.xml
CristiGuijarro: So  is 40 a maximum cap now and not an aim/goal? I think the use and importance of outgroups in reference species needs to be looked into further. Outgroups are essential to comparative analyses in evolutionary biology. In my experience, outgroups should equal or outweigh the clade of interest. Where that is not possible, the disparity in outgroups should be considered, for example, looking at different levels of outgroup relationship. In this case, we could consider the closely related outgroups to mammalia being non-mammalian vertebrates, and then we should hop further and look to non-vertebrate chordates. Ideally, every clade in vertebrates should be compared to at least one outgroup in each of the deuterostome phyla/subphyla - this might not be possible given a cap, but I leave this here.

conf/references/allowed_species.json
```diff
@@ -1,85 +1,121 @@
...
-    "paramecium_tetraurelia",
-    "petromyzon_marinus",
-    "phascolarctos_cinereus",
+    "eptatretus_burgeri",
+    "daphnia_pulex",
+    "amphimedon_queenslandica",
+    "clytia_hemisphaerica",
+    "lottia_gigantea",
+    "trichoplax_adhaerens",
+    " mnemiopsis_leidyi",
```
CristiGuijarro, line 22:
```suggestion
    "mnemiopsis_leidyi",
```

conf/references/mlss_conf.xml
dthybert: Yes the cap is important as it makes sure that we are able to process the query genome on a given time. I agree with you about the importance of outgroups, and I  see this will need more discussion to formerly define how we select outgroups, at the moment I was more focusing on the representative of the clade targeted, but yes we need to define a formal way to identify outgroups . In the example of vertebrates,  having a deuterostome for all vertebrates collections may no be always needed especially for collection that are deep into the clade (if we consider to make a primate collection or rodent one for instance).

scripts/references/create_allowed_species.py
```diff
@@ -0,0 +1,88 @@
...
+        print("issue with " + mlss_conf_file)
+        print(err)
+        sys.exit(1)
+
+    root = compara_db_tree.getroot()
+    genomes = root.findall("./collections/collection/genome")
+    species_dic = {}
+    for genome in genomes:
+        species_dic[genome.get('name')] = 1
+    species_list = list(species_dic.keys())
```
JAlvarezJarreta, line 49: Great approach! The use of `set` would have made this slightly faster, but even for thousands of species I think the performance boost would be negligible.

scripts/references/create_allowed_species.py
```diff
@@ -0,0 +1,88 @@
...
+
+    :param mlss_conf_file: path to the mlss_conf.xml
+    :param allowed_species_file: path to the output allowed_species.json file
+    :return: None
+    """
+
+    # get the list fo species from the mlss_conf.xml
+    species = get_species_list(mlss_conf_file)
+    # create the json file with the list of species
+    create_allowed_species_json(species, allowed_species_file)
```
JAlvarezJarreta, line 78:
```suggestion
    with open(allowed_species_file, "w") as file_handle:
        json.dump(species, file_handle, indent=4)
```
I don't think such a small code needs its own function.

scripts/references/create_allowed_species.py
```diff
@@ -0,0 +1,88 @@
...
+def main(mlss_conf_file: str, allowed_species_file: str) -> None:
+    """
+    Main function of the script
+
+    :param mlss_conf_file: path to the mlss_conf.xml
+    :param allowed_species_file: path to the output allowed_species.json file
+    :return: None
+    """
+
+    # get the list fo species from the mlss_conf.xml
```
JAlvarezJarreta, line 75:
```suggestion
```
The code is really readable, so the comment is not necessary.

scripts/references/create_allowed_species.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+    Create the json file with th list of allowed species
+
+    :param species: the list of species
+    :param allowed_species_file: path to the output json file
+    :return: None
+    """
+
+    with open(allowed_species_file, "w") as file_handle:
+        json.dump(species, file_handle, indent=4)
```
JAlvarezJarreta, line 63:
```suggestion
```
See comment below.

scripts/references/create_allowed_species.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+
+    # get the list fo species from the mlss_conf.xml
+    species = get_species_list(mlss_conf_file)
+    # create the json file with the list of species
+    create_allowed_species_json(species, allowed_species_file)
+
+
+if __name__ == '__main__':
+    # argument handling
```
JAlvarezJarreta, line 82:
```suggestion
```
Same as before.

scripts/references/create_allowed_species.py
```diff
@@ -0,0 +1,88 @@
...
+        print(err)
+        sys.exit(1)
+
+    root = compara_db_tree.getroot()
+    genomes = root.findall("./collections/collection/genome")
+    species_dic = {}
+    for genome in genomes:
+        species_dic[genome.get('name')] = 1
+    species_list = list(species_dic.keys())
+    return species_list
```
JAlvarezJarreta, line 50: I'm wondering if we should `sort` the list of species. This way, the number of differences between runs/commits will be much smaller and easier to track.

scripts/references/create_allowed_species.py
```diff
@@ -0,0 +1,88 @@
...
+
+def main(mlss_conf_file: str, allowed_species_file: str) -> None:
+    """
+    Main function of the script
+
+    :param mlss_conf_file: path to the mlss_conf.xml
+    :param allowed_species_file: path to the output allowed_species.json file
+    :return: None
+    """
+
```
JAlvarezJarreta, line 74:
```suggestion
def process_mlss_conf_file(mlss_conf_file: str, allowed_species_file: str) -> None:
    """Gets the list of species in ``mlss_conf_file`` and writes them in ``allowed_species_file`` as a JSON list.

    Args:
        mlss_conf_file: Path to the MLSS XML configuration file.
        allowed_species_file: Path to the allowed species JSON file.

    Raises:
        FileNotFoundError: If ``allowed_species_file`` cannot be created (wrong path).

    """
```
Same reason as before for the docstring. I have renamed the method because `main` is quite uninformative.

scripts/references/create_allowed_species.py
```diff
@@ -0,0 +1,88 @@
...
+
+if __name__ == '__main__':
+    # argument handling
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--mlss_conf', help="path to mlss_conf.xml file given as input")
+    parser.add_argument('--allowed', help="path to the allowed_species.json given as output")
+    args = parser.parse_args()
+
+    main(args.mlss_conf, args.allowed)
```
JAlvarezJarreta, line 88:
```suggestion
    process_mlss_conf_file(args.mlss_conf, args.allowed)
```

scripts/references/create_allowed_species.py
```diff
@@ -0,0 +1,88 @@
...
+    species = get_species_list(mlss_conf_file)
+    # create the json file with the list of species
+    create_allowed_species_json(species, allowed_species_file)
+
+
+if __name__ == '__main__':
+    # argument handling
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--mlss_conf', help="path to mlss_conf.xml file given as input")
+    parser.add_argument('--allowed', help="path to the allowed_species.json given as output")
```
JAlvarezJarreta, line 85:
```suggestion
    parser = argparse.ArgumentParser(description="Creates the JSON file with the list of allowed species from a MLSS XML configuration file.")
    parser.add_argument('--mlss_conf', help="Path to the MLSS XML configuration file.")
    parser.add_argument('--allowed', help="Path to the allowed species JSON file.")
```

JAlvarezJarreta (commented):
The changes regarding the Google style for docstrings and the comment about sorting the species list are the only two really relevant. The others are just suggestions to improve the code, but it is good as it is.

scripts/references/create_allowed_species.py
```diff
@@ -0,0 +1,88 @@
...
+from typing import List
+
+
+def get_species_list(mlss_conf_file: str) -> List:
+    """
+    Return a list with all species present in the mlss_conf.xml. Each species are unci in the list
+
+    :param mlss_conf_file: path to the mlss_conf.xml
+    :return: list of species
+    """
```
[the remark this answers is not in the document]
dthybert: Yes I use the default format defined by pycharm. I will update this.

scripts/references/create_allowed_species.py
```diff
@@ -0,0 +1,88 @@
...
+
+    try:
+        compara_db_tree = xml.etree.ElementTree.parse(mlss_conf_file)
+    except ParseError as err:
+        print("issue with " + mlss_conf_file)
+        print(err)
+        sys.exit(1)
+
```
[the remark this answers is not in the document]
dthybert: Yes I red this, as well but I did not know what the message would be. I tend to catch exception in case like this as we often do not know what the error message will be. But I guess the script will die anyway if there is an error here so not need to catch it and it will make the code simpler

CristiGuijarro (approved):
Approved on the basis that @JAlvarezJarreta comments/suggestions are addressed

scripts/references/create_allowed_species.py
dthybert: I always name the main function of my script main, it is a way to encapsulated the higher level of the script logic into a function If I had name this function if would have been the name of the script as the function is not only processing mlss conf file but also creating the allowed species file.
The advantage to have principal function named main  is that we know without reading the whole script where is written the main logic of the script. I would prefer keeping that way.

scripts/references/create_allowed_species.py
dthybert: yes in this case you are right `set` would have been better It makes more sense to use it.

JAlvarezJarreta (approved):
Just a couple of typos to correct (and apply the suggestions if you want as well). Could you please use the script to generate `allowed_species.json` again with the species sorted? That way we will merge a "clean" file.
Once that is done, this is good to go :heavy_check_mark:

scripts/references/create_allowed_species.py
```diff
@@ -0,0 +1,99 @@
...
+    # get the list fo species from the mlss_conf.xml
+    species = get_species_list(mlss_conf_file)
+    # create the json file with the list of species
+    create_allowed_species_json(species, allowed_species_file)
+
+
+if __name__ == '__main__':
+    # argument handling
+    parser = argparse.ArgumentParser(description="Creates the JSON file with the list of allowed species from"
+                                                 " a MLSS XML configuration file.")
```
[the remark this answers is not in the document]
dthybert: It is the first time I hear about surrounding by brackets I always go back the next line. The script did work.

scripts/references/create_allowed_species.py
```diff
@@ -0,0 +1,99 @@
...
+
+def get_species_list(mlss_conf_file: str) -> List:
+    """Returns a list with all species present in the mlss_conf.xml. Each species will appear only once
+
+    Args:
+        mlss_conf_file: Path to the mlss_conf.xml
+
+    Returns:
+        List of species
+
```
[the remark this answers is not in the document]
dthybert: Yes I was not sure why in your change you did not write this. Will this give an issue if we generate doc? if not I prefer to keep it because it is more readable that way

scripts/references/create_allowed_species.py
```diff
@@ -0,0 +1,99 @@
...
+def create_allowed_species_json(species: List, allowed_species_file: str) -> None:
+    """Create the json file with th list of allowed species
+
+    Args:
+        species: The list of species
+        allowed_species_file: Path to the output json file
+
+    Returns:
+        None
+
```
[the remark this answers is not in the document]
dthybert: ok, so the Returns is not systematic

dthybert:
> Just a couple of typos to correct (and apply the suggestions if you want as well). Could you please use the script to generate `allowed_species.json` again with the species sorted? That way we will merge a "clean" file.
> Once that is done, this is good to go ✔️

PR #267, opened 2021-03-17
Add rename reference genome

JAlvarezJarreta:
## Description

When a reference species is renamed it is reported by metadata as such, so it needs to be handled separately. In our case, that means: if the reference was already there, rename it and all its dumps files; if it wasn't there, add it as new.

**Related JIRA tickets:**
- ENSCOMPARASW-4381

#### Change 1
- Added method `rename_reference_genome()` in `ReferenceDatabase.pm`, following the same ideas as in `update_reference_genome`. I have also added some tests in `Reference_Database.t`.

#### Change 2
- Created the runnable `RenameReferenceGenome.pm` and add it to `UpdateReferenceDatabase` pipeline. This runnable not only calls the previous method, but also updates the configuration files (if they haven't been updated beforehand) and all the reference dumps (named in some places member dumps).

#### Change 3
- Added new logic to `UpdateGenomesFromMetadataFactory.pm` to take into account outdated config files (MLSSs XML and allowed species), and to add (instead of rename) a new reference that is reported as renamed from metadata when it is not present in the database.

## Testing
Travis has done its part with the newly added test. Besides that, I have run the pipeline twice, one in e102 [[pipeline](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-2&port=4522&dbname=jalvarez_update_references_102)] and another one in e104 [[pipeline](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-2&port=4522&dbname=jalvarez_update_references_104)] with `physcomitrium_patens`, that was renamed in e104. Furthermore, I tested a load from scratch from e104 with this species in [this pipeline](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-2&port=4522&dbname=jalvarez_update_references_rename_104_bis), and it was correctly added.

## Notes
- Hope you like the rename in the test!


CristiGuijarro (approved):
Easy to follow and elegantly implemented :+1: 
(Approved when Travis issues fixed for fake mouse core test db)

CristiGuijarro (approved):
Thanks Travis :+1:

PR #271, opened 2021-03-26
Do not store genome components of reference genomes by default

JAlvarezJarreta:
## Description

Added new flag to store the genome components (`store_components`) that defaults to 0.

## Testing
I'm running the reference pipeline again to test it works: [jalvarez_update_references_103_fresh](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-3&port=4523&dbname=jalvarez_update_references_103_fresh). All jobs in `update_reference_genome` have finished and the genome components are not in the `genome_db` table.

## Notes
- The idea of this PR is just to share the approach I have opted for to see if you would prefer something slightly different.
- In the pipeline the flag is named `no_components`, but I have thought this new name is more readable.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/ReferenceGenomes/UpdateReferenceGenome.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/ReferenceGenomes/UpdateReferenceGenome.pm
@@ -1,4 +1,3 @@
-
 =head1 LICENSE
 
 See the NOTICE file distributed with this work for additional information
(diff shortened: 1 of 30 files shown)


CristiGuijarro (commented):
Sure, that seems to make sense to me - at a glance, how was it wired with the `no_components` flag before? I think this flag should be included in the `ReferenceDatabase.t` - but possibly you were going to do that anyway following approval of this shared approach?

CristiGuijarro (approved):

PR #273, opened 2021-04-01
GOC score stats & plots

ipilizota:
[In progress]
## Description

- Getting average and median GOC score for the whole dataset. 
- Getting average and median GOC score for each pair of species & plotting corresponding heatmaps.

**Related JIRA tickets:**
- ENSCOMPARASW-4178

## Overview of changes
- In `generateGocBreakout.pl`: added new code, made `$sql1` compatible with the current schema, edited SQL formatting to increase readability. 
- In `plotGocData.r`: edited existing code to make it work with new output files from `generateGocBreakout.pl`.

#### Change 1
Added to `generateGocBreakout.pl`: 
- `$sql_pairwise_stats` to calculate the average GOC score for each pair of species and get values to compute the median GOC score for each pair of species. Perl code to compute medians and write them together with averages to `heatmap_avg_median.data`.
- `$sql_goc_avg` to calculate the average GOC score for the whole dataset. `$sql_goc_median` to get values to compute the median GOC score for the whole dataset. Perl code to calculate the median. Perl code to write the average and median to `dataset_stats.txt`.

#### Change 2
Capitalised SQL keywords and broke queries into multiple lines to increase readability. Also formatted some comments throughout the script.

#### Change 3
Not required for this Jira ticket but to keep the functionality from the past:
- Modified `sql1` so it is compatible with the current schema.
- Added `sub match_species_name_to_taxon_list` so the species -> taxon assignment does not have to be done manually (`@taxon_list` might have to be updated depending on the dataset under investigation, tho). Updated `%colors` accordingly.
- In `$sql_heatmap`:
    - Abandoned using NCBI scientific names and went with `genome_db.name` because different strains and hybrid species were ending up with the same name in `heatmap.data` . That meant multiple rows with the same `(name1, name2)` pairs which caused errors in `plotGocData.r` when calling `heatmap.phylo` function for plotting.
    - Excluded `sus_scrofa_%`.

#### Change 4
Added to `plotGocData.r`:
- In `heatmap.phylo` function: 
    - Added `legend_text` parameter (used to be fixed) to make it fit for the purpose of this ticket. 
    - Minor changes in figure and text sizes to make it nice and readable for bigger datasets (e.g. 200 species like this time round).
- Reading in `heatmap_avg_median.data` and plotting average & median GOC score heatmaps (`goc_avg.pdf`, `goc_median.pdf`). This also included defining colors and legend texts accordingly. In addition, 3 lines of code were added to accommodate for cases when the input species tree contains extra species not present in the dataset (as happened with the dataset under investigation).

#### Change 5
Not required for this Jira ticket:
- Updated taxon colours in `list` so they match those in `generateGocBreakout.pl`.

#### Before the PR
- Only checked SQL queries by querying a database and observing something that made sense. 

#### While addressing PR review

**`generateGocBreakout.pl`**
- `sub match_species_name_to_taxon_list` : checked manually for XXX, XXX, XXX
- `dataset_stats.txt` and `heatmap_avg_median.data` look sensible to me. I had a look at two cases where `| goc_avg -goc_median | > 25` but it made sense when I saw `n_goc_*`.
- Please see examples of new output files (`farm`):
    - `/homes/ivana/goc_new/dataset_stats.txt`
    - `/homes/ivana/goc_new/heatmap_avg_median.data`

**`plotGocData.r`**

- Just checked that heatmaps generally make sense given the data being plotted and phylogenetic distances between species.

- `heatmap.phylo` throws:
```
Warning messages:
1: In plot.window(...) : "fill" is not a graphical parameter
2: In plot.window(...) : "border" is not a graphical parameter
3: In plot.xy(xy, type, ...) : "fill" is not a graphical parameter
4: In plot.xy(xy, type, ...) : "border" is not a graphical parameter
5: In title(...) : "fill" is not a graphical parameter
6: In title(...) : "border" is not a graphical parameter
```
That's ok given the parameters in the `heatmap.phylo` calls and functions within `heatmap.phylo` that do or do not use them.

- Please see examples of new output files (`farm`):
    - `/homes/ivana/plot_new/goc_avg.pdf`
    - `/homes/ivana/plot_new/goc_median.pdf`

## Notes
I did not do much testing on the `generateGocBreakout.pl` and `plotGocData.r` output files which are beyond the scope of this ticket. In addition, the original `generateGocBreakout.pl` was generating some empty files (remember the schema change) which are now not empty. Then `plotGocData.r` had no input to plot. The point is there are no reference files for comparison. Nevertheless, here are my observations which could be useful for creating future tasks and tickets.

#### `generateGocBreakout.pl`

**Output files that had some content before my changes (new vs. old output):** 
- `heatmap.data`
	- Checked for random `(name1, name2)` not affected by move to `genome_db.name` -> looked ok
	- Checked for some affected by move to `genome_db.name` (`oryzias_latipes_*`) -> looked ok
	- Some new appeared (as now `genome_db.name1 != genome_db.name2`) -> no point of reference but it looked ok for e.g. `(oryzias_latipes_hni, oryzias_latipes_hsok)`; the number of new lines is as expected for this particular dataset
- `homology.data`
	- The same
- `gene_count.data`
	- The same    

**Files that are not empty anymore:**
- `homo_sapiens_above_with_splits.dat, homo_sapiens_ref_above_threshold.dat, homo_sapiens_ref.dat`
- `sphenodon_punctatus_above_with_splits.dat, sphenodon_punctatus_ref_above_threshold.dat, sphenodon_punctatus_ref.dat`
Should be checked if they contain the info that was intended to be there, especially in the light of error(s) in `plotGocData.r`.

#### `plotGocData.r`

**Plots that look ok:**
- `gene_count.pdf`
- `goc_species_name.pdf` (Should be edited to look nicer; the initial code was probably used on smaller datasets)

**Plots that don't look ok:**
- `number_of_orthologues.pdf`: No points. I think the problem is that the header in the input file has 6 columns and the data is actually spread over 7 columns. Probably changes in schema or information that is now stored in the db.
- `ordered_goc_100_references.pdf`: Blank file.
	Input: *_ref.dat

**Plots that are not generated:**
- `above_with_splits_references.pdf`. Input: `*_above_with_splits.dat`. I think this is the corresponding error:
```
Error: Continuous value supplied to discrete scale
Execution halted
```
To me it looks like the input file contains 2 lines for each `(reference_species, another_species)` : 1 with `% orthologs` having GOC score above the threshold and 1 with  `% orthologs` having GOC score below the threshold. It perhaps expects the number of orthologs or sum of GOC scores instead.
- `above_threshold_references.pdf`. No surprise; comes after `above_with_splits_references.pdf`.


carlacummins (commented):
A few general comments:

1. `n_goc_null` was technically removed, but only because we moved the default value from `NULL` to `0`, so `n_goc_0` is the new equivalent. `$sql1` is designed to pick out all the MLSSes that have GOC data (as we only compare certain pairs of species) - is this taken into account in your script? It looks like you are just converting `NULL`s to `0` (`ISNULL(n_goc_25, 0)`), so you are likely to end up with a strange pattern of missing data in the resulting heatmap. Did you notice that in your output?
2. `mus_musculus_%` was excluded because that is the naming pattern of the mouse strains. If this script will only be run on pipeline databases, then this is not a problem. But if you run it on a release database, this will be a problem. There are also pig breeds to consider now too (`sus_scrofa_%`). It may be worth restricting the analysis to a species set (can default to `collection-default`)
3. Most of what you've removed relates to running this analysis on groups of species/clades - are we sure we won't want to do that in the future? You're definitely happy that this functionality will be lost and the analysis will only ever be run on pairs of species?
4. Are you sure the outputs you removed from `plotGocData.r` are not useful at all?

Overall, I would say that these changes are essentially a completely new script, but the filenames have been reused because Mateus' original script was overly specific to his purpose at the time of writing. Some logic has been reused and a lot has been removed. If this really was the intention, then I think the changes are fine.

CristiGuijarro:
Will review soon, the `n_goc_null` removal and not replacement was likely my poor advice, sorry about that :+1:

ipilizota:
Thank you very much for your feedback @carlacummins. Very valid points! I will address them so I think @CristiGuijarro should review the PR only afterwards.

I also feel like this is turning into a new script rather than a modification of an old one. I suggest to either
- Keep the old name but change `n_goc_null` to `n_goc_0`, don't remove anything, add new functionality or
- Make it a new script having only parts relevant for the Gene tree benchmarking project.

I'm not sure what to do with the R script. But it might be more clear when I incorporate your feedback in `*.pl` script.

CristiGuijarro:
I don't think `n_goc_null` can be replaced by `n_goc_0` exactly. `n_goc_null` was used when not all mlsss were considered for high confidence. Since all mlsss are now considered for high confidence, which would leave  we introduced the `goc_expected` tag - which is stored afterwards - although this is not useful for e101 because it was not in place until e102/e103. `n_goc_0` expects the number of orthologs without gene order conservation. So it is likely that these columns - whilst allowed to be `NULL` should never be `NULL` and will otherwise have a number >=0. An alternative could be something like: `SELECT method_link_species_set_id FROM method_link_species_set_attr where perc_orth_above_goc_thresh !=0;`

ipilizota:
Thank you @CristiGuijarro for the explanation. As of taxonomic/clade thing, I realised it was used in regard to `n_goc_null` only.

PR #274, opened 2021-04-01
Improve performance of HomologyAnnotation pipeline

JAlvarezJarreta:
## Description

Improve the performance of homology annotation pipeline. As a side effect, the guiHive shows a much cleaner pipeline now.

**Related JIRA tickets:**
- ENSCOMPARASW-4399

#### Change 1
- When the references are added to `genome_db` table, `AUTO_INCREMENT` is set after the latest `genome_db_id`, so the pipeline can no longer be topped up without adding extra analyses, so I have removed that experimental feature.

#### Change 2
- Moved the copy of reference genomes to the beginning, and all the references are copied (it doesn't take much longer anyway and it may be useful to have them all in the future). It is now done at the same time the members of query genomes are being loaded, so no additional time is added to the pipeline.

#### Change 3
- Removed `load_query_genomedb_factory` and `load_genomedb` analyses as these two tasks were already being taken care of by `locate_and_add_genomes`.

#### Change 4
- Put back 3 indexes and 2 foreign keys for `peptide_align_feature` table to add robustness and improve performance when querying this table.

#### Minor changes
- I have removed `ref_taxa` in `diamond_blastp` because it was not being used.
- `mlss_adaptor->store()` already takes care of storing the species set and method if they are not present in the database, so I have removed the unnecessary `store()`.

## Testing
[jalvarez_homology_annotation_halibut_103_alt](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-3&port=4523&dbname=jalvarez_homology_annotation_halibut_103_alt)

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/Parts/LoadCoreMembers.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/Parts/LoadCoreMembers.pm
@@ -76,6 +76,7 @@ sub pipeline_analyses_copy_ncbi_and_core_genome_db {
                     { 'load_fresh_members_from_db' => { 'genome_db_id' => '#genome_db_id#' } },
                 ],
                 'A->1' => [ 'hc_members_globally' ],
+                '1'    => [ 'copy_ref_genomes_factory' ],
             },
         },
     #--------------------Query genome member loading------------------#
@@ -116,6 +117,28 @@ sub pipeline_analyses_copy_ncbi_and_core_genome_db {
             -module             => 'Bio::EnsEMBL::Compara::RunnableDB::GeneTrees::SqlHealthChecks',
             -parameters         => {
                 mode   => 'members_globally',
+            },
+        },
+    #--------------------Reference genome loading------------------#
+        {   -logic_name => 'copy_ref_genomes_factory',
+            -module     => 'Bio::EnsEMBL::Compara::RunnableDB::GenomeDBFactory',
+            -parameters => {
+                'compara_db'    => '#rr_ref_db#',
+                'all_current'   => 1,
+            },
+            -flow_into  => {
+                2 => [ 'copy_ref_genomes' ],
+            },
+        },
+
+        {   -logic_name => 'copy_ref_genomes',
+            -module     => 'Bio::EnsEMBL::Hive::RunnableDB::MySQLTransfer',
+            -parameters => {
+                'src_db_conn'   => '#rr_ref_db#',
+                'mode'          => 'insertignore',
+                'filter_cmd'    => 'sed "s/ENGINE=MyISAM/ENGINE=InnoDB/"',
+                'table'         => 'genome_db',
+                'where'         => 'genome_db_id = #genome_db_id#',
             },
         },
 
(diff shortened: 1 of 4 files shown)


modules/Bio/EnsEMBL/Compara/PipeConfig/Parts/LoadCoreMembers.pm
```diff
@@ -145,6 +119,28 @@ sub pipeline_analyses_copy_ncbi_and_core_genome_db {
                 mode   => 'members_globally',
             },
         },
+    #--------------------Reference genome loading------------------#
+        {   -logic_name => 'copy_ref_genomes_factory',
+            -module     => 'Bio::EnsEMBL::Compara::RunnableDB::GenomeDBFactory',
+            -parameters => {
+                'compara_db'    => '#rr_ref_db#',
+                'all_current'   => 1,
```
CristiGuijarro, line 127: Something to consider later on rather than now is that a RR version will be used and the `software_version()` will be redundant, this may or may not have an impact on the `compara_references` - but non-current are unlikely to be kept in the db - more release style rather then `master`. Another future option could be to take the `genome_db` names from `references/allowed_species.json` file and copy these across instead.

sql/table.sql
```diff
@@ -1405,13 +1405,13 @@ CREATE TABLE peptide_align_feature (
...
 
-  PRIMARY KEY (peptide_align_feature_id)
+  PRIMARY KEY (peptide_align_feature_id),
 #  KEY hmember_hit (hmember_id, hit_rank)
 
-#  KEY qmember_id  (qmember_id),
-#  KEY hmember_id  (hmember_id),
+  KEY qmember_id  (qmember_id),
+  KEY hmember_id  (hmember_id),
 #  KEY hmember_qgenome  (hmember_id, qgenome_db_id),
```
CristiGuijarro, line 1413:
```suggestion
```
May as well delete this :shrug:

sql/table.sql
```diff
@@ -1405,13 +1405,13 @@ CREATE TABLE peptide_align_feature (
   FOREIGN KEY (qgenome_db_id) REFERENCES genome_db(genome_db_id),
   FOREIGN KEY (hgenome_db_id) REFERENCES genome_db(genome_db_id),
 
-  PRIMARY KEY (peptide_align_feature_id)
+  PRIMARY KEY (peptide_align_feature_id),
 #  KEY hmember_hit (hmember_id, hit_rank)
```
CristiGuijarro, line 1409:
```suggestion
```
May as well delete this :shrug:

CristiGuijarro (approved):
There should have been some other changes linked to these ones, but they have already been pushed. Approving because I already checked those prior commits out to be sure and all appears to be in working order. The pipeline is prettier now :+1:

CristiGuijarro:
Something else I nearly forgot, but the `locator` in each `genome_db` entry has now been removed in this PR - at some point this may need to go back, and included in the reference `genome_db` entries too.

modules/Bio/EnsEMBL/Compara/PipeConfig/Parts/LoadCoreMembers.pm
JAlvarezJarreta: I think the best approach is to keep a clean reference database so there is no need to add such logic to this pipeline, but these are good thoughts to keep into account indeed.

PR #279, opened 2021-04-15
Dump and delete wga_expected tags

ipilizota:
## Description

In e104, the `wga_expected` tags from protein and ncRNA trees clashed and caused issues in `MergeDBIntoRelease` pipeline. To prevent this in the future, I added an analysis at the end of `ProteinTrees` and `ncRNAtrees` pipelines which dumps the `wga_expected` tags and deletes them from corresponding pipeline databases.

**Related JIRA tickets:**
- ENSCOMPARASW-4295

## Overview of changes
- Created a runnable `DumpWGAExpectedTags.pm`.
- Modified `ProteinTrees_conf.pm` and `ncRNAtrees_conf.pm` accordingly.
- Wrote unit tests `DumpWGAExpected.t` and added some db entries for testing to `method_link_species_set_tag.txt`.

#### Change 1
- `DumpWGAExpectedTags.pm` is supposed to dump the `wga_expected` tags into a file (`.../dumps/wga_expected.mlss_tags.tsv`) and delete the tags from the database.

#### Change 2
- `ProteinTrees_conf.pm` and `ncRNAtrees_conf.pm` have been updated to include a new `wga_expected_dumps` analysis at their ends. 

## Testing
- **Unit tests:** `DumpWGAExpectedTags.t` loads some `wga_expected` and non-`wga_expected` tags from `method_link_species_set_tag.txt`, checks if the output file is correct and checks if the number of entries in the `method_link_species_set_tag` table is correct after deleting `wga_expected` tags.
- **Integration tests:** Tested on a copy of e104 [Plants ProteinTrees](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensadmin&host=mysql-ens-compara-prod-7&port=4617&dbname=ivana_copy_plants_protein_trees_104&passwd=xxxxx) and [Vertebrates ncRNAtrees](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensadmin&host=mysql-ens-compara-prod-7&port=4617&dbname=ivana_copy_vertebrates_ncrna_trees_104&passwd=xxxxx) db. Since the `wga_expected` tags had already been deleted from the Vertebrates db, I added some entries for testing. In both cases, `wga_expected.mlss_tags.tsv` was correct and no `wga_expected` tags were in the `mlss_tag` table. 

## Notes
[...]

PR #280, opened 2021-04-19
Dump & delete wga_expected tags

ipilizota:
## Description

In e104, the `wga_expected` tags from protein and ncRNA trees clashed and caused issues in `MergeDBIntoRelease` pipeline. To prevent this in the future, I added an analysis at the end of `ProteinTrees` and `ncRNAtrees` pipelines which dumps the `wga_expected` tags and deletes them from corresponding pipeline databases.

**Related JIRA tickets:**
- ENSCOMPARASW-4295

## Overview of changes
- Created a runnable `DumpWGAExpectedTags.pm`.
- Modified `ProteinTrees_conf.pm` and `ncRNAtrees_conf.pm` accordingly.
- Wrote unit tests `DumpWGAExpected.t` and added some db entries for testing to `method_link_species_set_tag.txt`.

#### Change 1
- `DumpWGAExpectedTags.pm` is supposed to dump the `wga_expected` tags into a file (`.../dumps/wga_expected.mlss_tags.tsv`) and delete the tags from the database.

#### Change 2
- `ProteinTrees_conf.pm` and `ncRNAtrees_conf.pm` have been updated to include a new `wga_expected_dumps` analysis at their ends. 

## Testing
- **Unit tests:** `DumpWGAExpectedTags.t` loads some `wga_expected` and non-`wga_expected` tags from `method_link_species_set_tag.txt`, checks if the output file is correct and checks if the number of entries in the `method_link_species_set_tag` table is correct after deleting `wga_expected` tags.
- **Integration tests:** Tested on a copy of e104 [Plants ProteinTrees](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensadmin&host=mysql-ens-compara-prod-7&port=4617&dbname=ivana_copy_plants_protein_trees_104&passwd=xxxxx) and [Vertebrates ncRNAtrees](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensadmin&host=mysql-ens-compara-prod-7&port=4617&dbname=ivana_copy_vertebrates_ncrna_trees_104&passwd=xxxxx) db. Since the `wga_expected` tags had already been deleted from the Vertebrates db, I added some entries for testing. In both cases, `wga_expected.mlss_tags.tsv` was correct and no `wga_expected` tags were in the `mlss_tag` table. 

## Notes
- As I'm writing this, I realised that `notify_pipeline_completed` (ncRNA) might send an email before `wga_expected_dumps` finishes. I'm not sure if this is of particular concern. I'd say that `copy_dumps_to_shared_loc` might still be running as well.
- This is an updated PR of the #279. (I messed up with `commit` and `rebase` so at one point #279 included unrelated commits from other PRs.) This updated version encompasses commits and  suggestions by J. Alvarez-Jarreta related to his review of #279.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/DumpWGAExpectedTags.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/DumpWGAExpectedTags.pm
@@ -1,0 +1,71 @@
+=head1 LICENSE
+
+See the NOTICE file distributed with this work for additional information
+regarding copyright ownership.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=head1 NAME
+
+Bio::EnsEMBL::Compara::RunnableDB::GeneTrees::DumpWGAExpectedTags
+
+=head1 DESCRIPTION
+
+A runnable to dump wga_expected tag and delete it from a database.
+
+=cut
+
+package Bio::EnsEMBL::Compara::RunnableDB::GeneTrees::DumpWGAExpectedTags;
+
+use strict;
+use warnings;
+
+use base ('Bio::EnsEMBL::Compara::RunnableDB::BaseRunnable');
+
+
+sub run {
+    my ($self) = shift @_;
+   
+    $self->dump_wga_expected();
+
+    $self->delete_wga_expected();
+}
+
+
+sub dump_wga_expected {
+    my ($self) = shift @_;
+
+    my $dump_file = $self->param_required('wga_expected_file');
+    open( my $fh_out, '>', $dump_file ) || die "Could not open output file $dump_file";
+    print $fh_out "method_link_species_set_id\twga_expected\n";
+
+    my $sql = "SELECT method_link_species_set_id, value FROM method_link_species_set_tag WHERE tag = 'wga_expected'";
+    my $sth = $self->compara_dba->dbc->prepare($sql);
+    $sth->execute();
+    while (my @row = $sth->fetchrow_array()) {
+        print "Dumping mlss_id " . $row[0] . " wga_expected tag into $dump_file\n" if $self->debug;
+        print $fh_out join("\t", @row) . "\n";
+    }
+    $sth->finish;
+    close($fh_out);
+}
+
+sub delete_wga_expected {
+	my ($self) = shift @_;
+
+	print "Deleting wga_expected tag\n" if $self->debug;
+	my $sql = "DELETE FROM method_link_species_set_tag WHERE tag = 'wga_expected'";
+	$self->compara_dba->dbc->do($sql);
+}
+
+1;
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/DumpWGAExpectedTags.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/DumpWGAExpectedTags.pm
@@ -21,7 +21,7 @@ Bio::EnsEMBL::Compara::RunnableDB::GeneTrees::DumpWGAExpectedTags
 
 =head1 DESCRIPTION
 
-A runnable to dump wga_expected tag and delete it from a database.
+Dumps 'wga_expected' tags to the specified TSV file and deletes them from the database.
 
 =cut
 
@@ -34,7 +34,7 @@ use base ('Bio::EnsEMBL::Compara::RunnableDB::BaseRunnable');
 
 
 sub run {
-    my ($self) = shift @_;
+    my $self = shift;
    
     $self->dump_wga_expected();
 
@@ -43,7 +43,7 @@ sub run {
 
 
 sub dump_wga_expected {
-    my ($self) = shift @_;
+    my $self = shift;
 
     my $dump_file = $self->param_required('wga_expected_file');
     open( my $fh_out, '>', $dump_file ) || die "Could not open output file $dump_file";
@@ -61,7 +61,7 @@ sub dump_wga_expected {
 }
 
 sub delete_wga_expected {
-	my ($self) = shift @_;
+	my $self = shift;
 
 	print "Deleting wga_expected tag\n" if $self->debug;
 	my $sql = "DELETE FROM method_link_species_set_tag WHERE tag = 'wga_expected'";
--- a/modules/t/DumpWGAExpectedTags.t
+++ b/modules/t/DumpWGAExpectedTags.t
@@ -1,0 +1,82 @@
+#!/usr/bin/env perl
+# See the NOTICE file distributed with this work for additional information
+# regarding copyright ownership.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+use strict;
+use warnings;
+
+use Test::More;
+
+use Bio::EnsEMBL::Test::MultiTestDB;
+use Bio::EnsEMBL::Hive::DBSQL::DBConnection;
+use Bio::EnsEMBL::Hive::Utils::Test qw(standaloneJob);
+
+use File::Temp qw/tempfile/;
+
+# Check module can be seen and compiled
+use_ok('Bio::EnsEMBL::Compara::RunnableDB::GeneTrees::DumpWGAExpectedTags');
+
+# Load test DB
+my $multi_db = Bio::EnsEMBL::Test::MultiTestDB->new( "test_master" );
+my $compara_dba = $multi_db->get_DBAdaptor( "compara" );
+my $dbc = Bio::EnsEMBL::Hive::DBSQL::DBConnection->new(-dbconn => $compara_dba->dbc);
+my $compara_db = $dbc->url;
+
+# Number of tags before calling the module
+my $sql1 = "SELECT COUNT(*) AS total_rows FROM method_link_species_set_tag";
+my $db_vals1 = $dbc->db_handle->selectrow_hashref($sql1);
+my $all_count = $db_vals1->{total_rows};
+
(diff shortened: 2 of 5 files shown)


modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/DumpWGAExpectedTags.pm
```diff
@@ -0,0 +1,71 @@
...
+
+use base ('Bio::EnsEMBL::Compara::RunnableDB::BaseRunnable');
+
+
+sub run {
+    my ($self) = shift @_;
+   
+    $self->dump_wga_expected();
+
+    $self->delete_wga_expected();
```
JAlvarezJarreta, line 41: Just a reminder of my comment in the other PR: these two functions are really small so you could move their code in here (less lines of code, easier to follow and read in one go).

modules/t/DumpWGAExpectedTags.t
```diff
@@ -0,0 +1,82 @@
...
+# Number of tags before calling the module
+my $sql1 = "SELECT COUNT(*) AS total_rows FROM method_link_species_set_tag";
+my $db_vals1 = $dbc->db_handle->selectrow_hashref($sql1);
+my $all_count = $db_vals1->{total_rows};
+
+my $sql2 = "SELECT COUNT(*) AS wga_exp_rows FROM method_link_species_set_tag WHERE tag = 'wga_expected'";
+my $db_vals2 = $dbc->db_handle->selectrow_hashref($sql2);
+my $we_count = $db_vals2->{wga_exp_rows};
+
+my ($wefh, $test_wga_expected_file) = tempfile();
```
JAlvarezJarreta, line 46:
```suggestion
my ($wga_exp_fh, $test_wga_expected_file) = tempfile();
```

modules/t/DumpWGAExpectedTags.t
```diff
@@ -0,0 +1,82 @@
...
+    "method_link_species_set_id\twga_expected\n",
+    "101\t1\n",
+    "102\t0\n",
+    "104\t0\n",
+    "105\t1\n",
+    "106\t1\n",
+];
+
+open( my $fh, '<', $test_wga_expected_file) or die "Cannot open $test_wga_expected_file for reading\n";
+my @got_output = <$fh>;
```
JAlvarezJarreta, line 66: I'm not 100% sure, but cannot you use `$wga_exp_fh` here? The file handler is open to read but hasn't started reading, so it should work as expected.
Alternatively, you could use `Bio::EnsEMBL::Utils::IO::slurp()` to read the contents into a reference variable. It has some additional checks that can be useful.

modules/t/DumpWGAExpectedTags.t
```diff
@@ -0,0 +1,82 @@
...
+
+is_deeply( \@got_output, $exp_output, 'wga_expected tags dumped correctly to a file' );
+
+# Check that wga_expected tags are deleted from the database and no others
```
JAlvarezJarreta, line 70:
```suggestion
# Check that only wga_expected tags are deleted from the database
```

modules/t/DumpWGAExpectedTags.t
```diff
@@ -0,0 +1,82 @@
...
+# Load test DB
+my $multi_db = Bio::EnsEMBL::Test::MultiTestDB->new( "test_master" );
+my $compara_dba = $multi_db->get_DBAdaptor( "compara" );
+my $dbc = Bio::EnsEMBL::Hive::DBSQL::DBConnection->new(-dbconn => $compara_dba->dbc);
+my $compara_db = $dbc->url;
+
+# Number of tags before calling the module
+my $sql1 = "SELECT COUNT(*) AS total_rows FROM method_link_species_set_tag";
+my $db_vals1 = $dbc->db_handle->selectrow_hashref($sql1);
+my $all_count = $db_vals1->{total_rows};
```
JAlvarezJarreta, line 40:
```suggestion
my $num_tags_query = "SELECT COUNT(*) FROM method_link_species_set_tag";
my ($num_tags_before) = $dbc->db_handle->selectrow_array($num_tags_query);
```

modules/t/DumpWGAExpectedTags.t
```diff
@@ -0,0 +1,82 @@
...
+
+my $sql2 = "SELECT COUNT(*) AS wga_exp_rows FROM method_link_species_set_tag WHERE tag = 'wga_expected'";
+my $db_vals2 = $dbc->db_handle->selectrow_hashref($sql2);
+my $we_count = $db_vals2->{wga_exp_rows};
```
JAlvarezJarreta, line 44:
```suggestion
my $num_wga_exp_query = "SELECT COUNT(*) FROM method_link_species_set_tag WHERE tag = 'wga_expected'";
my ($num_wga_exp_before) = $dbc->db_handle->selectrow_array($num_wga_exp_query);
```

modules/t/DumpWGAExpectedTags.t
```diff
@@ -0,0 +1,82 @@
...
+
+open( my $fh, '<', $test_wga_expected_file) or die "Cannot open $test_wga_expected_file for reading\n";
+my @got_output = <$fh>;
+
+is_deeply( \@got_output, $exp_output, 'wga_expected tags dumped correctly to a file' );
+
+# Check that wga_expected tags are deleted from the database and no others
+my $sql3 = "SELECT COUNT(*) AS wga_exp_rows_after FROM method_link_species_set_tag WHERE tag = 'wga_expected'";
+my $db_vals3 = $dbc->db_handle->selectrow_hashref($sql3);
+my $we_count_after = $db_vals3->{wga_exp_rows_after};
```
JAlvarezJarreta, line 73:
```suggestion
my ($num_wga_exp_after) = $dbc->db_handle->selectrow_array($num_wga_exp_query);
```

modules/t/DumpWGAExpectedTags.t
```diff
@@ -0,0 +1,82 @@
...
+
+my $sql4 = "SELECT COUNT(*) AS total_rows_after FROM method_link_species_set_tag";
+my $db_vals4 = $dbc->db_handle->selectrow_hashref($sql4);
+my $all_count_after = $db_vals4->{total_rows_after};
```
JAlvarezJarreta, line 77:
```suggestion
my ($num_tags_after) = $dbc->db_handle->selectrow_array($num_tags_query);
```

modules/t/DumpWGAExpectedTags.t
```diff
@@ -0,0 +1,82 @@
...
+
+is( $we_count_after, 0, 'All wga_expected tags deleted from the method_link_species_set_tag table' );
+is( $all_count_after, $all_count - $we_count, 'The number of remaining entries in method_link_species_set_tag correct' );
```
JAlvarezJarreta, line 80:
```suggestion
is( $num_wga_exp_after, 0, 'All wga_expected tags deleted from the method_link_species_set_tag table' );
is( $num_tags_after, $num_tags_before - $num_wga_exp_before, 'Only wga_expected tags have been removed' );
```

JAlvarezJarreta (changes requested):
Not all the changes are mandatory, but most work towards simplicity of the code and readability (in my opinion).

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/DumpWGAExpectedTags.pm
```diff
@@ -0,0 +1,71 @@
...
+}
+
+
+sub dump_wga_expected {
+    my ($self) = shift @_;
```
carlacummins, line 46: This is quite an unusual way to get the arguments. If you look at examples throughout the codebase, we generally do this sort of thing in one of two ways:
`my $self = shift;` : for cases where we're **not** expecting any additional arguments, or
`my ($self, $x, $y) = @_;` : for cases where we _are_ expecting additional arguments.

In perl, `$_` and `@_` are like defaults, so calling `shift` is the same as calling `shift @_`. But you're also kind of dereferencing the first element of `@_` here by putting `($self)`. It works in this case because perl does a lot of guesswork on your behalf (which is bad sometimes because it masks things like this) and happened to guess what you're trying to do correctly in this case, but in a more complex case, I think this could lead to issues. Would recommend:
```suggestion
    my $self = shift;
```

modules/t/DumpWGAExpectedTags.t
carlacummins, line 40: Or
```suggestion
my $sql1 = "SELECT COUNT(*) AS total_rows FROM method_link_species_set_tag";
my $all_count = $dbc->sql_helper()->execute_single_result(-SQL => $sql1);
```

carlacummins (approved):

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/DumpWGAExpectedTags.pm
ipilizota: Yes, I remember your comment @JAlvarezJarreta. That's the only one that I initially didn't do anything about as I wanted to hear @carlacummins opinion ;)

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/DumpWGAExpectedTags.pm
ipilizota: Thank you very much for a concise explanation @carlacummins :)

modules/t/DumpWGAExpectedTags.t
ipilizota: Honestly, I haven't tried. I was following the logic from [here](https://github.com/Ensembl/ensembl-compara/blob/master/modules/t/HomologyIDMapping.t#L47) (lines 47, 55, 69).

modules/t/DumpWGAExpectedTags.t
ipilizota: Your suggestions regarding my queries and counts are so nice and elegant :) Why do we need brackets around `$num_tags_before`?

modules/t/DumpWGAExpectedTags.t
ipilizota: Regarding the 2nd `is(..)`, I am a bit reluctant to claim that `Only wga_expected tags have been removed` because it only checks the number of entries. In theory (if we had not output from the 1st `is(...)`), some other tags could've been deleted and the final numbers would still be ok. Perhaps it's safer to keep the original message if someone takes the code to repurpose it for its own use. But then, if your suggestion is in line with typical Compara practices on similar checks, let's go for your message :)

modules/t/DumpWGAExpectedTags.t
carlacummins: Because `selectrow_array` is returning an array, we are telling perl to store the first element of the array into `$num_tags_before`, rather than the whole array. If you were expecting 2 fields returned, you would do something like:
```
my ($element1, $element2) = $dbc->db_handle->selectrow_array($query);
```

ipilizota: Perfectly clear, as always @carlacummins :) Thank you.

modules/t/DumpWGAExpectedTags.t
JAlvarezJarreta: No problems about the message :+1:

ipilizota:
Thank you @JAlvarezJarreta. I appreciate the learning experience :)

JAlvarezJarreta (approved):

PR #281, opened 2021-04-27
Feature/homology annotation

CristiGuijarro:
## Description

The diamond results as they were favoured high `perc_ids` across local alignments regardless of the size of the aligned region. We would prefer a high `perc_id` in favour of covering as much of the query sequence as possible.

**Related JIRA tickets:**
- ENSCOMPARASW-4415

#### Change 1
- There were a couple of small bugfixes following untested commited review suggestions - see bugfix commit

#### Change 2
- The `PeptideAlignFeature` and `PeptideAlignFeatureAdaptor` were altered to take into account the `alignment_length` for sorting and with a cut off of 50% coverage. Furthermore the stats that are included in the `homology_member` table are explicitly stored negating the previous `update_alignment_stats` method.

## Testing
The pipeline was run cleanly through using a few parameter tweaks to pick the best. The distribution of `perc_id` was discussed during a scrum. See attached rainbow histogram.
[homology_dist_plot.pdf](https://github.com/Ensembl/ensembl-compara/files/6384194/homology_dist_plot.pdf)

## Notes
We may want to include a cut off for the `perc_id` in the future. This was discussed because we are not looking to provide a blastp tool - however as shown in histogram, the number of results below 25% `perc_id` are likely real, but very distant homologies for now - when the scope of data provided by this pipeline changes, it will be the most appropriate time to include these cut offs.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/DBSQL/PeptideAlignFeatureAdaptor.pm
+++ b/modules/Bio/EnsEMBL/Compara/DBSQL/PeptideAlignFeatureAdaptor.pm
@@ -365,7 +365,7 @@ sub filter_top_PAFs {
 
     foreach my $hit_genome_db_id (keys %by_query) {
         foreach my $sub_features (values %{$by_query{$hit_genome_db_id}}) {
-            my @pafList = sort sort_by_score_evalue_pid_and_alnlen @$sub_features;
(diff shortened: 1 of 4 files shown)


JAlvarezJarreta (approved):
Simple, easy changes, and some bugfixes, which are always nice to see :nerd_face:

JAlvarezJarreta (approved):

PR #286, opened 2021-04-30
Bugfix/prune supertrees

carlacummins:
## Description

The `$gene_tree_adaptor->delete_tree` method did not consider whether the tree being deleted was a subtree linked to a supertree. This can result in supertrees with unlinked leaf nodes, which breaks our supertree model and subsequent `$tree->expand_subtrees` calls. Here, supertree handling is introduced. Deleted trees are checked for supertree links and handled accordingly. If subtree deletion would result in:
- <2 subtrees remaining, the whole supertree is removed, leaving the other subtrees
- \>=2 subtrees, the supertree is pruned to remove unlinked leaves. The supertree's alignment is also pruned to remove members from the deleted subtree.

**Related JIRA tickets:**
- ENSCOMPARASW-4339
- ENSCOMPARASW-4338

#### _clean_supertree method
- automatically detect subtrees in the `delete_tree` method and pass off to a new `_clean_supertree` method
- clean the supertree using the rules described above

#### GeneAlignMemberAdaptor->delete_members
- new method to delete some members from a gene_align objects to help with cleaning of the supertree alignment

#### Improved detection of supertree leaves
- old code defined as a node with 0 children, but this doesn't hold true for supertrees
- leaves are now defined as 
    - for non-supertrees : node with 0 children
    - for supertrees : node with 1 child, where that child is a root node

## Testing
Many runs of the reindexing pipeline with manual checking of resulting trees and alignments

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/DBSQL/GeneAlignAdaptor.pm
+++ b/modules/Bio/EnsEMBL/Compara/DBSQL/GeneAlignAdaptor.pm
@@ -158,6 +158,7 @@ sub store {
 =head2 delete
 
  Arg [1]    : Bio::EnsEMBL::Compara::AlignedMemberSet $aln
+ Arg [2]    : (optional) arrayref of members to remove
  Example    : $AlignedMemberAdaptor->delete($aln)
  Description: Deletes an AlignedMemberSet object from a Compara database
  Returntype : none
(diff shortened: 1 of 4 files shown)


modules/Bio/EnsEMBL/Compara/GeneTreeNode.pm
```diff
@@ -312,7 +312,9 @@ sub is_leaf {
     if ( $child_count == 0 ) {
         return 1;
     } elsif ( $child_count == 1 && $self->tree->tree_type eq 'supertree' ) {
-        return 1;
+        my $child = $self->children->[0];
+        return 1 if $child->node_id == $child->root->node_id;
+        return 0;
```
JAlvarezJarreta, line 317: This is just out of curiosity, could you do `return ($child->node_id == $child->root->node_id)`?

modules/Bio/EnsEMBL/Compara/DBSQL/GeneAlignAdaptor.pm
```diff
@@ -175,4 +176,31 @@ sub delete {
...
+    my ($self, $aln, $members) = @_;
+
+    assert_ref_or_dbID($aln, 'Bio::EnsEMBL::Compara::AlignedMemberSet', 'aln');
+    my $dbID = ref($aln) ? $aln->dbID : $aln;
+
+    my $sql = 'DELETE FROM gene_align_member WHERE gene_align_id = ? AND seq_member_id = ?';
+    my $delete_sth = $self->dbc->prepare($sql);
+    foreach my $member ( @$members ) {
+        $delete_sth->execute($dbID, $member->dbID);
+    }
```
JAlvarezJarreta, line 201: Maybe a `seq_member_id IN (?)` and then `join(',', map {$_->dbID} @$members)` would have been more efficient, but I don't think the performance is going to be an issue here :)

JAlvarezJarreta (approved):
This must have been a super hard debugging process. Great job! :+1:

modules/Bio/EnsEMBL/Compara/DBSQL/GeneAlignAdaptor.pm
carlacummins: I thought about this - apparently `IN` performance degrades ~500 values, so I thought it was safest to do a loop just incase the list of members is long.

modules/Bio/EnsEMBL/Compara/GeneTreeNode.pm
carlacummins: Yeah, I like that
```suggestion
        return ($child->node_id == $child->root->node_id);
```

PR #288, opened 2021-05-05
Prevent rapid release reference comparing to self

CristiGuijarro:
## Description

If references can come from the rapid release, then they need to be prevented from running against themselves as rapid release queries.

**Related JIRA tickets:**
- ENSCOMPARASW-4446

## Overview of changes

Small necessary code additions in order to prevent the rapid release query comparing against itself as a reference.
Additional bug fixes spotted during PR #287  canonical `seq_members` were not being wholly copied. Reverted to a more simple method of copying. This was tested in the same pipeline.

## Testing
[See pipeline run here.](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensadmin&host=mysql-ens-compara-prod-3&port=4523&dbname=cristig_homology_annotation_no_self&passwd=xxxxx)
Database checked for any matches against itself and that `species_sets` are not contaminated with itself.

## Notes

There will still be a duplicate entry for the same genome, but hopefully different assembly/genebuild in the `genome_db` table. See ENSCOMPARASW-3545.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/HomologyAnnotation/BlastFactory.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/HomologyAnnotation/BlastFactory.pm
@@ -91,6 +91,7 @@ sub write_output {
 
     my $step              = $self->param('step');
     my @query_member_list = @{$self->param('query_members')};
+    my $gdb_adaptor       = $self->compara_dba->get_GenomeDBAdaptor;
     my @funnel_output;
 
     foreach my $genome ( @query_member_list ) {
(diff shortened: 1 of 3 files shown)


dthybert (commented):
I have one question about the change "Bugfix in data copying as detected in DCs". You replaced all the code based on the compara API by a simple db copy, so you bypass the API layer by doing so. While the code is simpler to understand, in longer term it may be less maintainable (for example if there is a change of schema ). Does it follow the coding spirit in the compara code overall? For the second commit it seems fine to me.

CristiGuijarro:
> I have one question about the change "Bugfix in data copying as detected in DCs". You replaced all the code based on the compara API by a simple db copy, so you over pass the API layer by doing so. While the code is simpler to understand, in longer term it may be less maintainable (for example if there is a change of schema ). Does it follow the coding spirit in the compara code overall? For the second commit it seems fine to me.

This db copy is a method we use in many other places too, regardless of maintenance and schema changes - so I would say it does follow the coding spirit. I intended to use the API for that reason, but as you can see elsewhere in this script, using the db copy is a necessary evil, the API simply won't work for the copying in this particular per-species compara database type. For other examples of the use of this method see: [populate_new_database.pl](https://github.com/Ensembl/ensembl-compara/blob/e11758e3af2c16076fe6d8eea043dfca6d9dfaf9/scripts/pipeline/populate_new_database.pl#L1).

dthybert (approved):

PR #295, opened 2021-06-08
Adaptation of species tree pipeline to use components

awgymer:
## Description
A grasses species tree is to be generated and changes are required to make the current code work for polyploid genomes with components.

**Related JIRA tickets:**
- [ENSCOMPARASW-4432](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-4432)
- [ENSCOMPARASW-4418](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-4418)

#### Change 1
- Apply filter for polyploid genomes with components such that genome dbs for the principals are excluded.
The largest change is a new method, filter_and_convert_genome_db_ids, in modules/Bio/EnsEMBL/Compara/Utils/DistanceMatrix.pm
Code for filtering before calling this method is added to runnables in modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree.

#### Change 2
- A change to the names used in the tree is made to include component information. Due to newick file format restrictions the use of display_name turns out to regularly cause errors and the parameter controlling this option has been removed.

#### Change 3
- PipeConfig changes to pass `genome_dump_dirs` parameter to runnables which require it for filtering

#### Change 4
- Changes to the `empty_submatrix` subroutine in `DistanceMatrix.pm` to allow it to take an array of `genome_db_ids` which should not be used to determine "emptiness". Currently the use-case for this is to consider a submatrix empty if it includes only groups beginning `mrg_` and/or the `genome_db_id` of the overall outgroup species. 

## Testing
[Pipeline run-through with grasses](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensadmin&host=mysql-ens-compara-prod-5&port=4615&dbname=agymer_grasses_species_tree&passwd=xxxxx)
[Pipeline run-through for all plants](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensadmin&host=mysql-ens-compara-prod-5&port=4615&dbname=agymer_all_plants_tree&passwd=xxxxx) (This doesn't produce an ideal tree but that is unrelated to the components changes)

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/CreateSpeciesTree_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/CreateSpeciesTree_conf.pm
@@ -235,6 +235,7 @@ sub pipeline_analyses {
           -parameters => {
               'taxonomic_ranks' => $self->o('taxonomic_ranks'),
               'custom_groups'   => $self->o('custom_groups'  ),
+              'genome_dumps_dir' => $self->o('genome_dumps_dir'),
           },
           -flow_into => {
               '2->A' => [ 'neighbour_joining_tree' ],
@@ -271,6 +272,7 @@ sub pipeline_analyses {
               'erable_exe'    => $self->o('erable_exe'),
               'unroot_script' => $self->o('unroot_script'),
               'reroot_script' => $self->o('reroot_script'),
+              'genome_dumps_dir' => $self->o('genome_dumps_dir'),
           },
           -flow_into => {
               1 => ['copy_files_to_sketch_dir'],
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/AdjustBranchLengths.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/AdjustBranchLengths.pm
@@ -52,6 +52,8 @@ sub param_defaults {
     my $self = shift;
     return {
         %{$self->SUPER::param_defaults},
+        'output_display_name'  => 0, # when genome_db_ids are replaced in the final tree, 
+        							 # use $gdb->display_name instead. default is $gdb->name
     };
 }
 
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/AdjustBranchLengths.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/AdjustBranchLengths.pm
@@ -52,8 +52,6 @@ sub param_defaults {
     my $self = shift;
     return {
         %{$self->SUPER::param_defaults},
-        'output_display_name'  => 0, # when genome_db_ids are replaced in the final tree, 
-        							 # use $gdb->display_name instead. default is $gdb->name
     };
 }
 
@@ -64,8 +62,18 @@ sub run {
 	# generate file in correct format for erable
 	my $working_dir = $self->param('working_dir') || $self->worker_temp_directory;
 	my $erable_phylip = "$working_dir/erable.phy";
+
+	my $ss_adaptor = $self->compara_dba->get_SpeciesSetAdaptor;
+	my $species_set = $ss_adaptor->fetch_by_dbID($self->param_required('species_set_id'));
+	my @genome_dbs = grep {
+		$_->name ne 'ancestral_sequences'
+		&& !($_->is_polyploid && !defined $_->genome_component)
+	} @{ $species_set->genome_dbs };
+
+	my $gdb_id_map = { map {$_->_get_genome_dump_path($self->param('genome_dumps_dir')) => $_->dbID} @genome_dbs };
+
 	my $mash_matrix = Bio::EnsEMBL::Compara::Utils::DistanceMatrix->new( -file => $self->param_required('mash_dist_file') );
-	$mash_matrix = $mash_matrix->convert_to_genome_db_ids($self->compara_dba->get_GenomeDBAdaptor);
+	$mash_matrix = $mash_matrix->filter_and_convert_genome_db_ids($gdb_id_map);
 	$mash_matrix->phylip_from_matrix($erable_phylip, 'multi');
 
 	# write topology tree to file
@@ -99,7 +107,8 @@ sub run {
 	while ( $tree =~ m/gdb([0-9]+)/ ) {
 		my $this_gdb_id = $1;
 		my $gdb = $genome_db_adaptor->fetch_by_dbID($this_gdb_id);
-		my $species_name = $self->param('output_display_name') ? $gdb->display_name : $gdb->name;
+		my $gcomp = $gdb->genome_component ? ".comp" . $gdb->genome_component : '';
+		my $species_name = $gdb->name . $gcomp;
 		$tree =~ s/gdb$this_gdb_id/$species_name/;
 	}
 
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/AdjustBranchLengths.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/AdjustBranchLengths.pm
@@ -108,7 +108,7 @@ sub run {
 		my $this_gdb_id = $1;
 		my $gdb = $genome_db_adaptor->fetch_by_dbID($this_gdb_id);
 		my $gcomp = $gdb->genome_component ? ".comp" . $gdb->genome_component : '';
-		my $species_name = $gdb->name . $gcomp;
+		my $species_name = $self->param('output_display_name') ? $gdb->display_name : $gdb->name . $gcomp;
 		$tree =~ s/gdb$this_gdb_id/$species_name/;
 	}
 
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/CheckSketches.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/CheckSketches.pm
@@ -49,7 +49,6 @@ use base ('Bio::EnsEMBL::Compara::RunnableDB::BaseRunnable');
 use Bio::EnsEMBL::Compara::Utils::DistanceMatrix;
 use File::Basename;
 use Data::Dumper;
-use Try::Tiny;
 
 sub fetch_input {
 	my $self = shift;
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/CheckSketches.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/CheckSketches.pm
@@ -181,14 +181,9 @@ sub _check_for_distance_files {
 	foreach my $dfile ( @dist_files ) {
 		print " --- checking $dfile\n";
 		my $dist_matrix = Bio::EnsEMBL::Compara::Utils::DistanceMatrix->new( -file => $dfile );
-		try {
-			$dist_matrix = $dist_matrix->filter_and_convert_genome_db_ids($gdb_id_map);
-			return $dfile;
-		} catch {
-			warn "Caught error whilst parsing distance matrix file $dfile\nError: $_";
-		}
-	}	
-
+		$dist_matrix = $dist_matrix->filter_and_convert_genome_db_ids($gdb_id_map);
+		return $dfile if defined $dist_matrix;
+	}
 	return undef;
 }
 
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/CheckSketches.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/CheckSketches.pm
@@ -49,6 +49,7 @@ use base ('Bio::EnsEMBL::Compara::RunnableDB::BaseRunnable');
 use Bio::EnsEMBL::Compara::Utils::DistanceMatrix;
 use File::Basename;
 use Data::Dumper;
+use Try::Tiny;
 
 sub fetch_input {
 	my $self = shift;
@@ -77,7 +78,11 @@ sub fetch_input {
 			$species_set = $ss_adaptor->fetch_by_dbID($species_set_id);
 			$self->param('collection', $species_set->name); # set this for file naming later
 		}
-		@genome_dbs = @{ $species_set->genome_dbs };
+		@genome_dbs = grep {
+			$_->name ne 'ancestral_sequences'
+			&& !($_->is_polyploid && !defined $_->genome_component)
+		} @{ $species_set->genome_dbs };
+
 	}
 
 	$self->param( 'genome_dbs', \@genome_dbs );
@@ -155,7 +160,7 @@ sub find_file_for_gdb {
 	
 	$suffixes = [''] unless defined $suffixes->[0];
 
-	my $prefix = $gdb->name . "." . $gdb->assembly;
+	my $prefix = basename($gdb->_get_genome_dump_path($self->param('genome_dumps_dir')));
 	foreach my $suffix ( @$suffixes ) {
 		my @found_files = glob "$dir/$prefix.*$suffix";
 		return $found_files[0] if $found_files[0];
@@ -169,19 +174,19 @@ sub _check_for_distance_files {
 	my $dir = $self->param_required('sketch_dir');
 	my $collection = $self->param_required('collection');
 	my @dist_files = glob "$dir/*$collection*.dists";
-	my @needed_gdb_ids = map {$_->dbID} @{ $self->param('genome_dbs') };
+	my $gdb_id_map = { map {$_->_get_genome_dump_path($self->param('genome_dumps_dir')) => $_->dbID} @{ $self->param('genome_dbs') } };
 
 	my $gdb_adaptor = $self->compara_dba->get_GenomeDBAdaptor;
 
 	foreach my $dfile ( @dist_files ) {
 		print " --- checking $dfile\n";
 		my $dist_matrix = Bio::EnsEMBL::Compara::Utils::DistanceMatrix->new( -file => $dfile );
-		$dist_matrix = $dist_matrix->convert_to_genome_db_ids($gdb_adaptor);
-
-		my @matrix_members = $dist_matrix->members;
-		my $overlap = $self->_overlap(\@needed_gdb_ids, \@matrix_members);
-		print " --- --- overlap = $overlap (" . scalar @needed_gdb_ids . " needed)\n";
-		return $dfile if $overlap >= scalar @needed_gdb_ids;
+		try {
+			$dist_matrix = $dist_matrix->filter_and_convert_genome_db_ids($gdb_id_map);
+			return $dfile;
+		} catch {
+			warn "Caught error whilst parsing distance matrix file $dfile\nError: $_";
+		}
 	}	
 
 	return undef;
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/PermuteMatrix.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/PermuteMatrix.pm
@@ -70,6 +70,7 @@ sub fetch_input {
 	my $self = shift;
 
 	my $gdb_adaptor = $self->compara_dba->get_GenomeDBAdaptor;
+	my $ss_adaptor = $self->compara_dba->get_SpeciesSetAdaptor;
 	my $ncbi_adaptor = $self->compara_dba->get_NCBITaxonAdaptor;
 	
 	# parse matrix and replace filenames with genome_db_ids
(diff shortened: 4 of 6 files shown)


modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/PermuteMatrix.pm
```diff
@@ -96,9 +108,23 @@ sub fetch_input {
...
+
+		# Make sure that the outgroup is in the list if this is the determined root taxon
+		my @excepts = ();
+		if ( $group_taxon_id == $self->param('root_id') ) {
+			print "\t -- root_id is " . $self->param('root_id') . "\n" if $self->debug;
+			my $outgroup_gdb = $gdb_adaptor->fetch_by_dbID($self->param_required('outgroup_id'));
+			push(@group_gdbs, $outgroup_gdb) unless grep{$_->dbID == $self->param_required('outgroup_id')} @group_gdbs;
+		}
+		else {
+			@excepts = ($self->param_required('outgroup_id'));
```
CristiGuijarro, line 120:
```suggestion
		my $excepts = [];
		if ( $group_taxon_id == $self->param('root_id') ) {
			print "\t -- root_id is " . $self->param('root_id') . "\n" if $self->debug;
			my $outgroup_gdb = $gdb_adaptor->fetch_by_dbID($outgroup_id);
			push(@group_gdbs, $outgroup_gdb) unless grep{$_->dbID == $outgroup_id} @group_gdbs;
		}
		else {
			$excepts = [$outgroup_id];
```
Purely to improve readability

modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/PermuteMatrix.pm
```diff
@@ -113,10 +139,11 @@ sub fetch_input {
 			next unless $prev_group_taxon->has_ancestor($current_group_taxon);
 			print "\t -- " . $prev_group_taxon->name . " is a member of this group - collapsing it\n" if $self->debug;
 			my $prev_group_gdbs = $gdb_adaptor->fetch_all_current_by_ancestral_taxon_id($prev_group);
+			print "\t -- Prev IDs: " . join(", ", map {$_->dbID} @{$prev_group_gdbs}) . "\n" if $self->debug;
 			$submatrix = $submatrix->collapse_group_in_matrix( $prev_group_gdbs, "mrg_$prev_group" );
 		}
         
-        next if Bio::EnsEMBL::Compara::Utils::DistanceMatrix->empty_submatrix($submatrix);
+        next if Bio::EnsEMBL::Compara::Utils::DistanceMatrix->empty_submatrix($submatrix, \@excepts);
```
CristiGuijarro, line 146:
```suggestion
        next if Bio::EnsEMBL::Compara::Utils::DistanceMatrix->empty_submatrix($submatrix, $excepts);
```
If follow above suggestion

modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/PermuteMatrix.pm
```diff
@@ -150,23 +177,27 @@ sub _add_outgroup {
 	my @sub_gdb_ids = $submatrix->members;
 	my $rep_gdb_id  = $sub_gdb_ids[0];
 
+	#print " -- submatrix genomes: " . join(", ", @sub_gdb_ids) . "\n" if $self->debug;
+	#print " -- fullmatrix genomes: " . join(", ", $full_matrix->members) . "\n" if $self->debug;
+
```
CristiGuijarro, line 182:
```suggestion
```

modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/PermuteMatrix.pm
```diff
@@ -150,23 +177,27 @@ sub _add_outgroup {
...
 	my ( $closest_gdb_id, $min_distance) = (undef, 100);
 	foreach my $full_key ( $full_matrix->members ) {
 		next if ( defined $self->param('blacklisted_genome_db_ids') && grep { $full_key eq $_ } @{$self->param('blacklisted_genome_db_ids')} ); # skip any that are on the naughty list
 		next if (grep { $_ eq $full_key } @sub_gdb_ids); # skip any that exist in the submatrix
 		if ( $full_matrix->distance($rep_gdb_id, $full_key) < $min_distance ) {
 			$closest_gdb_id = $full_key;
 			$min_distance = $full_matrix->distance($rep_gdb_id, $full_key);
+			#print "$full_key gives dist $min_distance\n";
```
CristiGuijarro, line 190:
```suggestion
```

modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/PermuteMatrix.pm
```diff
@@ -150,23 +177,27 @@ sub _add_outgroup {
...
 		}
 	} 
 
 	unless ($closest_gdb_id) {
 		print "Can't find a suitable outgroup\n";
 		return $submatrix;
 	}
-
```
CristiGuijarro, line 167: Put back

modules/Bio/EnsEMBL/Compara/Utils/DistanceMatrix.pm
```diff
@@ -96,13 +96,20 @@ sub members {
...
 	my ($matrix, $s1, $s2, $new_distance) = @_;
 
-	return $matrix->{$s1}->{$s2} unless defined $new_distance;
+	if ( defined $new_distance ){
 	$matrix->{$s1}->{$s2} = $new_distance;
 	$matrix->{$s2}->{$s1} = $new_distance;
 	$matrix->{$s1}->{$s1} = 0;
 	$matrix->{$s2}->{$s2} = 0; # deal with diagonals
 	return $matrix;
 }
```
CristiGuijarro, line 105: It's sometimes hard to determine indentation on github - but you are allowed to correctly indent this curly brace, if it is not already correctly indented - to fit the function change.

modules/Bio/EnsEMBL/Compara/Utils/DistanceMatrix.pm
```diff
@@ -253,14 +243,14 @@ sub collapse_group_in_matrix {
 	my ($matrix, $to_collapse_full, $new_group_id) = @_;
 
 	# print "!!!! collapsing [" . join(',', map {$_->name . '(' . $_->dbID . ')'} @$to_collapse) . "] to $new_group_id\n";
-
```
CristiGuijarro, line 256: Put back deleted line (and optionally uncomment print to a `if $self->$debug` - since this is out of the scope of changes you were making) , or remove commented line too :+1:

modules/Bio/EnsEMBL/Compara/Utils/DistanceMatrix.pm
```diff
@@ -253,14 +243,14 @@ sub collapse_group_in_matrix {
...
 	# first, check species are in the matrix
 	my @to_collapse;
 	foreach my $gdb ( @$to_collapse_full ) {
 		push( @to_collapse, $gdb ) if ( defined $matrix->{$gdb->dbID} );
 	}
 	return $matrix if scalar @to_collapse == 0;
 
+
```
CristiGuijarro, line 253:
```suggestion
```

modules/Bio/EnsEMBL/Compara/Utils/DistanceMatrix.pm
```diff
@@ -130,44 +137,27 @@ sub parse_matrix {
 	return $matrix;
 }
 
-=head2 convert_to_genome_db_ids
+=head2 filter_and_convert_genome_db_ids
```
CristiGuijarro, line 140: I think this method change is a huge improvement :+1:

modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/CheckSketches.pm
```diff
@@ -169,19 +174,19 @@ sub _check_for_distance_files {
...
-		my @matrix_members = $dist_matrix->members;
-		my $overlap = $self->_overlap(\@needed_gdb_ids, \@matrix_members);
-		print " --- --- overlap = $overlap (" . scalar @needed_gdb_ids . " needed)\n";
-		return $dfile if $overlap >= scalar @needed_gdb_ids;
+		try {
+			$dist_matrix = $dist_matrix->filter_and_convert_genome_db_ids($gdb_id_map);
+			return $dfile;
+		} catch {
+			warn "Caught error whilst parsing distance matrix file $dfile\nError: $_";
+		}
```
CristiGuijarro, line 189: eHive has built in warnings that we use : `$self->warning()` - this means the warning will be appropriately captured and stored in the database - so we should stick to this.
```suggestion
        if ( -e $dfile ) {
            $dist_matrix = $dist_matrix->filter_and_convert_genome_db_ids($gdb_id_map);
            return $dfile;
        }
        else {
            $self->warning("Could not parse distance matrix file $dfile");
        }
```
(Excuse the indentation in this suggestion - it probably doesn't fit)

modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/CheckSketches.pm
```diff
@@ -49,6 +49,7 @@ use base ('Bio::EnsEMBL::Compara::RunnableDB::BaseRunnable');
 use Bio::EnsEMBL::Compara::Utils::DistanceMatrix;
 use File::Basename;
 use Data::Dumper;
+use Try::Tiny;
```
CristiGuijarro, line 52:
```suggestion
```
(If follow suggestion for try-catch block below)

CristiGuijarro (changes requested):
Some nice and simple changes, just a few comments to address, plenty are "suggestions".

CristiGuijarro:
I recommend looking at review as a whole before addressing the comments, because many are linked and won't make sense alone :sweat_smile:

modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/CheckSketches.pm
```diff
@@ -155,7 +160,7 @@ sub find_file_for_gdb {
 	
 	$suffixes = [''] unless defined $suffixes->[0];
 
-	my $prefix = $gdb->name . "." . $gdb->assembly;
+	my $prefix = basename($gdb->_get_genome_dump_path($self->param('genome_dumps_dir')));
 	foreach my $suffix ( @$suffixes ) {
 		my @found_files = glob "$dir/$prefix.*$suffix";
```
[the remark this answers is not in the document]
awgymer: The use of `_get_genome_dump_path` here was to get the filename, which the `.msh` files mimic - there is no api call for generating the filename separate from the path. The reason for not using the method itself to generate the whole path was concern over this line: `my $subdir = $self->dbID ? Bio::EnsEMBL::Hive::Utils::dir_revhash($self->dbID) : '';`. To me it appears that it will add a subdir if the genome_db has an ID? The `.msh` files so far as I can tell are never in subdirs?

modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/CheckSketches.pm
awgymer: The error here doesn't occur when the `$dfile` doesn't exist (I mean, I guess one would, but the way we gather them they should), but rather when `filter_and_convert_genome_db_ids` dies because an expected genome_db_id is missing. Therefore I don't believe you can rely on `-e $dfile`. 

This wasn't required before because the previous code here had a different method of checking if all the expected IDs were in the distance matrix.  `filter_and_convert_genome_db_ids` now just dies if this is the case.

modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/CheckSketches.pm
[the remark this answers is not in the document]
CristiGuijarro: :+1: forgot `dir_rev_hash` - although not in this scope, I wonder if this is something we should consider adding as an optional flag for $subdir.

modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/CheckSketches.pm
CristiGuijarro: When `filter_and_convert_genome_db_ids` dies that should be enough. eHive will capture that. Additionally you could include a `$self->warning()` in `filter_and_convert_genome_db_ids` before it dies.
I meant to put `if ( -s $dfile ) {` rather than `-e` - but had only had one coffee.

awgymer: But we don't want the whole process to die if `filter_and_convert` dies. That step is supposed to loop through all prior/current dist files and look for one which already covers all the genome_dbs we need. The `$dfile` come from a glob so should always exist (technically you have a race condition there but oh well...). Maybe `filter_and_convert` should return `undef` if there are missing db IDs but my python mindset is to make stuff fail at the point of error and deal with it when you need to, not to mysteriously pass around null values as a proxy for failure

CristiGuijarro: Drop the python mindset then, this is Perl. Return `undef` for missing gdbs. You have over-complicated this.

CristiGuijarro:
Something I just noticed - you have put the jira ticket link in the commit - this is not something we do - can you edit the commit to remove the ticket?

modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/AdjustBranchLengths.pm
```diff
@@ -52,8 +52,6 @@ sub param_defaults {
     my $self = shift;
     return {
         %{$self->SUPER::param_defaults},
```
awgymer, line 54:
```suggestion
        %{$self->SUPER::param_defaults},
        'output_display_name'  => 0, # when genome_db_ids are replaced in the final tree, 
        							 # use $gdb->display_name instead. default is $gdb->name
```

awgymer:
> Something I just noticed - you have put the jira ticket link in the commit - this is not something we do - can you edit the commit to remove the ticket?

I can. But note that it simply follows the format suggested [here](https://www.ebi.ac.uk/seqdb/confluence/display/ENS/Ensembl+Coding+Guidelines#EnsemblCodingGuidelines-CommitMessages) for larger commit messages.

CristiGuijarro:
> > Something I just noticed - you have put the jira ticket link in the commit - this is not something we do - can you edit the commit to remove the ticket?
> 
> I can. But note that it simply follows the format suggested [here](https://www.ebi.ac.uk/seqdb/confluence/display/ENS/Ensembl+Coding+Guidelines#EnsemblCodingGuidelines-CommitMessages) for larger commit messages.

The tickets are in the PR description, and the commit is not so large "Adaptation of species tree pipeline to use components" is enough. Because we work with PRs, and this is dev work, our commit messages are unlikely to require a ticket - they are not issue-linked in that way either.

modules/Bio/EnsEMBL/Compara/PipeConfig/CreateSpeciesTree_conf.pm
```diff
@@ -235,6 +235,7 @@ sub pipeline_analyses {
           -parameters => {
               'taxonomic_ranks' => $self->o('taxonomic_ranks'),
               'custom_groups'   => $self->o('custom_groups'  ),
+              'genome_dumps_dir' => $self->o('genome_dumps_dir'),
```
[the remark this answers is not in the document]
awgymer: It's unclear to me where `genome_dumps_dir` gets set in the first place, it isn't in the `default_options` - possibly an eHive default somewhere. I simply copied the way it was already being passed to `check_sketches` in the PipeConfig.

modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/CheckSketches.pm
awgymer: Changes applied in later commits.

modules/Bio/EnsEMBL/Compara/PipeConfig/CreateSpeciesTree_conf.pm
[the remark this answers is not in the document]
CristiGuijarro: [ENV.pm](https://github.com/Ensembl/ensembl-compara/blob/master/modules/Bio/EnsEMBL/Compara/PipeConfig/ENV.pm#L83
)
[ComparaGeneric_conf.pm](https://github.com/Ensembl/ensembl-compara/blob/master/modules/Bio/EnsEMBL/Compara/PipeConfig/ComparaGeneric_conf.pm#L45)
[CreateSpeciesTree_conf.pm](https://github.com/Ensembl/ensembl-compara/blob/master/modules/Bio/EnsEMBL/Compara/PipeConfig/CreateSpeciesTree_conf.pm#L47)
Here is a breadcrumb trail :slightly_smiling_face:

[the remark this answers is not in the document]
CristiGuijarro: It is fine to do it the way you have done it, but I would be tempted to move all of those specific parameter calls - because they are needed in most analyses, to `pipeline_wide_parameters` - including the one in `check_sketches` - just to reduce the code repetition. And in `pipeline_wide_parameters` it is nice to have a quick look in the guihive too.

modules/Bio/EnsEMBL/Compara/RunnableDB/SpeciesTree/PermuteMatrix.pm
```diff
@@ -150,23 +178,27 @@ sub _add_outgroup {
...
 			$closest_gdb_id = $full_key;
 			$min_distance = $full_matrix->distance($rep_gdb_id, $full_key);
+			#print "$full_key gives dist $min_distance\n";
 		}
 	} 
 
 	unless ($closest_gdb_id) {
 		print "Can't find a suitable outgroup\n";
 		return $submatrix;
 	}
```
awgymer, line 198:
```suggestion
	}
	
```

modules/Bio/EnsEMBL/Compara/Utils/DistanceMatrix.pm
awgymer: Well I did indent it originally but since the only chnages to those lines were whitespace changes they got dropped.

modules/Bio/EnsEMBL/Compara/Utils/DistanceMatrix.pm
```diff
@@ -253,14 +244,14 @@ sub collapse_group_in_matrix {
 	my ($matrix, $to_collapse_full, $new_group_id) = @_;
 
 	# print "!!!! collapsing [" . join(',', map {$_->name . '(' . $_->dbID . ')'} @$to_collapse) . "] to $new_group_id\n";
```
awgymer, line 246:
```suggestion
	# print "!!!! collapsing [" . join(',', map {$_->name . '(' . $_->dbID . ')'} @$to_collapse) . "] to $new_group_id\n";
	
```

modules/Bio/EnsEMBL/Compara/PipeConfig/CreateSpeciesTree_conf.pm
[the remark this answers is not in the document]
awgymer: I think if this is a stylistic choice in that manner I'm opposed to it because it actually makes it _less_ clear which parameters a particular step requires/uses at a glance.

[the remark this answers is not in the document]
CristiGuijarro: Up to you, but this wouldn't make a difference in the case of parameters that are passed from the runnables prior to the analysis - so for the sake of consistency and wholesome "knowing what parameters a particular step requires" - you will find that it is more than `genome_dumps_dir`. We generally pass them in individually in this way if we want to override the `default` or `pipeline_wide_parameters` - which we do not in this case.

modules/Bio/EnsEMBL/Compara/Utils/DistanceMatrix.pm
CristiGuijarro: So it this indented as it should be to fit the `if ( ) {` block? (I think no, since it doesn't show up as a change, and it would be odd if they were already indented). If not, please do :slightly_smiling_face:

CristiGuijarro (approved):
Approved on the basis that a final test run is clean following addressing the review comments and that the un-indented `if` block is beautifully indented to fit the code change. This does not count as an unnecessary white space change, but a necessary one. Will merge following confirmation.

modules/Bio/EnsEMBL/Compara/Utils/DistanceMatrix.pm
```diff
@@ -253,7 +244,7 @@ sub collapse_group_in_matrix {
...
-
+
```
CristiGuijarro, line 247:
```suggestion

```

modules/Bio/EnsEMBL/Compara/Utils/DistanceMatrix.pm
```diff
@@ -253,7 +244,6 @@ sub collapse_group_in_matrix {
 	my ($matrix, $to_collapse_full, $new_group_id) = @_;
 
 	# print "!!!! collapsing [" . join(',', map {$_->name . '(' . $_->dbID . ')'} @$to_collapse) . "] to $new_group_id\n";
-
```
CristiGuijarro, line 256: Giving up on this stubborn line :facepalm: :sweat_smile:

CristiGuijarro (approved):

PR #302, opened 2021-06-24
Inserion of Insecta reference set as well a replacement of mouse by an…

dthybert:
## Description

_Describe the problem you're addressing here._
A reference set for insecta has been created and in the mammalian reference set the mouse define(define on the shared vertebrate and the mammalian)has been replaced by the guinea pig in the mammalian
**Related JIRA tickets:**
- ENSCOMPARASW-4495
- ENSCOMPARASW-4633

## Overview of changes

#### Change 1
- Add a new reference set for insecta.

#### Change 2
- change the mouse described in the mammalian set by the guinae pig

## Testing
running the create_allowed_species.py , if this script had faile it would ahve meant that the xml format was wrong. It successfully created the list of allowed species.

## Notes
_Optional extra information._


CristiGuijarro (approved):
Approved on the basis that for these new reference genomes, that there are no duplicates between servers cores - for example rapid release and main release. If there is, for now they will need to be manually included in the `references/production_reg_conf.pl` as `@overlap_species` :+1:

CristiGuijarro (changes requested):
Just noticed the name `insecta` (may) need to be swapped - this is a really closed off reference set - which means an awful lot of hexapods will be redirected to the "default" collection. This is okay if it is the intention, but got to check.

conf/references/mlss_conf.xml
```diff
@@ -252,6 +252,56 @@
       <genome name="apteryx_rowi"/>
     </collection>
 
+    <collection name="insecta">
```
CristiGuijarro, line 255: Can be `insects` or `hexapods` I think

CristiGuijarro:
Something else going forward - so not to be done in this PR, if we are having a very restrictive cladistically narrow `insecta` reference set, then perhaps we should consider an arthropod and hexapod shared base group, - to collectively cover other collections within arthropod that may also be cladistically narrow. Currently here we have the equivalent of a reference set for mammalia and none of the other vertebrates in keeping with a narrowed `insecta` reference set. That's equivalent to only mammals getting a specific reference set, and all fish, amphibians and sauropsids would use the `default`.

dthybert:
Good point, I think it is better to move to hexapod. It will be difficult at the moment to have an Arthropod dataset as the current hexapod reference dataset includes most of Metazoa arthropods (The spiders and some diptera were numerous enough to be bale to choose). But I agree that in the future we will need anyway an Arthropod reference set.

CristiGuijarro (approved):

PR #312, opened 2021-07-12
base_location parameter (analysis create_default_pairwise_mlss in EPOwExt)

ipilizota:
Edited definition of `base_location` parameter which affects `create_default_pairwise_mlss` analysis in `EPOwExt` pipeline

## Description

`EPOwExt` failed for Primates and Mammals 105 at the analysis `create_default_pairwise_mlss` due to undefined `base_location` parameter. Jorge realised it was due to changes in `release/104`, in particular that the `base_location` was not correctly defined at this place (as well).
 
**Related JIRA tickets:**
- ENSCOMPARASW-4669

## Overview of changes
- Defined `base_location` as in line 104

## Testing
- Not tested

## Notes
- Shall I initialise one pipeline and try to run it until this analysis? 
- Are we sure this sorts out the problem of prematurely released semaphores?


JAlvarezJarreta (approved):
This commit fixes the dependency issue (I have checked the code), and it is a simple change that has worked before, so there is no need to test it :+1: Regarding the semaphore issue, that needs to be clarified in the ticket first.

PR #317, opened 2021-07-26
Improve error handling before adding unit testing

twalsh-ebi:
## Description

Error handling bugfixes and improvements will facilitate unit testing.

**Related JIRA tickets:**
- ENSCOMPARASW-4528

## Overview of changes
- Check region parts in parse_region: where possible, the problematic part of the region string will be flagged.
- Check flank length in make_src_region_file: this will allow flank-length tests to be done in the unit test for this function.
- Fix handling of subprocess return codes: positive integers are treated as exit codes, negative integers are treated as signals.

## Testing
Tested with basic test runs.

### DIFF
--- a/scripts/hal_alignment/hal_gene_liftover.py
+++ b/scripts/hal_alignment/hal_gene_liftover.py
@@ -63,7 +63,7 @@ def make_src_region_file(regions: Iterable[Union[pybedtools.cbedtools.Interval, 
 
     """
     if flank_length < 0:
-        raise ValueError(f'Flank length must be greater than or equal to 0: {flank_length}')
+        raise ValueError(f"'flank_length' must be greater than or equal to 0: {flank_length}")
 
     with open(bed_file, 'w') as f:
         name = '.'
--- a/scripts/hal_alignment/hal_gene_liftover.py
+++ b/scripts/hal_alignment/hal_gene_liftover.py
@@ -58,7 +58,12 @@ def make_src_region_file(regions: Iterable[Union[pybedtools.cbedtools.Interval, 
         bed_file: Path of BED file to output.
         flank_length: Length of upstream/downstream flanking regions to request.
 
+    ValueError: If the flank length is negative.
+
     """
+    if flank_length < 0:
+        raise ValueError(f'Flank length must be greater than or equal to 0: {flank_length}')
+
     with open(bed_file, 'w') as f:
         name = '.'
         score = 0  # halLiftover requires an integer score in BED input
@@ -131,9 +136,6 @@ if __name__ == '__main__':
     src_bed_file = args.src_bed_file
     flank = args.flank
 
-
-    if flank < 0:
-        raise ValueError(f'Flank length must be greater than or equal to 0: {flank}')
 
     with TemporaryDirectory() as tmp_dir:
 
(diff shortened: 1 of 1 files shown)


scripts/hal_alignment/hal_gene_liftover.py
```diff
@@ -58,7 +58,12 @@ def make_src_region_file(regions: Iterable[Union[pybedtools.cbedtools.Interval,
         bed_file: Path of BED file to output.
         flank_length: Length of upstream/downstream flanking regions to request.
 
+    ValueError: If the flank length is negative.
```
JAlvarezJarreta, line 61: Following Google docstring format:
```suggestion
    Raises:
        ValueError: If `flank_length` is negative.
```

scripts/hal_alignment/hal_gene_liftover.py
```diff
@@ -58,7 +58,12 @@ def make_src_region_file(regions: Iterable[Union[pybedtools.cbedtools.Interval,
...
+
     """
+    if flank_length < 0:
+        raise ValueError(f'Flank length must be greater than or equal to 0: {flank_length}')
```
JAlvarezJarreta, line 65: [Suggestion] I'm usually prefer writing the variable name, so the user knows directly which one you are talking about:
```suggestion
        raise ValueError(f"'flank_length' must be greater than or equal to 0: {flank_length}")
```

scripts/hal_alignment/hal_gene_liftover.py
```diff
@@ -83,21 +88,28 @@ def parse_region(region: str) -> SimpleRegion:
...
+        r'^(?P<chrom>[^:]+):(?P<start>[0-9]+)-(?P<end>[0-9]+):(?P<strand>.+)$'
     )
     match = seq_region_regex.match(region)
 
     try:
         region_chrom = match['chrom']  # type: ignore
-        match_start = match['start']  # type: ignore
-        match_end = match['end']  # type: ignore
+        match_start = int(match['start'])  # type: ignore
+        region_end = int(match['end'])  # type: ignore
```
JAlvarezJarreta, line 98: I guess `# type: ignore` is not required any more?
```suggestion
        match_start = int(match['start'])
        region_end = int(match['end'])
```

scripts/hal_alignment/hal_gene_liftover.py
```diff
@@ -154,7 +163,11 @@ def parse_region(region: str) -> SimpleRegion:
             with Popen(cmd2, stdin=p1.stdout) as p2:
                 p2.wait()
                 if p2.returncode != 0:
-                    raise RuntimeError(f'pslPosTarget terminated with signal {-p2.returncode}')
+                    status_type = 'exit code' if p2.returncode > 0 else 'signal'
+                    raise RuntimeError(
+                        f'pslPosTarget terminated with {status_type} {abs(p2.returncode)}')
```
JAlvarezJarreta, line 168: If this gets any more complex (and the same applied below), I would encourage to do a "returncode function" to process it (just for future reference).

JAlvarezJarreta (changes requested):

scripts/hal_alignment/hal_gene_liftover.py
twalsh-ebi: Unfortunately `mypy` still produces an error message if this comment is not present.

scripts/hal_alignment/hal_gene_liftover.py
twalsh-ebi: Fair enough, it is starting to get quite complicated.

JAlvarezJarreta (approved):
Nice and easy changes :+1:

PR #322, opened 2021-08-10
Add ability to exclude species from GenomeDBFactory by name

awgymer:
## Description

GenomeDBFactory returns an array of genome dbs, which can be filtered by broad categories (i.e. polyploids). Production requested the ability to filter specific species. This functionality is added by these changes. `exclude_species` may be supplied as an array or a space-separated string.

**Related JIRA tickets:**
- ENSCOMPARASW-4673
- ENSPROD-6669

## Overview of changes
Add `exclude_species` parameter to relevant config files and code for filtering based on it within `GenomeDBFactory.pm`

#### Change 1
- `exclude_species` added to `PipeConfig/Parts/UpdateMemberNamesDescriptions.pm` and `PipeConfig/UpdateMemberNamesDescriptions_conf.pm` and defaults to an empty array 

#### Change 2
- Code to filter genome dbs using the supplied `exclude_species` array added to `GenomeDBFactory.pm`  

## Testing
`LoadMembers` plants pipeline run with and without `exclude_species` array and reduction in genome dbs passed from `load_genomedb_factory` to `load_genomedb`

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMemberNamesDescriptions_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMemberNamesDescriptions_conf.pm
@@ -53,6 +53,7 @@ sub default_options {
 
         #Pipeline capacities:
         'update_capacity'                           => '5',
+        #Outgroup species to exclude from GenomeDBFactory
         'exclude_species'  => [],
 
     };
@@ -80,5 +81,4 @@ sub pipeline_analyses {
 }
 
 1;
-
 
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMemberNamesDescriptions_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMemberNamesDescriptions_conf.pm
@@ -53,6 +53,7 @@ sub default_options {
 
         #Pipeline capacities:
         'update_capacity'                           => '5',
+        'exclude_species'  => [],
 
     };
 }
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/GenomeDBFactory.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/GenomeDBFactory.pm
@@ -127,7 +127,7 @@ sub fetch_input {
     # filter out exclude_species
     my %exclude_species_map;
     my $excluded_species = $self->param('exclude_species');
-    if ( ref $_exclude_species eq 'ARRAY' ) {
+    if ( ref $excluded_species eq 'ARRAY' ) {
         @exclude_species_map{ @$excluded_species } = ();
     } else {
         @exclude_species_map{ split(/,/, $excluded_species) } = ();
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/GenomeDBFactory.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/GenomeDBFactory.pm
@@ -61,7 +61,7 @@ sub param_defaults {
         'component_genomes' => 1,
         'normal_genomes'    => 1,
         'ancestral_genomes' => 0,
-        # list of genome db names to exlcude
+        # list of genome db names to exclude
         'exclude_species' => [],
 
         # List of GenomeDB attribute names that will be added to the output_ids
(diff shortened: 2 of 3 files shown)


modules/Bio/EnsEMBL/Compara/RunnableDB/GenomeDBFactory.pm
```diff
@@ -121,6 +123,21 @@ sub fetch_input {
     $genome_dbs = [grep {not $_->genome_component} @$genome_dbs] if not $self->param('component_genomes');
     $genome_dbs = [grep {($_->name eq 'ancestral_sequences') or $_->is_polyploid or $_->genome_component} @$genome_dbs] if not $self->param('normal_genomes');
 
+
+    # filter out exclude_species
+    my %exclude_species;
+    my $_exclude_species = $self->param('exclude_species');
```
CristiGuijarro, line 129: Following the general code style of compara, we never precede a variable name with `_`. A better way to distinguish might be to name the parameter variable `$exclude_species`, and the array version later on something like `$excluded_species`. The same goes for `@_gdbs`. In Perl, the preceding underscore convention is restricted to subroutines and variables locally scoped in just those subrutines.

modules/Bio/EnsEMBL/Compara/RunnableDB/GenomeDBFactory.pm
```diff
@@ -121,6 +123,21 @@ sub fetch_input {
...
+    if ( ref $_exclude_species eq 'ARRAY' ) {
+        @exclude_species{ @$_exclude_species } = ();
+    } else {
+        @exclude_species{ split(/\s+/, $_exclude_species) } = ();
```
CristiGuijarro, line 133: Perhaps this has been discussed as a necessary use-case - but I otherwise don't think this needs to be compatible with a space separated string at all. Was this specified somewhere?

CristiGuijarro: If space separated strings were not specified to be expected, a more common use-case would be to pass in a comma-separated list, or simply restrict it to either arrays or a single string variable for a single genome to exclude.

modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMemberNamesDescriptions_conf.pm
```diff
@@ -53,6 +53,7 @@ sub default_options {
 
         #Pipeline capacities:
         'update_capacity'                           => '5',
+        'exclude_species'  => [],
```
CristiGuijarro, line 56:
```suggestion
        #Outgroup species to exclude from GenomeDBFactory
        'exclude_species'  => [],
```

modules/Bio/EnsEMBL/Compara/RunnableDB/GenomeDBFactory.pm
awgymer: It was something discussed with Jorge and he linked the example of some code in `Utils`. I just lifted it pretty directly so it may be that a csv string is better?

CristiGuijarro: It might make sense in some `Utils` code, but I think parameter wise for a hive runnable, passing just an array would be enough, and this code could be simplified - it would also reduce confusion with the variable names as a bonus.
I would approve based on a change to csv  rather than space separated - because I could see this being used more on the command line when initialising pipeline

modules/Bio/EnsEMBL/Compara/PipeConfig/UpdateMemberNamesDescriptions_conf.pm
CristiGuijarro: Can we stick a comment or a new line in above the parameter, just to separate it from the pipeline capacities, because it is not a pipeline capacity?

modules/Bio/EnsEMBL/Compara/RunnableDB/GenomeDBFactory.pm
awgymer: Have changed to split on a comma in subsequent commits

modules/Bio/EnsEMBL/Compara/RunnableDB/GenomeDBFactory.pm
```diff
@@ -61,6 +61,8 @@ sub param_defaults {
         'component_genomes' => 1,
         'normal_genomes'    => 1,
         'ancestral_genomes' => 0,
+        # list of genome db names to exlcude
```
CristiGuijarro, line 64:
```suggestion
        # list of genome db names to exclude
```

PR #323, opened 2021-08-12
add python wrapper to dump a list of genomes from core

thiagogenez:
## Description

This script is a generic (not Cactus specific) Python script that parser the Perl `dump_genome_from_core.pl` to dump a list of genomes from core dbs.

### DIFF
--- a/scripts/dumps/dump_genome_from_core.py
+++ b/scripts/dumps/dump_genome_from_core.py
@@ -145,7 +145,8 @@ def parse_yaml(file, dest):
                     core_db=core_db,
                     query='SELECT meta_value FROM meta WHERE meta_key="assembly.accession";',
                 )
-                gca_number = gca_number.replace(".", "v").lower()
+                # fix the name
+                gca_number = gca_number.replace(".", "v").replace('_','').lower()
                 specie_name = "{}_{}".format(specie_name, gca_number)
 
             if specie_name is not None:
--- a/scripts/dumps/dump_genome_from_core.py
+++ b/scripts/dumps/dump_genome_from_core.py
@@ -4,7 +4,16 @@ import fileinput
 import argparse
 import subprocess
 import os
-import yaml
+
+
+try:
+    import yaml
+except ModuleNotFoundError as err:
+    # Error handling
+    print(err)
+    print('Please, run "pip install PyYAML" to install PyYAML module')
+    exit(1)
+
 from yaml.loader import SafeLoader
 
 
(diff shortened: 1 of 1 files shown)


scripts/dumps/dump_genome_from_core.py
```diff
@@ -0,0 +1,184 @@
...
+                    fasta_filename="{}/{}.fa".format(dest, specie_name),
+                )
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--yaml", required=True, type=str, help="YAML input file")
+    parser.add_argument(
+        "--output", required=False, default=None, type=str, help="Processed output file"
+    )
```
twalsh-ebi, line 166: Unless there is an overriding reason, I would omit arguments if they are being assigned their default values.
I've also suggested changing the help string for the `--output` option to be explicit that it is a directory.
```suggestion
    parser.add_argument("--yaml", required=True, help="YAML input file")
    parser.add_argument("--output", help="Output directory.")
```

scripts/dumps/dump_genome_from_core.py
```diff
@@ -0,0 +1,184 @@
...
+    args = parser.parse_args()
+
+    with open(args.yaml, mode="r", encoding="utf-8") as f:
```
twalsh-ebi, line 169: Could this line be incorporated in the `parse_yaml` function?

scripts/dumps/dump_genome_from_core.py
```diff
@@ -0,0 +1,184 @@
...
+
+        if args.output is None:
+            args.output = os.path.dirname(os.path.realpath(f.name))
+        else:
+            args.output = os.path.abspath(args.output)
```
twalsh-ebi, line 174: Is it necessary to get the real/absolute path of the output directory here? Would the script still function with the relative path?

scripts/dumps/dump_genome_from_core.py
```diff
@@ -0,0 +1,184 @@
...
+
+        if not os.path.isdir(args.output):
+            print(
+                "{} does not exist for output, please create it first".format(
+                    args.output
+                )
+            )
+            sys.exit(1)
```
twalsh-ebi, line 182: Could this be changed to create the output directory if it does not exist?

twalsh-ebi (changes requested):
Overall it's a nice script, and I can attest that it does the job. But I have added some comments and suggestions for adjustments.

PR #335, opened 2021-09-03
mlss_conf.xml for Gene Tree Benchmark project with a test collection

ipilizota:
## Description

mlss_conf file for the Gene Tree Benchmark project

**Related JIRA tickets:**
- ENSCOMPARASW-4738

## Overview of changes
* Placed in `benchmark/` as agreed with DT and JA
* Contains a collection of 12 species for development purposes (6 Vertebrates + 6 Metazoa; the choice of divisions was suggested in the previously created related ticket)   

## Testing
No testing.


JAlvarezJarreta (commented):
I can see the division `benchmark` exists, as you mention, but based on its content, what makes it different than this one?

ipilizota:
I used that division when running the protein tree pipeline `release/101` code on `e98` dataset. It was used to time protein tree pipelines on `release/XXX` code with `e98` dataset. It also contains corresponding species tree and `production_reg_conf.pl`. I might end up with a different `production_reg_conf.pl`. (More info: It looks like I won't create a master database yet. In the next piece of code, I will be accessing/downloading cores from a specified host and perhaps I will put that in the `production_reg_conf.pl`. The current file seems `e98`-centric. I understand this can all be easily edited but I wonder if it'd be cleaner and more convenient to keep this project separate from that.)

JAlvarezJarreta:
I have spoken with @dthybert and he agrees with me this should be moved into the existing division `benchmark`. If we need to retrieve what was there, we can always roll back in time.

ipilizota:
Done.

conf/benchmark/mlss_conf.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<compara_db division="benchmark">
+
+  <collections>
+
+    <collection name="dev_test">
```
JAlvarezJarreta, line 6: If you plan running gene trees pipelines maybe "default" might make your life easier in terms of configuration. But who doesn't like an additional challenge? :sunglasses:

JAlvarezJarreta (approved):

conf/benchmark/mlss_conf.xml
ipilizota: Ah, I didn't think that'd be an issue. Would it require much fiddling around beside e.g. setting `'collection' => 'dev_test'` in `ProteinTrees_conf.pm`?
I was planning to use `default` for a proper size dataset, e.g. the one that will be used for benchmarking.
I could name this one `default` to spare myself some trouble or keep it as it is for the learning purposes O:)

JAlvarezJarreta: Keep it! Keep it! :smiling_imp:

dthybert (approved):

ipilizota:
Travis failed because species weren't found in the species tree -> expected

PR #341, opened 2021-09-09
Make copy_data.pl work for polyploid PWAs

JAlvarezJarreta:
## Description

The latest changes introduced to fix `group_id` (because it is based in LASTZ_CHAIN instead of LASTZ_NET for trace-back purposes, and its range needs to be based in the LASTZ_NET MLSS id in the release database) in `copy_data.pl` were not compatible with polyploid PWAs, as the `group_id` range is based on more than one LASTZ_CHAIN MLSS id, and the code was not compatible with a fix of different offsets. This PR fixes that, altogether with some changes on how to process `group_id`.

**Related JIRA tickets:**
- ENSCOMPARASW-4496

## Overview of changes
Since `group_id` has mainly an internal purpose, I don't think it is worth checking it is within the correct range (if we are interested in this, we may want to create a DC instead). Thus, the script now only checks if at least one `group_id` needs to be fixed, and if so, two generic queries are applied that take care of fixing every `group_id` of that MLSS id, regardless of their "source offset".

## Testing
I have cloned the e105 release database and used the script to copy one of Arthur's LastZ pipelines (wheat self-alignment), one of the LastZ pipelines without polyploids and rice EPO (doesn't have `group_id`). They have all been copied correctly and the `group_id` has been modified successfully (when needed).

### DIFF
--- a/scripts/pipeline/copy_data.pl
+++ b/scripts/pipeline/copy_data.pl
@@ -932,7 +932,7 @@ sub copy_conservation_scores {
   if ($max_cs < 10**10) {
     ## Need to add $method_link_species_set_id * 10^10 to the internal_ids
     $fix = $lower_limit;
-  } elsif ($max_cs and $min_cs >= $lower_limit) {
+  } elsif ($min_cs >= $lower_limit and $max_cs < $upper_limit) {
     ## Internal IDs are OK.
     $fix = 0;
   } else {
@@ -1037,7 +1037,7 @@ sub copy_constrained_elements {
   if ($max_ce < 10**10) {
     ## Need to add $method_link_species_set_id * 10^10 to the internal_ids
     $fix = $lower_limit;
-  } elsif ($max_ce and $min_ce >= $lower_limit) {
+  } elsif ($min_ce >= $lower_limit and $max_ce < $upper_limit) {
     ## Internal IDs are OK.
     $fix = 0;
   } else {
--- a/scripts/pipeline/copy_data.pl
+++ b/scripts/pipeline/copy_data.pl
@@ -538,7 +538,6 @@ sub copy_genomic_align_blocks {
 
   my $minmax_sql = qq{ SELECT
         MIN(gab.genomic_align_block_id), MAX(gab.genomic_align_block_id),
-        MIN(gab.group_id), MAX(gab.group_id),
         MIN(ga.genomic_align_id), MAX(ga.genomic_align_id),
         MIN(gat.node_id), MAX(gat.node_id),
         MIN(gat.root_id), MAX(gat.root_id)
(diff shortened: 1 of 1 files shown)


JAlvarezJarreta:
I haven't rebased the branch properly from `master`, but this PR is picking the `copy_data.pl` changes, so I think Travis failing on the branch but not the PR is a green light.

scripts/pipeline/copy_data.pl
```diff
@@ -709,16 +686,28 @@ sub copy_genomic_align_blocks {
...
          " FROM genomic_align_block WHERE method_link_species_set_id = $mlss_id",
        @copy_data_args);
   } else {
     copy_table($from_dbc, $to_dbc, 'genomic_align_block', "method_link_species_set_id = $mlss_id", undef, "skip_disable_keys");
   }
 
+  # Fix the group_id when it is not in its MLSS id range
+  if ($fix_gab_gid) {
+      $to_dbc->do("UPDATE genomic_align_block
+          SET group_id = CAST(CONCAT(method_link_species_set_id, LPAD(group_id, 10, '0')) AS UNSIGNED INTEGER)
```
CristiGuijarro, line 702: Just want to check that this is the behaviour I am seeing: `LPAD` will pad the existing `group_id` to a length of 10 digits, and the `method_link_species_set_id` will essentially be prepended to this in the `CONCAT` step, if the `group_id` is less than `10000000000` - then as a form of overkill, the next update cuts it down to size/ups the size when not between the constraints?

scripts/pipeline/copy_data.pl
```diff
@@ -709,16 +686,28 @@ sub copy_genomic_align_blocks {
   my @copy_data_args = (undef, 'skip_disable_keys');
 
   #copy genomic_align_block table
-  if ($fix_gab or $fix_gab_gid) {
+  if ($fix_gab) {
     copy_data($from_dbc, $to_dbc,
        "genomic_align_block",
-       "SELECT genomic_align_block_id+$fix_gab, method_link_species_set_id, score, perc_id, length, group_id+$fix_gab_gid, level_id, direction".
+       "SELECT genomic_align_block_id+$fix_gab, method_link_species_set_id, score, perc_id, length, group_id, level_id, direction" .
```
CristiGuijarro, line 692: This is a vague and general formatting-related comment, but linked to both readability and potential future changes, and applies to the code prior to change too.
These longish queries could benefit from being written:
```
    my $sql = qq/
        SELECT 
            genomic_align_block_id + $fix_gab,
            method_link_species_set_id,
            score,
            perc_id,
            length,
            group_id,
            level_id,
            direction
        FROM
            genomic_align_block
        WHERE
            method_link_species_set_id = $mlss_id;
    /;
```
And then included in the `copy_data` code.

CristiGuijarro (approved):
1 comment can be disregarded as a formatting comment, not necessary to apply, the other is more a sanity check. Approved on the basis this urgently needs to be merged for `e106` and testing has proven successful. Overall a nice clean implementation and tidy-up :+1:

scripts/pipeline/copy_data.pl
JAlvarezJarreta: I completely agree, but I tried to keep the formatting changes to a minimum because this file needs plenty! (starting from the fact it is 2 spaces instead of 4 :facepalm:) I can do this one though :+1:

scripts/pipeline/copy_data.pl
JAlvarezJarreta: You got it right. As a note, I wouldn't call the second update an overkill since those IDs below 10^10 cannot be updated using `RIGHT` (Arthur did a great job noticing this and including it in his notes).

scripts/pipeline/copy_data.pl
CristiGuijarro: Yes, it is the fact that a lot of the formatting sucks that wasn't going to strongly suggest that one!

scripts/pipeline/copy_data.pl
CristiGuijarro: I didn't spot that :+1: 
But in any case, sanity satisfied :smile:

ipilizota (approved):
Looking at the problem and PR description, I can understand the reasons behind each code change. I approve the PR on the basis of successful testing and hope that the introduced changes are sufficient to resolve the issue :)

PR #346, opened 2021-09-14
Set 'default' collection for each non-vertebrate protein tree

JAlvarezJarreta:
## Description

Change all divisions' default collection to `default` in order to make it easier for us (no renaming back and forth for reruns of parts of the gene trees pipelines) and for web (to access the data via the default API configuration).

**Related JIRA tickets:**
- ENSCOMPARASW-4706

## Overview of changes
Created `default` collection for metazoa and pan (was already done for plants), and changed the protein trees collection from `<division>` to `default` (set in the base [`ProteinTrees_conf`](https://github.com/Ensembl/ensembl-compara/blob/6daeeac20e895de36293316a6dd84a91fc19edcf/modules/Bio/EnsEMBL/Compara/PipeConfig/ProteinTrees_conf.pm#L72) pipeconfig).

## Testing
Not tested as the changes are similar to those in the recent plants e106 PR.

## Notes
No further changes are required as the pipeline already handles everything as `default` and then renames the gene trees `clusterset_id` at the end. Even the email analysis has `default` hardcoded, so everything should be fine.

### DIFF
--- a/conf/metazoa/mlss_conf.xml
+++ b/conf/metazoa/mlss_conf.xml
@@ -1,6 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <compara_db division="metazoa">
 
+  <!-- Collections are species-sets that are needed to define several mlsss -->
+  <collections>
+
+    <!-- All metazoa -->
+    <collection name="default">
+      <taxonomic_group taxon_name="Eukaryota"/>
+    </collection>
+
+  </collections>
+
   <pairwise_alignments>
 
     <!-- bees and stuff (A.mel, A.cep, B.imp, B.ter, N.vit) -->
@@ -54,7 +64,7 @@
   </pairwise_alignments>
 
   <gene_trees>
-    <protein_trees collection="metazoa"/>
+    <protein_trees collection="default"/>
   </gene_trees>
 
 </compara_db>
--- a/conf/pan/mlss_conf.xml
+++ b/conf/pan/mlss_conf.xml
@@ -1,8 +1,20 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <compara_db division="pan">
 
+  <!-- Collections are species-sets that are needed to define several mlsss -->
+  <collections>
+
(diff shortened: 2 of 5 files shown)


CristiGuijarro (approved):
Happy with all changes :+1: 

One quick note, because for some reason this works. In the `ProteinTrees_conf` - the pipeline name is based on:
```
'pipeline_name' => $self->o('collection') . '_' . $self->o('division').'_protein_trees_'.$self->o('rel_with_suffix'),
```
But the collection variable is only assigned/defaultified after this?
```
'collection'            => 'default',
```
A quick check/confirmation on that?

JAlvarezJarreta:
> Happy with all changes 
> 
> One quick note, because for some reason this works. In the `ProteinTrees_conf` - the pipeline name is based on:
> 
> ```
> 'pipeline_name' => $self->o('collection') . '_' . $self->o('division').'_protein_trees_'.$self->o('rel_with_suffix'),
> ```
> 
> But the collection variable is only assigned/defaultified after this?
> 
> ```
> 'collection'            => 'default',
> ```
> 
> A quick check/confirmation on that?

I think we should be fine since this has been like this for vertebrates and the pipeline never complained, right? :man_shrugging:

CristiGuijarro:
> > Happy with all changes
> > One quick note, because for some reason this works. In the `ProteinTrees_conf` - the pipeline name is based on:
> > ```
> > 'pipeline_name' => $self->o('collection') . '_' . $self->o('division').'_protein_trees_'.$self->o('rel_with_suffix'),
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But the collection variable is only assigned/defaultified after this?
> > ```
> > 'collection'            => 'default',
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > A quick check/confirmation on that?
> 
> I think we should be fine since this has been like this for vertebrates and the pipeline never complained, right?

Yes, that was my thoughts too :+1:

ipilizota (approved):

PR #350, opened 2021-09-23
Updated config files for new carp.

ens-carlos:
Added cyprinus_carpio_carpio for the new carp for e106.

## Description

_Describe the problem you're addressing here._

**Related JIRA tickets:**
- ENSCOMPARASW-XXXX

## Overview of changes

#### Change 1
- _detail 1.1_

#### Change 2
- _detail 2.1_

## Testing
_How was this tested? Have new unit tests been included?_

## Notes
_Optional extra information._

### DIFF
--- a/conf/vertebrates/allowed_species.json
+++ b/conf/vertebrates/allowed_species.json
@@ -49,6 +49,7 @@
     "cynoglossus_semilaevis",
     "cyprinodon_variegatus",
     "cyprinus_carpio",
+    "cyprinus_carpio_carpio",
     "danio_rerio",
     "dasypus_novemcinctus",
     "delphinapterus_leucas",
--- a/conf/vertebrates/allowed_species.json
+++ b/conf/vertebrates/allowed_species.json
@@ -48,7 +48,6 @@
     "cyclopterus_lumpus",
     "cynoglossus_semilaevis",
     "cyprinodon_variegatus",
-    "cyprinus_carpio",
     "cyprinus_carpio_carpio",
     "danio_rerio",
     "dasypus_novemcinctus",
--- a/conf/vertebrates/mlss_conf.xml
+++ b/conf/vertebrates/mlss_conf.xml
@@ -261,7 +261,7 @@
         <genome name="cyclopterus_lumpus" assembly="fCycLum1.pri"/>
         <genome name="nothobranchius_furzeri" assembly="Nfu_20140520"/>
         <genome name="oncorhynchus_kisutch" assembly="Okis_V2"/>
-        <genome name="cyprinus_carpio_carpio" assembly="Cypcar_WagV4.0"/>
+        <genome name="cyprinus_carpio_carpio"/>
       </species_set>
     </multiple_alignment>
     <multiple_alignment method="EPO_EXTENDED" gerp="1">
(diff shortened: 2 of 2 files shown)


conf/vertebrates/mlss_conf.xml
```diff
@@ -261,7 +261,7 @@
         <genome name="cyclopterus_lumpus" assembly="fCycLum1.pri"/>
         <genome name="nothobranchius_furzeri" assembly="Nfu_20140520"/>
         <genome name="oncorhynchus_kisutch" assembly="Okis_V2"/>
-        <genome name="cyprinus_carpio"/>
+        <genome name="cyprinus_carpio_carpio" assembly="Cypcar_WagV4.0"/>
```
twalsh-ebi, line 264:
```suggestion
        <genome name="cyprinus_carpio_carpio" assembly="Cypcar_WagV4.0"/>
```

twalsh-ebi, line 264: The inclusion of the assembly name here has implications for how the assembly is used.

If this genome is intended to be a reference only until a future assembly update replaces it, then the assembly name should be included in the genome element.

If, on the other hand, this assembly is to be preserved in the EPO MSA indefinitely, the assembly attribute should be omitted.

To put this in context, all of the other AQUA-FAANG species with the exception of _Sparus aurata_ have been configured without the assembly name.

twalsh-ebi (commented):
These configuration changes look good to me. I've commented on one change that may be appropriate, depending on the intended use of the Carp assembly in the EPO pipeline.

conf/vertebrates/allowed_species.json
```diff
@@ -49,6 +49,7 @@
     "cynoglossus_semilaevis",
     "cyprinodon_variegatus",
     "cyprinus_carpio",
```
dthybert, line 51: in allowed_species we keep only the species/production name that we will process. Is there any reason why cyprinus_carpio is not removed or just an oversight? We need to keep the number of non strain species at 200.

conf/vertebrates/mlss_conf.xml
dthybert: so we need to remove the assembly name for all aqua-faang species then.

dthybert (changes requested):

conf/vertebrates/allowed_species.json
ens-carlos: I initially left it because all the other carps have that scientific name and I wasn't sure what the purpose of this list was. I've removed it.

conf/vertebrates/mlss_conf.xml
ens-carlos: Assembly name removed.

ipilizota:
@twalsh-ebi Re. Species tree: The same crossed my mind but then I remembered that the tree is computed for every release by Compara (unlike for e.g. Plants) so perhaps that's not a problem.

twalsh-ebi:
> @twalsh-ebi Re. Species tree: The same crossed my mind but then I remembered that the tree is computed for every release by Compara (unlike for e.g. Plants) so perhaps that's not a problem.

It may well be, but the mismatch between allowed-species list and species tree might nonetheless cause Travis to fail in the meantime.

twalsh-ebi:
Thanks @ens-carlos for these config changes. As we need to merge them into the `release/106` branch, all these changes have been committed separately to that branch in commit [6002627](https://github.com/Ensembl/ensembl-compara/commit/600262747324110d2920725464cf476c711c1551), along with an update of the Carp taxon name in `species_tree.branch_len.nw` so that it matches the updated name in `allowed_species.json`.

PR #353, opened 2021-09-27
Solve bugfix in update_master_db.pl

JAlvarezJarreta:
## Description

Avoid using vertannot as reference for the species that should be in compara master. Instead, use the list of current genomes in the database and just confirm they have a matching core db in the registry and the meta-data matches.

**Related JIRA tickets:**
- ENSCOMPARASW-4906

## Overview of changes
- The previous behaviour of this script was working as expected before freezing the list of species and when we were in charge of copying the species to vertannot. Since these two things have changed, relying on the information in the registry can end on releasing species that shouldn't. Thus, the script now uses the current species in compara master to check everything is correct.
- With this change, the `$division` information is no longer required.
- Reverted usage of `--check_species_missing_from_compara` flag, as we are the only ones using this script and we were using it with the `no` prefix.

## Testing
The script has been tested for plants e106, where vertannot has `theobroma_cacao_criollo` and it was being re-released before this update.

## Notes
- Updated POD (because why not!)

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/Parts/PrepareMasterDatabaseForRelease.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/Parts/PrepareMasterDatabaseForRelease.pm
@@ -188,7 +188,7 @@ sub pipeline_analyses_prep_master_db_for_release {
             -parameters => {
                 'update_metadata_script' => $self->o('update_metadata_script'),
                 'reg_conf'               => $self->o('reg_conf'),
-                'cmd' => 'perl #update_metadata_script# --reg_conf #reg_conf# --compara #master_db# --division #division# --nocheck_species_missing_from_compara'
+                'cmd'                    => 'perl #update_metadata_script# --reg_conf #reg_conf# --compara #master_db#',
             },
             -flow_into  => [ 'update_collection' ],
         },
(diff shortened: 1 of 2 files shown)


JAlvarezJarreta:
Tiny PR, only one reviewer required.

CristiGuijarro (commented):
I think this looks reasonable, I think. I am a little confused about the removal of the `$division` parameter. Without it, I believe the reports will be output for all divisions? Is there any conflict when there are multiple genomes with the same name, but different `taxon_id` on the server for example?

JAlvarezJarreta:
> I think this looks reasonable, I think. I am a little confused about the removal of the `$division` parameter. Without it, I believe the reports will be output for all divisions? Is there any conflict when there are multiple genomes with the same name, but different `taxon_id` on the server for example?

I have run it for plants and only `drosophila` and `caernohabditis` (hope I got this one right) were reported twice. Since the metadata information should be the same (besides the division, which we don't check any more), I don't foresee any problems (the "control" is now from the master database, so the division is "forced" that way). But happy to put it back if you feel it shouldn't be removed.

CristiGuijarro (approved):
I think keeping the `$division` functionality would be good. For the report on the genomes, worth keeping an eye on for `Pan`, `Metazoa` and `Fungi/Protists` because they as divisions will contain by default the conflicting genomes such as `drosophila` and `caernohabditis`  :crossed_fingers:

PR #358, opened 2021-10-22
Dump gene tree and MSA stats reports into TSV files

JAlvarezJarreta:
## Description

Two new runnables, one new script and several pipeline updates so the gene tree and MSA stats are no longer emailed but dumped into TSV (sorted by taxon name)in our shared space, making it easier to share the stats in the team and compare them between releases.

**Related JIRA tickets:**
- ENSCOMPARASW-4681

## Overview of changes
- Introduced required changes/additions in `ENV.pm`.

### MSA stats
- "New" `GenomicAlignBlock::StatsReport.pm` runnable: quite similar to the previous `EmailStatsReport.pm` (mainly removed the email stuff).
- Updated `epo_stats.pl` script and renamed to `msa_stats.pl`: added better usage handling + POD, and only one MLSS can be processed at a time (this is the usage we are doing now in any case, and I don't see the usefulness of combining stats of several MLSSs). Unfortunately, `git` has not been able to catch the renaming and changes, so a removal + new file have been introduced in the history. Sorry!
- Updated `EPO`, `EPOExtended`, `EPOwithExt` and `MercatorPecan` pipelines.

### Gene tree stats
- New `GeneTree::StatsReport.pm` runnable in a similar fashion to `GenomicAlignBlock::StatsReport.pm`: the main difference is that in this case the three tables returned are separated in three different files to ease comparison between releases.
- New `gene_tree_stats.pl` script to mimic `msa_stats.pl`, so both sets of stats are generated in a similar fashion, easing ad-hoc requirements.
- Updated `ProteinTrees` and `ncRNATrees` pipelines.

## Testing
1. Created every single pipeline to make sure the code is not buggy: [Protein trees](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-10&port=4648&dbname=jalvarez_default_vertebrates_protein_trees_106), [ncRNA trees](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-10&port=4648&dbname=jalvarez_default_vertebrates_ncrna_trees_106), [EPOwithExt](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-10&port=4648&dbname=jalvarez_mammals_epo_with_ext_106) and [MercatorPecan](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-10&port=4648&dbname=jalvarez_amniotes_mercator_pecan_106).
2. I have replaced the stats for the new runnable for both plants protein trees (after they finished), and the resulting files can be found in `/hps/nobackup/flicek/ensembl/compara/shared/gene_tree_stats`. I have compared the content with that received in the email and it is exactly the same.

## Notes
- I haven't touched `CreateHmmProfiles` pipeline as I couldn't decide if this runnable was suitable for this pipeline as well.
- I would suggest every division to run this as `standaloneJob.pl` for e106 so the comparisons can be done from e107.
- There are some cases where no column value can be used to ensure the sorting is always done in the same way, but it is still way less work than before.
- Sorry, couldn't preserve both the file and email functionality without making a much more time consuming change. The runnable can still be easily replaced if needed.


modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/StatsReport.pm
```diff
@@ -0,0 +1,94 @@
...
+
+
+sub fetch_input {
+    my $self = shift @_;
+
+    my $mlss_id = $self->param_required('mlss_id');
+    my $mlss = $self->compara_dba->get_MethodLinkSpeciesSetAdaptor->fetch_by_dbID($mlss_id);
+    $self->param('method_name', lc $mlss->method->type);
+
+    make_path($self->param_required('gene_tree_stats_shared_dir'));
```
CristiGuijarro, line 56: This should be in the `pipeline_create_commands` - I am assuming that you did not put it in using the `pipeline_create_commands_rm_mkdir` because of the `rm` part of `...rm_mkdir`. However, worry not, you can skip the `rm`: https://github.com/Ensembl/ensembl-compara/blob/bbc39e549207d9e8c09d6100744770afe401d75f/modules/Bio/EnsEMBL/Compara/PipeConfig/ComparaGeneric_conf.pm#L270

CristiGuijarro: But also, while you are there, could you change the description in the Arg to "if true" instead of "if false"? - you'll see what I mean :sweat_smile:  - my bad.

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/StatsReport.pm
```diff
@@ -0,0 +1,94 @@
...
+
+
+sub dump_stats_into_tsv_file {
+    my ($self, $type, $stats) = @_;
+
+    my $dump_file = $self->param('gene_tree_stats_shared_dir') . '/' . $self->param('method_name') . '_' . $type . '.tsv';
+    open( my $fh_tsv, '>', $dump_file ) || die "Could not open output file $dump_file";
+    print $fh_tsv "$stats\n";
+    close($fh_tsv);
+}
```
CristiGuijarro, line 91: Going to be awkward here and suggest that this is written either as a separate script or somehow placed in utils or the flatfiles.pm module. All new runnables should contain a maximum of `fetch_input`, `run` and `write_output` and if possible could avoid `fetch_input` in cases where the API is not needed to "fetch the input" - in this case you did require `fetch_input`.

modules/Bio/EnsEMBL/Compara/RunnableDB/GenomicAlignBlock/StatsReport.pm
```diff
@@ -0,0 +1,76 @@
...
+
+
+sub fetch_input {
+    my $self = shift @_;
+
+    my $mlss_id = $self->param_required('mlss_id');
+    my $mlss = $self->compara_dba->get_MethodLinkSpeciesSetAdaptor->fetch_by_dbID($mlss_id);
+    $self->param('method_name', lc $mlss->method->type);
+
+    make_path($self->param_required('msa_stats_shared_dir'));
```
CristiGuijarro, line 48: Same comment as for the gene trees - should be in pipeconf

modules/Bio/EnsEMBL/Compara/RunnableDB/GenomicAlignBlock/StatsReport.pm
```diff
@@ -0,0 +1,76 @@
...
+
+
+sub write_output {
+    my $self = shift @_;
+
+    my $stats_table = $self->param('stats_table');
+    my $dump_file = $self->param('msa_stats_shared_dir') . '/' . $self->param('method_name') . '.tsv';
+    open( my $fh_tsv, '>', $dump_file ) || die "Could not open output file $dump_file";
+    print $fh_tsv "$stats_table\n";
+    close($fh_tsv);
```
CristiGuijarro, line 72: This could use the utils module mentioned before in the gene tree stats report, and whilst it feels like scraping the bottom of the pot here, could reduce these 3 lines to 1.

CristiGuijarro (changes requested):
A huge improvement :+1: Thank you! 
The comments I have made are mostly linked to the code structure, and I think they need to be addressed.
Optionally, in the `gene_tree_stats.pl` and `msa_stats.pl` are several internalised formatting related methods, which are generic and not necessarily specific to the scripts in which they are contained. As we move forward to the future with gene tree overhaul/orthofinder, cactus integration etc, we are likely to generate more stats, which we may want to email or provide html for (if we get public html stuff sorted for codon) as well as dump to tsv - and these formatting modules could be useful at that point also. I leave this as an optional to ignore comment, because this is futuristic talking, and we don't know what we will want, and it is not hugely important right now.

modules/Bio/EnsEMBL/Compara/RunnableDB/GenomicAlignBlock/StatsReport.pm
JAlvarezJarreta: Although I would usually agree, I find these three lines easier to read, especially to people less familiar with Perl.

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/StatsReport.pm
JAlvarezJarreta: I thought these rules were going to be applied only to Python code :sweat_smile: But happy to move it to the `Utils` module.

modules/Bio/EnsEMBL/Compara/RunnableDB/GenomicAlignBlock/StatsReport.pm
CristiGuijarro: Easier to read than `dump_to_tsv_file( $dump_file, $stats_table )`?

JAlvarezJarreta: I thought you meant a different coding for these three lines :man_facepalming:

CristiGuijarro (approved):

scripts/production/gene_tree_stats.pl
```diff
@@ -0,0 +1,375 @@
...
+    unshift @gene_events_sorted, [
+        'Taxon ID',
+        'Taxon name',
+        'Nb of nodes',
+        'Nb of duplication nodes',
+        'Nb of gene splits',
+        'Nb of speciation nodes',
+        'Nb of dubious nodes',
+        'Avg confidence score',
+        'Avg confidence score on non-dubious nodes',
```
twalsh-ebi, line 331: With the values in these columns containing just the numbers, I'd suggest to indicate in the column heading that these are percentages:
```suggestion
        'Avg confidence score (%)',
        'Avg confidence score (%) on non-dubious nodes',
```

twalsh-ebi (commented):
It's looking very nice. I have just a couple of suggestions.

scripts/production/msa_stats.pl
```diff
@@ -0,0 +1,138 @@
...
+) or pod2usage(-verbose => 2);
+
+# Handle "print usage" scenarios
+pod2usage(-exitvalue => 0, -verbose => 1) if $help;
+pod2usage(-verbose => 1) if !$url or !$mlss_id;
+
+my $dba = Bio::EnsEMBL::Compara::DBSQL::DBAdaptor->go_figure_compara_dba($url);
+
+my $mlss = $dba->get_MethodLinkSpeciesSetAdaptor->fetch_by_dbID($mlss_id);
+my $mlss_id = $mlss->dbID;
```
twalsh-ebi, line 85: Could this assignment to `$mlss_id` be removed?

twalsh-ebi (approved):

PR #362, opened 2021-11-02
Remove the delete statement prior to insert

marcoooo:
## Description

GeneTreeHighlighting pipeline runtime is getting too long - simplify scripts and modules 

- ENSPROD-7140 

Relates to 
- https://github.com/Ensembl/ensembl-datacheck/pull/425 
- https://github.com/Ensembl/ensembl-production/pull/620

## Overview of changes

Remove the delete statement from the module simply because the related insert in using `ignore` statement to prevent duplicates. 

#### Change 1
- Remove the deletion from the module.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/DBSQL/XrefAssociationAdaptor.pm
+++ b/modules/Bio/EnsEMBL/Compara/DBSQL/XrefAssociationAdaptor.pm
@@ -68,10 +68,6 @@ my $insert_member_sql = $insert_member_base_sql. q/ select gene_member_id,?,? fr
  
 my $get_member_id_sql = q/select gene_member_id from gene_member where stable_id=? and source_name='ENSEMBLGENE'/;
 
-my $delete_member_sql = q/delete mx.* from member_xref mx, gene_member m, genome_db g
-where g.name=? and mx.external_db_id=?
-and g.genome_db_id=m.genome_db_id and m.gene_member_id=mx.gene_member_id/;
-
 my $base_get_sql = q/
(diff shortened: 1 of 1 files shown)


marcoooo:
I tested with actual data and source on `mysql-ens-test-1` and it worked fine. With a real increase in pipeline runtime!

twalsh-ebi (approved):
This PR looks fine to me, and I'm approving it. I'd also suggest adding a note to the `Description` of the `store_member_associations`, something like:
```
  Description: Method to retrieve external database accessions for genes in the supplied core and store
               them in the Compara database. Note: this method assumes that the member associations for
               the given external database have not been previously stored in the Compara database. This
               is true for the typical use case of this method, and making this assumption improves speed
               significantly. If this assumption is violated, it is the responsibility of the caller to
               ensure that previously member associations have been removed before calling this method.
```

marcoooo:
> This PR looks fine to me, and I'm approving it. I'd also suggest adding a note to the `Description` of the `store_member_associations`, something like:
> 
> ```
>   Description: Method to retrieve external database accessions for genes in the supplied core and store
>                them in the Compara database. Note: this method assumes that the member associations for
>                the given external database have not been previously stored in the Compara database. This
>                is true for the typical use case of this method, and making this assumption improves speed
>                significantly. If this assumption is violated, it is the responsibility of the caller to
>                ensure that previously member associations have been removed before calling this method.
> ```

Agreed, and pushed accordingly, although the insert being done with `IGNORE`, that wouldn't hurt too much and not insert duplicates.

PR #364, opened 2021-11-04
Implementation of a step that dumps genomes from cores

ipilizota:
## Description

At the very beginning, the Orthology Benchmark pipeline has to retrieve input data for orthology tools. In a previous PR, I addressed the problem of fetching a list of species names for a species set (collection) from `mlss_conf.xml` file. In this PR, I'm addressing the problem of dumping cores for all species in the list.

**Related JIRA tickets:**
- ENSCOMPARASW-4766

## Overview of changes
Implemented the step of dumping all cores of interest. Wrote corresponding tests. Added (hundreds of) files needed for testing. 

#### Change 1
- Added new functionality to `orthology_benchmark.py`.

#### Change 2
- Wrote tests for new functions in `test_orthology_benchmark.py`.
- Added files for test core dbs to `src/python/tests/databases/core/`. Test cores for `danio_rerio`, `mus_musculus`, `mus_musculus_cbaj` are empty because the corresponding test/function just retrieves the names of existing core dbs on a `host`. Files for `gallus_gallus` and human are not empty because they are used as test cases for actual dumping.
- Added expected output `.fasta` dumps to `src/python/tests/flatfiles/orth_benchmark/`.

#### Change 3
- Silenced `mypy` error by modifying `mypy.ini`.

## Testing
In addition to `pytests`, I also ran:
```
ibsub -m 2gb python /hps/software/users/ensembl/repositories/compara/ivana/gene_tree_bench/ensembl-compara/scripts/pipeline/orthology_benchmark.py --mlss_conf /hps/software/users/ensembl/repositories/compara/ivana/gene_tree_bench/ensembl-compara/conf/benchmark/mlss_conf.xml --species_set test_test --host mysql-ens-compara-prod-10 --port 4648 --user ensro --out_dir $HPS_HOME
```
Output `/hps/nobackup/flicek/ensembl/compara/ivana/test_test` looks as expected.

## Notes
Lines `40-57` in `test_orthology_benchmark.py` will be replaced with a nice and elegant solution using `import` code that TW is currently writing. `orthology_benchmark.py` and `test_orthology_benchmark.py` are far from finished anyways, and will gradually grow. Let's focus on extending functionality and keep moving 🙂


JAlvarezJarreta (changes requested):
Although `table.sql` for each core could be present just once in `core` and then symlinked in each corresponding folder, I can see the utility of having copies (furthermore, I'm going to assume they are all exact copies). However, the Python unit test db was designed to not require the empty files for each table: the schema is loaded and then the db is populated. So every empty file you have inside `databases/core` can be removed.

I think there are several design decisions that require further explanation in order to evaluate this code properly. For instance, the decision of picking up all cores in a server (which has a high chance of producing unexpected outcomes).

ipilizota:
* I removed all empty files from test core dbs. `table.sql` is indeed the same in all of them and it's still there. Shall I replace it with symlinks?
* The dataset for benchmarking might not contain core dbs from the same Ensembl release. That's why all available cores are listed and then the most recent one is chosen. That one query for the whole dataset could be replaced by one query per species if that's what you're thinking about. I thought the former would be less computationally intensive and requires just one connection to the server, so I went with it. Which unexpected outcomes do you refer to?
* Where did you gather info on incompatibility with SQLite? Travis? I thought Travis was failing because `pytest` examples weren't set up properly which we said would be more convenient to resolve through the PR review. Anyways, how can I ensure that the code is compatible with SQLite (or other database servers) before pushing it to the outside world?

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,134 @@
...
 
 from ensembl.compara.config import get_species_set_by_name
 
 
-def dump_genomes():
-    """Docstring"""
+def dump_genomes(species_list: List[str], species_set_name: str, host: str, port: int,
+                 user: str, out_dir: str) -> None:
+    """Dumps canonical peptides of protein-coding genes for a specified list of species
+    from the latest available core databases to FASTA files using `dump_gene_set_from_core.pl`.
```
JAlvarezJarreta, line 39: The advised format in docstrings is a short description in the first line, and then a larger description (if required). Something like:
```
    """Dumps canonical peptides of the given species' protein-coding genes.

    The dumping of each species is done for the latest available core databases to FASTA files
    using `dump_gene_set_from_core.pl`.
```

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,134 @@
...
+
+    Raises:
+        FileExistsError: If directory `out_dir/species_set_name` already exists.
+        MySQLdb.OperationalError: If connection to `host:port` cannot be established.
+        RuntimeError: If no cores are found for any species in `species_list` (`species_set_name`)
+                on `host:port`.
+
+    """
+    # Check access to the host & port
+    MySQLdb.connect(host=host, port=port, user=user)
```
JAlvarezJarreta, line 57: So I have just tested and it seems you can do:
```
from sqlalchemy import create_engine
eng = create_engine('mysql://ensro@mysql-ens-compara-prod-10:4648/')
eng.execute("SHOW DATABASES LIKE '%%_core_%%'").fetchall()
```
And you will get all core databases.

My concern about doing this is not about doing 1 search and then filtering (in most cases this will be more efficient), but with the fact that you are not specifying the release. If you are running your code, say for `vertannot`, you expect e105 cores and genebuild adds one e106 of one of the species you are searching for just before you run your code, you will end up with a different set of species. Not sure how relevant this is for your goal, but just mentioning a possible scenario.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,134 @@
...
+    dumps_dir = os.path.join(out_dir, species_set_name)
+    os.mkdir(dumps_dir)
+
+    script = os.path.join(os.environ["ENSEMBL_ROOT_DIR"], "ensembl-compara", "scripts", "dumps",
+                          "dump_gene_set_from_core.pl")
+
+    for core in dump_cores:
+        out_file = os.path.join(dumps_dir, core + ".fasta")
+        subprocess.run([script, "-core-db", core, "-host", host, "-port", str(port),
+                       "-outfile", out_file], capture_output=True, check=True)
```
JAlvarezJarreta, line 74: Why `subprocess.run([...], capture_output=True, check=True)` instead of `subprocess.check_call([...])`?

JAlvarezJarreta (commented):
Partial review, to give you some thoughts/ideas. I would leave the diverse `table.sql` just in case only a subset is required to be changed in the future.

ipilizota:
Thank you for getting started on this, @JAlvarezJarreta 👍

scripts/pipeline/orthology_benchmark.py
ipilizota: I'm aware that not all cores might be from the same release. That's favourable in situations when some species get retired/removed or if a core for newly available species is ready before others. There can be an option to specify a release, if needed.

scripts/pipeline/orthology_benchmark.py
ipilizota: `subprocess.check_call` is part of an older high-level API (prior to Python 3.5). https://docs.python.org/3/library/subprocess.html#older-high-level-api

From 3.5 on, `subprocess.run` is a recommended way to invoke a subprocess:
> The recommended approach to invoking subprocesses is to use the run() function for all use cases it can handle. For more advanced use cases, the underlying Popen interface can be used directly.
>
> The run() function was added in Python 3.5; if you need to retain compatibility with older versions, see the Older high-level API section.

https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module

dthybert: indeed I also prefer to use run than check_call

scripts/pipeline/orthology_benchmark.py
ipilizota: I like your idea Thomas 🙂  I had thought about it but very briefly. Now when I see all the pros listed at one place, I think I have to do it 🙂  Thank you very much 👍

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,225 @@
...
+            (["gallus_gallus", "homo_sapiens"], "default", does_not_raise()),
+            (["homo_sapiens", "zea_mays"], "default", raises(FileExistsError)),
+            (["felis_catus", "zea_mays"], "test", raises(RuntimeError,
+                    match=r"No cores found for the species set 'test' on the specified host."))
+        ]
+    )
+    def test_dump_genomes(self, species_list: List[str], species_set_name: str,
+                          tmp_dir: Path, expectation: ContextManager) -> None:
+        """Tests :func:`orthology_benchmark.dump_genomes()` function
+        when connection to the server host and port can be established.
```
twalsh-ebi, line 102: As mentioned by @JAlvarezJarreta, it is recommended to make the first line of a docstring self-contained. Here is one alternative wording to bring this one all on a single line:
```suggestion
        """Tests :func:`orthology_benchmark.dump_genomes()` when server connection can be established.
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,225 @@
...
+            out_files = tmp_dir / species_set_name
+            # pylint: disable-next=no-member
+            exp_out = pytest.files_dir / "orth_benchmark"  # type: ignore[attr-defined]
+            for db_name in self.core_dbs:
+                assert file_cmp(Path(out_files, os.environ['USER'] + "_" + db_name + ".fasta"),
+                                Path(exp_out, db_name + ".fasta"))
+
+    def test_dump_genomes_fake_connection(self, tmp_dir: Path) -> None:
+        """Tests :func:`orthology_benchmark.dump_genomes()` function
+        when provided fake server connection details.
```
twalsh-ebi, line 127: Another docstring summary line suggestion:
```suggestion
        """Tests :func:`orthology_benchmark.dump_genomes()` with fake server details.
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,225 @@
...
+    ]
+)
+def test_find_latest_core(core_names: List[str], exp_output: str, expectation: ContextManager) -> None:
+    """Tests :func:`orthology_benchmark.find_latest_core()` function.
+
+    Args:
+        core_names: A list of core database names.
+        exp_output: Expected return value of the function.
+        expectation: Context manager for the expected exception, i.e. the test will only pass if that
+                exception is raised. Use :class:`~contextlib.nullcontext` if no exception is expected.
```
twalsh-ebi, line 154: Tweak to 4-space hanging indent:
```suggestion
        expectation: Context manager for the expected exception, i.e. the test will only pass if that
            exception is raised. Use :class:`~contextlib.nullcontext` if no exception is expected.
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,225 @@
...
+    def test_get_core_names(self, species_names: List[str], exp_output: Dict[str, str],
+                            expectation: ContextManager) -> None:
+        """Tests :func:`orthology_benchmark.get_core_names()` function
+        when connection to the server host can be established.
+
+        Args:
+            species_names: Species (genome) names.
+            exp_output: Expected return value of the function.
+            expectation: Context manager for the expected exception, i.e. the test will only pass if that
+                    exception is raised. Use :class:`~contextlib.nullcontext` if no exception is expected.
```
twalsh-ebi, line 212: Tweak to 4-space hanging indent:
```suggestion
            expectation: Context manager for the expected exception, i.e. the test will only pass if that
                exception is raised. Use :class:`~contextlib.nullcontext` if no exception is expected.
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,225 @@
...
+                          tmp_dir: Path, expectation: ContextManager) -> None:
+        """Tests :func:`orthology_benchmark.dump_genomes()` function
+        when connection to the server host and port can be established.
+
+        Args:
+            species_list: A list of species (genome names).
+            species_set_name: Species set (collection) name.
+            tmp_dir: Unit test temp directory (fixture).
+            expectation: Context manager for the expected exception, i.e. the test will only pass if that
+                    exception is raised. Use :class:`~contextlib.nullcontext` if no exception is expected.
```
twalsh-ebi, line 109: Tweak to 4-space hanging indent:
```suggestion
            expectation: Context manager for the expected exception, i.e. the test will only pass if that
                exception is raised. Use :class:`~contextlib.nullcontext` if no exception is expected.
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,225 @@
...
+             "mus_musculus": os.environ['USER'] + "_" + "mus_musculus_core_106_39", "zea_mays": ""},
+             does_not_raise()),
+            ([], None, raises(RuntimeError,
+                              match=r"Empty list of species names. Cannot search for core databases."))
+        ]
+    )
+    def test_get_core_names(self, species_names: List[str], exp_output: Dict[str, str],
+                            expectation: ContextManager) -> None:
+        """Tests :func:`orthology_benchmark.get_core_names()` function
+        when connection to the server host can be established.
```
twalsh-ebi, line 206: Another docstring summary line suggestion:
```suggestion
        """Tests :func:`orthology_benchmark.get_core_names()` when server connection can be established.
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,225 @@
...
+        """
+        with expectation:
+            host = self.core_dbs["mus_musculus_core_106_39"].dbc.host
+            port = self.core_dbs["gallus_gallus_core_99_6"].dbc.port
+            # user "travis" hardcoded until we find a better solution
+            assert orthology_benchmark.get_core_names(species_names, host, port, "travis") == exp_output
+
+    def test_get_core_names_fake_connection(self) -> None:
+        """Tests :func:`orthology_benchmark.get_core_names()` function
+        when provided fake server connection details."""
```
twalsh-ebi, line 223: Another docstring summary line suggestion:
```suggestion
        """Tests :func:`orthology_benchmark.get_core_names()` with fake server details."""
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,225 @@
...
+        """Tests :func:`orthology_benchmark.dump_genomes()` function
+        when provided fake server connection details.
+
+        Args:
+            tmp_dir: Unit test temp directory (fixture).
+
+        """
+        with raises(sqlalchemy.exc.OperationalError):
+            orthology_benchmark.dump_genomes(["mus_musculus", "naja_naja"], "default",
+                                             "fake-host", 666, "compara", tmp_dir)
```
twalsh-ebi, line 135: The "fake" port number used here is an actual reserved TCP/UDP port number. For an "unreal" port number, I'd suggest `65536`.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,134 @@
...
+        raise RuntimeError("Empty list of species names. Cannot search for core databases.")
+
+    core_names = {}
+
+    eng = create_engine("mysql://" + user + "@" + host + ":" + str(port) + "/")
+    out_tmp = eng.execute("SHOW DATABASES LIKE '%%_core_%%'").fetchall()
+    all_cores = [i[0] for i in out_tmp]
+
+    for species in species_names:
+        core_name = [core for core in all_cores if species + "_core_" in core]
```
twalsh-ebi, line 139: Consider using f-string and `startswith` method:
```suggestion
        core_name = [core for core in all_cores if core.startswith(f"{species}_core_")]
```

twalsh-ebi (changes requested):
Overall this is a nice and well-structured code update. I've just made a few general comments and suggestions. One specific change that I'm requesting is to look at ways to scale down the test databases added in this PR, as I feel they could be smaller.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,134 @@
...
+
+    """
+    cores = get_core_names(species_list, host, port, user)
+    dump_cores = [core for species, core in cores.items() if core != ""]
+
+    if len(dump_cores) == 0:
+        raise RuntimeError(f"No cores found for the species set '{species_set_name}' on the specified host.")
+
+    dumps_dir = os.path.join(out_dir, species_set_name)
+    os.mkdir(dumps_dir)
```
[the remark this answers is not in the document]
ipilizota: I'd like to throw an error if `dumps_dir` already exists to prevent messing up its current content.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,134 @@
...
+    """Returns the latest core database among the available ones.
+
+    Args:
+        core_names: A list of cores for a species of interest.
+
+    Raises:
+        RuntimeError: If `core_names` is empty.
+
+    Returns:
+        Name of the latest core database (the latest Ensembl release, the latest version).
```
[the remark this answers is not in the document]
ipilizota: Very good point! For example, there are `caenorhabditis_elegans_core_106_279` and `caenorhabditis_elegans_core_53_106_279` at the `vertannot` at the moment. So, one name has 2 numbers and one has 3 numbers. In such cases, I should add zero(s) to the names with fewer numbers. If there were only e.g. `caenorhabditis_elegans_core_53_106_269` and `caenorhabditis_elegans_core_53_106_279`, the function would return `caenorhabditis_elegans_core_53_106_279`.

src/python/tests/test_orthology_benchmark.py
ipilizota: Changed it at both places 😄  Thank you 👍

scripts/pipeline/orthology_benchmark.py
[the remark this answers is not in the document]
JAlvarezJarreta: Allegedly, those two should contain the same genome information (which is what matters here), only the `meta` table differs (based on what I was told), so it shouldn't really matter. Maybe you can say something like:
> (the latest Ensembl release, the latest (non-EG) version).

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,134 @@
...
+
+    """
+    if len(core_names) == 0:
+        raise RuntimeError("Empty list of core databases. Cannot determine the latest one.")
```
JAlvarezJarreta, line 90: I'm prone to say that if the list is empty you can just return an empty string (and maybe a warning if you want to alert the user?). In a way, you would still match the description of the `Return` statement.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,134 @@
...
+
+    for core in dump_cores:
+        out_file = os.path.join(dumps_dir, f"{core}.fasta")
+        subprocess.run([script, "-core-db", core, "-host", host, "-port", str(port),
+                       "-outfile", out_file], capture_output=True, check=True)
+
+
+def find_latest_core(core_names: List[str]) -> str:
+    """Returns the latest core database among the available ones.
+
```
JAlvarezJarreta, line 78: Just in case, it may be worth adding the expected format of the core database names?
```suggest
    The core database names are expected in the format:
    ``<species-name>_core_(<eg-rel>_)<ensembl-rel>_<version>``, e.g. ``erinaceus_europaeus_core_106_1``.
```

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,134 @@
...
+    while i < n_cols:
+        # Find max value in the i-th column
+        # For the next iteration (i+1) consider only rows where i-th column == max value
+        max_col = numpy.amax(rel_ver_arr[:, i])
+        rows_ind = numpy.where(rel_ver_arr[:, i] == max_col)[0]
+        rel_ver_arr = rel_ver_arr[[rows_ind], :][0]
+        i += 1
+
+    latest_rel_ver = '_'.join(map(str, rel_ver_arr[0]))
+    core_name = [core for core in core_names if latest_rel_ver in core][0]
```
JAlvarezJarreta, line 107: Although it works, I find this code a little bit complex to follow in one go. This is just a suggestion to simply it and improve its readability:
```suggest
    versions = [name.split('_')[-2:] for name in core_names]
    latest = 0
    for i, row in enumerate(versions[1:], 1):
        if (row[0] > versions[latest][0] or (row[0] == versions[latest][0] and row[1] > versions[latest][1])):
            latest = i
    core_name = core_names[latest]
```
(hope I haven't missed any functionality)

scripts/pipeline/orthology_benchmark.py
```diff
@@ -52,11 +168,16 @@ def calculate_goc_scores():
     parser = argparse.ArgumentParser()
     parser.add_argument("--mlss_conf", required=True, type=str, help="Path to MLSS configuration XML file")
```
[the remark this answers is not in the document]
JAlvarezJarreta: It is not (it is the default type), but I don't mind it in this case for some reason :man_shrugging:

scripts/pipeline/orthology_benchmark.py
```diff
@@ -52,11 +168,16 @@ def calculate_goc_scores():
...
     parser.add_argument("--species_set", required=True, type=str, help="Species set (collection) name")
+    parser.add_argument("--host", required=True, type=str, help="Database host")
+    parser.add_argument("--port", required=True, type=int, help="Database port")
+    parser.add_argument("--user", required=True, type=str, help="Server username")
+    parser.add_argument("--out_dir", required=True, type=str,
+                        help="Location for'species_set/core_name.fasta' dumps")
```
JAlvarezJarreta, line 175:
```suggestion
                        help="Location for 'species_set/core_name.fasta' dumps")
```

JAlvarezJarreta (changes requested):
I have `test_orthology_benchmark.py` left to review, but to give some thoughts/ideas in the meantime.

scripts/pipeline/orthology_benchmark.py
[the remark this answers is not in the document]
ipilizota: The problem is (or was because I hope I fixed it 🙂 ) that the code would hit `IndexError` for `n_cols = rel_ver_arr.shape[1]`.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,134 @@
...
+        raise RuntimeError("Empty list of species names. Cannot search for core databases.")
+
+    core_names = {}
+
+    eng = create_engine(f"mysql://{user}@{host}:{port}/")
+    out_tmp = eng.execute("SHOW DATABASES LIKE '%%_core_%%'").fetchall()
+    all_cores = [i[0] for i in out_tmp]
+
+    for species in species_names:
+        core_name = [core for core in all_cores if core.startswith(f"{species}_core_")]
```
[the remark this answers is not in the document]
ipilizota: Yes, this was exactly to avoid strains, breeds, cultivars 😄

scripts/pipeline/orthology_benchmark.py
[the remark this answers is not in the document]
ipilizota: I put it there more or less to keep consistency across all arguments 😺

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,134 @@
...
+
+import numpy
+from sqlalchemy import create_engine
 
 from ensembl.compara.config import get_species_set_by_name
 
 
-def dump_genomes():
-    """Docstring"""
+def dump_genomes(species_list: List[str], species_set_name: str, host: str, port: int,
```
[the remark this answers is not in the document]
ipilizota: Thank you both 👍  I'll keep `Iterable[str]` in mind for my future code. I find it easier to immediately interpret its meaning compared to `typing.Sequence[str]`.

scripts/pipeline/orthology_benchmark.py
ipilizota: I just saw this. Hm hm, I've read it twice and couldn't understand it 👼  Out of curiosity, what would it return in case of core versions `105_279, 52_105_279, 106_3`?

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,224 @@
...
+# pylint: enable=import-error,wrong-import-position
+
+from ensembl.compara.filesys import file_cmp
+
+
+@pytest.mark.parametrize(
+    "multi_dbs",
+    [
+        [{'src': 'core/gallus_gallus_core_99_6', 'name': 'gallus_gallus_core_99_6'},
+         {'src': 'core/homo_sapiens_core_99_38', 'name': 'homo_sapiens_core_99_38'}]
```
JAlvarezJarreta, line 64: I had to remind myself how it worked, but based on the code, what you have written is equivalent to:
```suggest
        [{'src': 'core/gallus_gallus_core_99_6'}, {'src': 'core/homo_sapiens_core_99_38'}]
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,224 @@
...
+)
+class TestDumpGenomes:
+    """Tests :func:`orthology_benchmark.dump_genomes()` function.
+
+    Attributes:
+        core_dbs: A set of test core databases.
+
+    """
+
+    core_dbs = {} # type: Dict
```
JAlvarezJarreta, line 76: PEP8 (IIRC):
```suggestion
    core_dbs = {}  # type: Dict
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,224 @@
...
+            tmp_dir: Unit test temp directory (fixture).
+            expectation: Context manager for the expected exception, i.e. the test will only pass if that
+                exception is raised. Use :class:`~contextlib.nullcontext` if no exception is expected.
+
+        """
+        with expectation:
+            host = self.core_dbs["gallus_gallus_core_99_6"].dbc.host
+            port = self.core_dbs["gallus_gallus_core_99_6"].dbc.port
+            # user "travis" hardcoded until we find a better solution
+            orthology_benchmark.dump_genomes(species_list, species_set_name, host, port, "travis", tmp_dir)
```
JAlvarezJarreta, line 115:
```suggestion
Based on how `UnitTestDB` works, I think this is where Travis is failing:
```suggestion
        user = os.environ['USER']
        with expectation:
            host = self.core_dbs[f"{user}_gallus_gallus_core_99_6"].dbc.host
            port = self.core_dbs[f"{user}_gallus_gallus_core_99_6"].dbc.port

            orthology_benchmark.dump_genomes(species_list, species_set_name, host, port, user, tmp_dir)
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,225 @@
...
+
+            out_files = tmp_dir / species_set_name
+            # pylint: disable-next=no-member
+            exp_out = pytest.files_dir / "orth_benchmark"  # type: ignore[attr-defined]
+            for db_name in self.core_dbs:
+                assert file_cmp(Path(out_files, os.environ['USER'] + "_" + db_name + ".fasta"),
+                                Path(exp_out, db_name + ".fasta"))
```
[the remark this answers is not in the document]
JAlvarezJarreta: Based on my previous comment, I would replace `os.environ['USER']` by `user`.

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def test_dump_genomes_fake_connection(self, tmp_dir: Path) -> None:
```
JAlvarezJarreta, line 124: Nice! :+1:

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+@pytest.mark.parametrize(
+    "core_names, exp_output, expectation",
+    [
+        (["mus_musculus_core_105_1", "mus_musculus_core_52_105_3", "mus_musculus_core_104_4"],
+         "mus_musculus_core_52_105_3", does_not_raise()),
+        # Just to test trimming after padding; will remove it when Travis passes:
+        (["mus_musculus_core_106_1", "mus_musculus_core_52_105_3", "mus_musculus_core_51_104_4"],
+         "mus_musculus_core_106_1", does_not_raise()),
```
JAlvarezJarreta, line 143:
```suggestion
```
(a clear reminder ;) )

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,224 @@
...
+    with expectation:
+        assert orthology_benchmark.find_latest_core(core_names) == exp_output
+
+
+@pytest.mark.parametrize(
+    "multi_dbs",
+    [
+        [{'src': 'core/danio_rerio_core_105_11', 'name': 'danio_rerio_core_105_11'},
+         {'src': 'core/mus_musculus_cbaj_core_107_1', 'name': 'mus_musculus_cbaj_core_107_1'},
+         {'src': 'core/mus_musculus_core_106_39', 'name': 'mus_musculus_core_106_39'}]
```
JAlvarezJarreta, line 167: Same as before:
```suggestion
        [{'src': 'core/danio_rerio_core_105_11'},
         {'src': 'core/mus_musculus_cbaj_core_107_1'},
         {'src': 'core/mus_musculus_core_106_39'}]
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,224 @@
...
+)
+class TestGetCoreNames:
+    """Tests :func:`orthology_benchmark.get_core_names()` function.
+
+    Attributes:
+        core_dbs: A set of test core databases.
+
+    """
+
+    core_dbs = {} # type: Dict
```
JAlvarezJarreta, line 179: Same as before:
```suggestion
    core_dbs = {}  # type: Dict
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,224 @@
...
+
+        """
+        type(self).core_dbs = multi_dbs
+
+    @pytest.mark.parametrize(
+        "species_names, exp_output, expectation",
+        [
+            (["danio_rerio", "mus_musculus", "zea_mays"],
+             {"danio_rerio": f"{os.environ['USER']}_danio_rerio_core_105_11",
+             "mus_musculus": f"{os.environ['USER']}_mus_musculus_core_106_39", "zea_mays": ""},
```
JAlvarezJarreta, line 198:
```suggestion
              "mus_musculus": f"{os.environ['USER']}_mus_musculus_core_106_39", "zea_mays": ""},
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,224 @@
...
+            exp_output: Expected return value of the function.
+            expectation: Context manager for the expected exception, i.e. the test will only pass if that
+                exception is raised. Use :class:`~contextlib.nullcontext` if no exception is expected.
+
+        """
+        with expectation:
+            host = self.core_dbs["mus_musculus_core_106_39"].dbc.host
+            port = self.core_dbs["mus_musculus_core_106_39"].dbc.port
+            # user "travis" hardcoded until we find a better solution
+            assert orthology_benchmark.get_core_names(species_names, host, port, "travis") == exp_output
```
JAlvarezJarreta, line 219:
```suggestion
        with expectation:
            user = os.environ['USER']
            host = self.core_dbs[f"{user}_mus_musculus_core_106_39"].dbc.host
            port = self.core_dbs[f"{user}_mus_musculus_core_106_39"].dbc.port
            assert orthology_benchmark.get_core_names(species_names, host, port, user) == exp_output
```

JAlvarezJarreta (changes requested):
`test_orthology_benchmark.py` reviewed. Not sure if you have noticed, but you are testing your code twice for `find_latest_core` and `get_core_names`. I would just try to integrate these parametrizations inside `TestDumpGenomes` (although most are already covered in it in any case) and then you are done :)

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,224 @@
...
+
+        """
+        type(self).core_dbs = multi_dbs
+
+    @pytest.mark.parametrize(
+        "species_list, species_set_name, expectation",
+        [
+            (["gallus_gallus", "homo_sapiens"], "default", does_not_raise()),
+            (["homo_sapiens", "zea_mays"], "default", raises(FileExistsError)),
+            (["felis_catus", "zea_mays"], "test", raises(RuntimeError,
```
JAlvarezJarreta, line 95:
```suggestion
            ([f"{os.environ['USER']}_gallus_gallus", f"{os.environ['USER']}_homo_sapiens"], "default", does_not_raise()),
            ([f"{os.environ['USER']}_homo_sapiens", f"{os.environ['USER']}_zea_mays"], "default", raises(FileExistsError)),
            ([f"{os.environ['USER']}_felis_catus", f"{os.environ['USER']}_zea_mays"], "test", raises(RuntimeError,
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,224 @@
...
+        core_dbs: A set of test core databases.
+
+    """
+
+    core_dbs = {} # type: Dict
+
+    # autouse=True makes this fixture be executed before any test_* method of this class, and scope='class' to
+    # execute it only once per class parametrization
+    @pytest.fixture(scope='class', autouse=True)
+    def setup(self, multi_dbs: Dict) -> None:
```
JAlvarezJarreta, line 81:
```suggestion
    def setup(self, request: FixtureRequest, multi_dbs: Dict) -> None:
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,224 @@
...
+        """Loads the required fixtures and values as class attributes.
+
+        Args:
+            multi_dbs: Dictionary of unit test databases (fixture).
+
+        """
+        type(self).core_dbs = multi_dbs
```
JAlvarezJarreta, line 88:
```suggestion
        type(self).core_dbs = multi_dbs
        server_url = sqlalchemy.engine.url.make_url(request.config.getoption('server'))
        type(self).host = server_url.host
        type(self).port = server_url.port
```

JAlvarezJarreta (commented):
I have just realised a couple issues with Travis and the way you were getting the host/port. Hope this way fixes everything (you may need to update some imports and docstrings, so feel free to edit everything on your side and commit it without my reference).

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,134 @@
...
+    core_names = {}
+
+    eng = create_engine("mysql://" + user + "@" + host + ":" + str(port) + "/")
+    out_tmp = eng.execute("SHOW DATABASES LIKE '%%_core_%%'").fetchall()
+    all_cores = [i[0] for i in out_tmp]
+
+    for species in species_names:
+        core_name = [core for core in all_cores if species + "_core_" in core]
+        if len(core_name) == 0:
+            core_names[species] = ""
```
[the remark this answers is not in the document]
ipilizota: This way facilitates adding functionality to e.g. output the list of species for which no cores were found so a user can search for them on another `host` or just be aware of it.

ipilizota:
Thank you @twalsh-ebi and @JAlvarezJarreta for your reviews 👍 I've been working through them 🙂 

@JAlvarezJarreta:
> Not sure if you have noticed, but you are testing your code twice for find_latest_core and get_core_names. I would just try to integrate these parametrizations inside TestDumpGenomes (although most are already covered in it in any case) and then you are done :)

Yes, I'm aware of double testing. Related to that, I call `get_core_names` from `dump_genomes`. Since it turned out that I will also need `get_core_names` for getting GTF files, I will write a ticket to refactor this code - so, the pipeline will call `get_core_names` and pipe its output to `dump_genomes` and to a function like `get_gtf_files`. This will then change the tests and naturally reduce redundancy. Would you still prefer reducing the testing redundancy within this PR/ticket?

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,145 @@
...
+        RuntimeError: If `species_list` is empty.
+        sqlalchemy.exc.OperationalError: If `user` cannot read from `host:port`.
+
+    """
+    if len(species_names) == 0:
+        raise RuntimeError("Empty list of species names. Cannot search for core databases.")
+
+    core_names = {}
+
+    engg = create_engine(f"mysql://{user}@{host}:{port}/")
```
twalsh-ebi, line 145: This looks like a typo:
```suggestion
    eng = create_engine(f"mysql://{user}@{host}:{port}/")
```

scripts/pipeline/orthology_benchmark.py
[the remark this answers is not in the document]
twalsh-ebi: Even if we omit the species in a `species_list` for which no cores were found, we can then get the set of species lacking cores as follows:
```
species_sans_cores = set(species_list) - set(cores.keys())
```

Moreover, a user must check for missing cores either way. And if a user doesn't check for a core, an omitted species-core pair will be picked up as a `KeyError` as soon as we attempt to access it. For example:
```
db_url = f"mysql://{user}@{host}:{port}/{cores['homo_sapiens']}"
```
If the returned core name is an empty string, this would still be a valid URL. The lack of a core would still cause an error, but at a later stage, when the cause of the error would be less obvious.

src/python/tests/test_orthology_benchmark.py
ipilizota: Nope. Travis fails so I reverted back to `[{'src': 'core/gallus_gallus_core_99_6', 'name': 'gallus_gallus_core_99_6'}, {'src': 'core/homo_sapiens_core_99_38', 'name': 'homo_sapiens_core_99_38'}]`. I think the major problem is that omitting `'name'` leads to test dbs with names `gallus_gallus, homo_sapiens`, i.e. without release and version info.

ipilizota: OK, Travis fails again so I probably misinterpreted its messages. Let me try out your other suggestions and I will get back to this in the end.

JAlvarezJarreta:
> Yes, I'm aware of double testing. Related to that, I call `get_core_names` from `dump_genomes`. Since it turned out that I will also need `get_core_names` for getting GTF files, I will write a ticket to refactor this code - so, the pipeline will call `get_core_names` and pipe its output to `dump_genomes` and to a function like `get_gtf_files`. This will then change the tests and naturally reduce redundancy. Would you still prefer reducing the testing redundancy within this PR/ticket? I'd prefer to do it only once, i.e. in the next ticket.

I would keep a personal copy of the specific ones then for when these functions are moved elsewhere, and commit in this PR only the unit test for `dump_genomes` then. However, in this PR the parametrization should cover all functions' code, whilst in the coming one this unit test should then only cover the `dump_genome` code (i.e. it will need to be updated).

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,224 @@
...
+            tmp_dir: Unit test temp directory (fixture).
+            expectation: Context manager for the expected exception, i.e. the test will only pass if that
+                exception is raised. Use :class:`~contextlib.nullcontext` if no exception is expected.
+
+        """
+        with expectation:
+            host = self.core_dbs["gallus_gallus_core_99_6"].dbc.host
+            port = self.core_dbs["gallus_gallus_core_99_6"].dbc.port
+            # user "travis" hardcoded until we find a better solution
+            orthology_benchmark.dump_genomes(species_list, species_set_name, host, port, "travis", tmp_dir)
```
JAlvarezJarreta, line 115: Based on my previous comment:
```suggestion
            orthology_benchmark.dump_genomes(species_list, species_set_name, self.host, self.port,
                                             self.username, tmp_dir)
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,224 @@
...
+from importlib.util import module_from_spec, spec_from_file_location
+import os
+from pathlib import Path
+import sys
+from typing import ContextManager, Dict, List
+
+import sqlalchemy
+
+import pytest
+from pytest import raises
```
JAlvarezJarreta, line 35:
```suggestion
from pytest import FixtureRequest, raises
```
(see following comment in `setup()`)

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,224 @@
...
+            port = self.core_dbs["gallus_gallus_core_99_6"].dbc.port
+            # user "travis" hardcoded until we find a better solution
+            orthology_benchmark.dump_genomes(species_list, species_set_name, host, port, "travis", tmp_dir)
+
+            out_files = tmp_dir / species_set_name
+            # pylint: disable-next=no-member
+            exp_out = pytest.files_dir / "orth_benchmark"  # type: ignore[attr-defined]
+            for db_name in self.core_dbs:
+                assert file_cmp(Path(out_files, os.environ['USER'] + "_" + db_name + ".fasta"),
+                                Path(exp_out, db_name + ".fasta"))
```
JAlvarezJarreta, line 122: Following my previous comments and @twalsh-ebi's:
```suggestion
            for db_name in self.core_dbs:
                test_db_name = self.core_dbs[db_name].dbc.db_name
                assert file_cmp(out_files / f"{test_db_name}.fasta", exp_out / f"{db_name}.fasta")
```

src/python/tests/test_orthology_benchmark.py
ipilizota: I see in my current (local) code `from _pytest.fixtures import FixtureRequest` which I C/P-ed from our code base. What's the difference?

src/python/tests/test_orthology_benchmark.py
ipilizota: Yes, I'm doing that 🙂

src/python/tests/test_orthology_benchmark.py
ipilizota: Aaaaand it fails... 🙄

src/python/tests/test_orthology_benchmark.py
JAlvarezJarreta: It seems to be an update in `pytest` v6, which looks nicer, IMO (but doesn't really matter).

ipilizota: Aha, I see. I also prefer less convoluted solutions so will go with the suggested one 🙂

scripts/pipeline/orthology_benchmark.py
JAlvarezJarreta: My code just cares about the release version (ignoring EG) and core version, so in your list, it would return the `107_2` one.

scripts/pipeline/orthology_benchmark.py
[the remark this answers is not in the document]
twalsh-ebi: On reflection, though I'd still advocate the alternative approach of omitting cases where cores are not found, the existing approach works fine within the context of the script. But I'd suggest at least updating the function docstring to document this behaviour, so that there are no surprises.

src/python/tests/test_orthology_benchmark.py
ipilizota: I decided to keep `'name'` for the sake of clarity and as a help for fellow colleagues and future generations who will use this script as a template for their tests.

scripts/pipeline/orthology_benchmark.py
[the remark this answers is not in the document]
ipilizota: I'll omit adding `""` to the `core_names` 👍  Thank you 🙂

scripts/pipeline/orthology_benchmark.py
ipilizota: Similar/same logic to another point that we discussed earlier so I'll just C/P the comment: If there are no dumps, there is no input data for the orthology tools. I'd like to alarm the user ASAP. Perhaps they actually misspecified `--species_set` in the command line when running the pipeline.

ipilizota:
I had a better look at potential redundancies in testing and I think the tests are actually ok in their current state. Similar (but distinctive) situations and messages might trick a reader. Also, functions which first check if their input is an empty list although that is not possible given the rest of the code in this pipeline, have that step in case they're used outside this pipeline or if the pipeline changes along the way.

Relevant bits from `orthology_benchmark.py`: 
- `get_core_names()`
```
if len(species_names) == 0:
    raise ValueError("Empty list of species names. Cannot search for core databases.")
```
This is a situation when `get_core_names()` gets an empty list as input. 
- `find_latest_core()`
```
if len(core_names) == 0:
    raise ValueError("Empty list of core databases. Cannot determine the latest one.")
```
This is a situation when `find_latest_core()` gets an empty list as input
- `dump_genomes()`
```
cores = get_core_names(species_list, host, port, user)
dump_cores = [core for species, core in cores.items()]
if len(dump_cores) == 0:
    raise RuntimeError(f"No cores found for the species set '{species_set_name}' on the specified host.")
```
This is a situation when `get_core_names()` returned an empty dictionary.

ipilizota:
Travis is failing for reasons unrelated to this PR. Tests pass on Codon.

src/python/tests/test_orthology_benchmark.py
JAlvarezJarreta: Same as before.

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,234 @@
...
+
+    @pytest.mark.parametrize(
+        "species_names, exp_output, expectation",
+        [
+            (["danio_rerio", "mus_musculus", "zea_mays"],
+             {"danio_rerio": os.environ['USER'] + "_danio_rerio_core_105_11",
+              "mus_musculus": os.environ['USER'] + "_mus_musculus_core_106_39"},
+             does_not_raise()),
+            ([], None, raises(ValueError,
+                              match=r"Empty list of species names. Cannot search for core databases."))
```
JAlvarezJarreta, line 213:
```suggestion
            (
                ["danio_rerio", "mus_musculus", "zea_mays"],
                {"danio_rerio": os.environ['USER'] + "_danio_rerio_core_105_11",
                 "mus_musculus": os.environ['USER'] + "_mus_musculus_core_106_39"},
                does_not_raise()
            ),
            ([], None, raises(ValueError,
                              match=r"Empty list of species names. Cannot search for core databases."))
```

JAlvarezJarreta (changes requested):
We are almost there :tada:

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,145 @@
...
+    """Returns the latest core database names for a list of species.
+
+    Args:
+        species_names: Species (genome) names.
+        host: Host for core databases.
+        port: Host port.
+        user: Server username.
+
+    Returns:
+        Dictionary mapping species (genome) names to the latest version of available core names.
```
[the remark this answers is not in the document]
ipilizota: It also adds info on what the keys and values represent.

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,234 @@
...
+        Args:
+            request: Access to the requesting test context.
+            multi_dbs: Dictionary of unit test databases (fixture).
+
+        """
+        type(self).core_dbs = multi_dbs
+        server_url = sqlalchemy.engine.url.make_url(request.config.getoption('server'))
+        type(self).host = server_url.host
+        type(self).port = server_url.port
+        type(self).username = "ensro" if server_url.username == "ensadmin" else server_url.username
```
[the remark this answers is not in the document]
ipilizota: I copy-pasted this line from your suggestions to resolve the issues we discussed at the time. Let me know when there is a better solution.

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,234 @@
...
+        """
+        with expectation:
+            orthology_benchmark.dump_genomes(species_list, species_set_name, self.host, self.port,
+                                             self.username, tmp_dir)
+
+            out_files = tmp_dir / species_set_name
+            # pylint: disable-next=no-member
+            exp_out = pytest.files_dir / "orth_benchmark"  # type: ignore[attr-defined]
+            for db_name, unittest_db in self.core_dbs.items():
+                assert file_cmp(out_files / f"{unittest_db.dbc.db_name}.fasta", exp_out / f"{db_name}.fasta")
```
twalsh-ebi, line 128: Forgive me if I'm asking this a little late in the review process, but have you considered using the shorter extension `.fa` for the FASTA files? With the shorter extension, this line would be a little less of a squeeze.

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,234 @@
...
+"""
+
+from contextlib import nullcontext as does_not_raise
+from importlib.abc import Loader
+from importlib.machinery import ModuleSpec
+from importlib.util import module_from_spec, spec_from_file_location
+import os
+from pathlib import Path
+import sys
+from typing import ContextManager, Dict, List
```
twalsh-ebi, line 30: You'd need to import `Optional` here if you decide to use it later in the script.
```suggestion
from typing import ContextManager, Dict, List, Optional
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,234 @@
...
+
+    Attributes:
+        core_dbs: A set of test core databases.
+
+    """
+
+    core_dbs = {}  # type: Dict
+    host = None  # type: str
+    port = None  # type: int
+    username = None  # type: str
```
twalsh-ebi, line 78: These variables might be considered `Optional` because they can be set to `None`. But `mypy` seems happy with or without them, so I'll leave it to your discretion.
```suggestion
    host = None  # type: Optional[str]
    port = None  # type: Optional[int]
    username = None  # type: Optional[str]
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,234 @@
...
```
twalsh-ebi, line 184: Like before, this suggestion is entirely optional:
```suggestion
    host = None  # type: Optional[str]
    port = None  # type: Optional[int]
    username = None  # type: Optional[str]
```

twalsh-ebi (approved):
The code is really shaping up. I only have a few optional suggestions/questions, and I'm happy to leave it up to you how and whether to address them.

src/python/tests/test_orthology_benchmark.py
ipilizota: I came across `.fasta` files on Noah/Codon so decided to keep it long. The extension actually comes form `orthology_benchmark.py`, line 72.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,145 @@
...
+
+    Raises:
+        FileExistsError: If directory `out_dir/species_set_name` already exists.
+        sqlalchemy.exc.OperationalError: If `user` cannot read from `host:port`.
+        RuntimeError: If no cores are found for any species in `species_list` (`species_set_name`)
+            on `host:port`.
+
+    """
+    cores = get_core_names(species_list, host, port, user)
+    dump_cores = [core for species, core in cores.items()]
```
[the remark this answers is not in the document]
ipilizota: Thanks 🙂 I think `key, value` has to be used. An example that I googled:
```
>>> a_dict = {'color': 'blue', 'fruit': 'apple', 'pet': 'dog'}
>>> for item in a_dict.items():
...     print(item)
...
('color', 'blue')
('fruit', 'apple')
('pet', 'dog')
```

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,145 @@
...
+
+    core_names = {}
+
+    eng = create_engine(f"mysql://{user}@{host}:{port}/")
+    result = eng.execute("SHOW DATABASES LIKE '%%_core_%%'").fetchall()
+    all_cores = [i[0] for i in result]
+
+    user_env = os.environ['USER']
+    for species in species_names:
+        core_name = [core for core in all_cores if re.match(f"^({user_env}_)?{species}_core_", core)]
```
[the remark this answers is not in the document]
ipilizota: That's the only reason I can think of. @JAlvarezJarreta wanted to have that in so he might have more scenarios in mind. I can reassure you that `ivana_pilizoticus_core_107_1` won't be ready for the handover 😞

scripts/pipeline/orthology_benchmark.py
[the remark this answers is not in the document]
twalsh-ebi: Have you considered using [dict.values](https://docs.python.org/3.9/library/stdtypes.html#dict.values)?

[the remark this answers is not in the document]
ipilizota: I actually went with `.items()` because `pylint` suggested so 😁

scripts/pipeline/orthology_benchmark.py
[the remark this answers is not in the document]
twalsh-ebi: I'm sorry to hear it. 🙂 In any case, I suppose the need for testing is sufficient reason for the updated regex.

src/python/tests/test_orthology_benchmark.py
twalsh-ebi: There are also examples of files using `.fa` to be found (e.g. [here](https://ftp.ensembl.org/pub/release-104/fasta/), but both extensions are perfectly fine, so I'll leave it up to you.

scripts/pipeline/orthology_benchmark.py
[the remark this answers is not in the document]
twalsh-ebi: I'm not one to trifle with what `pylint` suggests, but why not try using `cores.values()` and see if that's something `pylint` accepts?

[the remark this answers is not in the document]
ipilizota: Done. Passes `pylint` locally 🙂 Thank you!

src/python/tests/test_orthology_benchmark.py
ipilizota: Moved to `.fa`. `pytest` passes locally.

scripts/pipeline/orthology_benchmark.py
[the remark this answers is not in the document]
JAlvarezJarreta: Not only testing. We may have a situation where we copy cores for our usage in `cp?` and those should always have our user prefixed, so I envisioned it may be useful to pick those as well.

src/python/tests/test_orthology_benchmark.py
JAlvarezJarreta: Same as before.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,145 @@
...
+    Args:
+        core_names: A list of cores for a species of interest.
+
+    Raises:
+        ValueError: If `core_names` is empty.
+
+    Returns:
+        Name of the latest core database (the latest Ensembl release, the latest version).
+
+    """
```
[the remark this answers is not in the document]
ipilizota: Must've been by mistake.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -17,18 +17,145 @@
...
+        # For the next iteration (i+1) consider only rows where i-th column == max value
+        max_col = numpy.amax(rel_ver_arr[:, i])
+        rows_ind = numpy.where(rel_ver_arr[:, i] == max_col)[0]
+        rel_ver_arr = rel_ver_arr[[rows_ind], :][0]
+        i += 1
+
+    # Trim padded zeros, if any, to get only numbers that appear in the core db name
+    rel_ver_trimmed = numpy.trim_zeros(rel_ver_arr[0], "f")
+    latest_rel_ver = '_'.join(map(str, rel_ver_trimmed))
+    core_name = [core for core in core_names if latest_rel_ver in core][0]
```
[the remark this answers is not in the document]
ipilizota: I find your code convoluted and less readable than mine. As you say, you proposed it, and I'll add - not imposed it 🙂

src/python/tests/test_orthology_benchmark.py
ipilizota: Aha, this one wasn't by mistake. Basically, I'm/was wondering why `(` and `)` should be singletons in lines 206 and 211 - this is perhaps the only such case in our code base. In all my `pytest` scripts, for none of test cases they are in their own lines.

JAlvarezJarreta: It is a matter of block readability: if you don't pay attention, one could mix like 210 as part of the block started in line 206, so "sacrificing" two lines for that seems as reasonable as when you put empty lines within the code in your functions.

JAlvarezJarreta (approved):
It has taken quite some time, but it is finally done :tada:

ipilizota:
Oh, happy day 🎉  Thank you all!

PR #365, opened 2021-11-04
Module for dealing with orthology tools

ipilizota:
## Description

Created a new module for dealing with orthology tools. Added functions necessary for running OrthoFinder.

**Related JIRA tickets:**
- ENSCOMPARASW-4949
- ENSCOMPARASW-4953

#### Change 1
- New module with functions to: 1) prepare input for OrthoFinder, and 2) run OrthoFinder.  

#### Change 2
- Corresponding `pytest` and test files. Looks like 

## Testing
I also did:
```
from ensembl.compara.orthtools import prepare_input_orthofinder
source_dir="/hps/nobackup/flicek/ensembl/compara/ivana/test_test"
target_dir="/hps/nobackup/flicek/ensembl/compara/ivana/test_ofinder"
prepare_input_orthofinder(source_dir, target_dir)

from ensembl.compara.orthtools import run_orthofinder
run_orthofinder(target_dir)
run_orthofinder(target_dir, 16, 4)
```
The outputs are in `/hps/nobackup/flicek/ensembl/compara/ivana/test_ofinder/OrthoFinder/`. Looks ok. According to commands in `Log.txt` files, the number of threads was correctly assigned.

## Notes
Needed for `orthology_benchmark.py` pipeline.


src/python/lib/ensembl/compara/orthtools/orthofinder.py
```diff
@@ -0,0 +1,82 @@
...
+    run_orthofinder('path/to/input/fasta/files', 16, 4)
+
+"""
+
+__all__ = ['prepare_input_orthofinder', 'run_orthofinder']
+
+import os
+import subprocess
+
+def prepare_input_orthofinder(source_dir: str, target_dir: str) -> None:
```
CristiGuijarro, line 42: https://github.com/Ensembl/ensembl-compara/blob/feature/homology_annotation/scripts/pipeline/symlink_fasta.py <- just spotted this, and wanted to direct you to this existing script. Not saying you have to use it - and perhaps this does not meet the specs, but a notable difference is that here you duplicate the data by `rsync` -ing the directory whilst for orthofinder, you could use symlinks and save the duplication, OrthoFinder will just write to the directory given with symlinked files/dir and use the real paths to utilise the data.

ipilizota: Great, thank you! I will swap copying files for creating symlinks 👍

src/python/lib/ensembl/compara/orthtools/orthofinder.py
```diff
@@ -0,0 +1,92 @@
...
+# limitations under the License.
+"""Methods for preparing input files and running OrthoFinder.
+
+This module provides::
+
+    :func:`prepare_input_orthofinder()` which copies input fasta files for OrthoFinder
+            into a specified working directory
+
+    :func:`run_orthofinder()` which runs OrthoFinder
+
```
JAlvarezJarreta, line 23: I find this too verbose: the documentation will automatically pull each function's docstring during its generation. However, this is just my personal preference (but bear in mind this would be required to be maintained and updated).

src/python/lib/ensembl/compara/orthtools/orthofinder.py
```diff
@@ -0,0 +1,92 @@
...
+Typical usage example::
+
+    from ensembl.compara.orthtools import prepare_input_orthofinder
+    prepare_input_orthofinder('path/to/source_dir', 'path/to/target_dir')
+
+    from ensembl.compara.orthtools import run_orthofinder
+    run_orthofinder('path/to/input/fasta/files')
+
+    from ensembl.compara.orthtools import run_orthofinder
+    run_orthofinder('path/to/input/fasta/files', 16, 4)
```
JAlvarezJarreta, line 33: Since these two are linked, what about:
```suggestion
    >>> from ensembl.compara.orthtools import prepare_input_orthofinder, run_orthofinder
    >>> prepare_input_orthofinder('path/to/src_dir', 'path/to/dst_dir')
    >>> run_orthofinder('path/to/dst_dir')
    # Alternatively, you can pass additional arguments to OrthoFinder:
    >>> run_orthofinder('path/to/dst_dir', t=16, a=4)
```
I have found examples that use `>>>` when it is specific "code", and without it when they are command lines. Furthermore, this nomenclature allows you to show each line's output, e.g.:
```
>>> print('hello')
hello
>>> exit()
```
I will try to track these down and update them accordingly in my next Python code review.

src/python/lib/ensembl/compara/orthtools/orthofinder.py
```diff
@@ -0,0 +1,92 @@
...
+
+"""
+
+__all__ = ['prepare_input_orthofinder', 'run_orthofinder']
+
+import fnmatch
+import os
+from pathlib import Path
+import subprocess
+
```
JAlvarezJarreta, line 43: PEP8:
```suggestion


```

JAlvarezJarreta (changes requested):
If `symlink_fasta.py` is required by several projects I would recommend for it to move into `master` and then every project can benefit from a single copy. I'm extremely reluctant on approving a PR with a script copied from another project (code duplication within our own code base is a bad idea).

I'm struggling to find the need of `orthtools/orthofinder.py` as a module: both functions are just wrappers (no additional logic). I could understand a wrapper script for `run_orthofinder`, but then a different folder and file structure should be followed. Futhermore, I would suggest you to avoid parameter renaming when wrapping external tools, to avoid confusion on usage for users.

src/python/tests/test_orthofinder.py
```diff
@@ -0,0 +1,65 @@
...
+from ensembl.compara.orthtools import prepare_input_orthofinder
+
+@pytest.mark.parametrize(
+    "source_dir, target_dir, expectation",
+    [
+        ("orthtools", "test1", does_not_raise()),
+        ("orthtools", "test1", raises(FileExistsError)),
+        ("orthology_tools", "test2", raises(FileNotFoundError,
+                                            match=r"Directory containing fasta files not found.")),
+        ("orthtools/no_fasta", "test2", raises(FileNotFoundError, match=r"No fasta files found."))
```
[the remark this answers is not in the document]
ipilizota: I'm not familiar with future plans and directions for citest, hence, not sure if it could/would contain fasta files in the future.

src/python/lib/ensembl/compara/orthtools/orthofinder.py
ipilizota: Yes, that could be combined. I wrote it the way I did to emphasise they could be used independently, picking just the one you need. We don't have any examples with `>>>` in the codebase and I find it nicer without it but I can see the point.

src/python/lib/ensembl/compara/orthtools/orthofinder.py
ipilizota: I was using `filecmp.py` as a guidance 👼  How much documentation should be written here in general?
(I see the description for `prepare_input_orthofinder` is wrong here (symlinks) but I think that won't be relevant any more.)

src/python/tests/test_orthofinder.py
```diff
@@ -0,0 +1,65 @@
...
+        source_dir: Path to the directory containing input fasta files.
+        target_dir: Path to the directory where symlinks to fasta files will be created
+                for OrthoFinder to use.
+        tmp_dir: Unit test temp directory (fixture).
+        expectation: Context manager for the expected exception, i.e. the test will only pass if that
+                exception is raised. Use :class:`~contextlib.nullcontext` if no exception is expected.
+
+    """
+    from_dir = os.fspath(pytest.files_dir / source_dir)
+    to_dir = os.fspath(tmp_dir / target_dir)
```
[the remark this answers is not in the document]
ipilizota: Because `prepare_input_orthofinder` uses `str` as input and `pylint` complains if it gets a `Path` object (or maybe it even fails compilation but for some reason I think it was about `pylint`...).

ipilizota:
@JAlvarezJarreta 
Sorry, I rushed to copy `symlink_fasta.py` without properly thinking through it. It seemed like work in progress (missing documentation, no `pytest` yet, `pylint` failing) so I thought a final version from `feature/homology_annotation` would replace the copy. Yup, I omitted thinking about git history...

Given the rest of your feedback, I will abandon work from this PR and close the PR once we finish the discussion.

I will use `symlink_fasta.py` once it's merged into master. In terms of the timeline, it'd be good to have it ready to use when we resolve the dumping PR.

What kind of folder and file structure should be followed for external tools?
I think I may not have an easy and straightforward situation with other orthology tools as it's the case with OrthoFinder.

src/python/tests/test_orthofinder.py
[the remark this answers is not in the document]
JAlvarezJarreta: Fair enough. What about `orthtools` only? The idea is to provide an existing folder where fasta files shouldn't be located. If this changes in the future the test can be easily updated (so you don't need to create new folders and an "empty" file for it).

src/python/lib/ensembl/compara/orthtools/orthofinder.py
JAlvarezJarreta: I'm afraid that is one of the first modules written in Python and we have changed slightly what we aim to include in the module's docstring: brief description, larger description and some usage examples. This is what people tend to do (even the Google docstring example covers it this way) as it is the most informative for users without duplicating much information.

src/python/lib/ensembl/compara/orthtools/orthofinder.py
JAlvarezJarreta: You have a tiny one in `ensembl-compara/src/python/lib/ensembl/compara/utils/tools.py`, but in this case, since you are not producing any output, I will leave it to your preference.

JAlvarezJarreta:
>What kind of folder and file structure should be followed for external tools?
>I think I may not have an easy and straightforward situation with other orthology tools as it's the case with OrthoFinder.

What I meant is that it seems it should be a wrapper script and not a wrapper module (unless there is some utility/need/design I'm missing). Regarding folder structure... If you plan on more elements around it, an `orthofinder` folder could be created. If not, maybe I would call it `orthology` or something like that, so more elements (ours or external) could be added. For scripts we ended up deciding to create a folder in each required module, so the structure would be something like `ensembl/compara/orthology/scripts/run_orthofinder.py`. Does it make sense?

src/python/tests/test_orthofinder.py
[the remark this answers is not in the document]
ipilizota: `orthtools` contains fasta files for other test cases specified in the lines above that one. So providing it as input wouldn't raise `FileNotFoundError: No fasta files found.`

src/python/lib/ensembl/compara/orthtools/orthofinder.py
ipilizota: All clear except the definition of `brief` and `larger`👼  I'll try with more recent examples from the codebase 😉

src/python/lib/ensembl/compara/orthtools/orthofinder.py
ipilizota: 🤔  Although the functions don't `return` anything, outputs are produced and written on the disk (symlinks or OrthoFinder output).

ipilizota:
Yes, it does! Thank you. Let's see how it turns out with other tools and take it from there. 

One of the reasons I called it `orthtools` is so we can have `orthology` for orthology-related stuff that will be in Compara production pipelines. I was planning to put analogous stuff for other chosen tools to `orthotools` as well - there shouldn't be too many methods/functions/scripts so I've been reluctant to create a module per tool.

src/python/tests/test_orthofinder.py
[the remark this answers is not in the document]
dthybert: I guess you should keep the consistency with  prepare_input_orthofinder that uses str, so I would keep fspath. Also I am not sure that in `subprocess.run([script, "-s", target_dir, "-d", source_dir], capture_output=True, check=True)` a Path object would be accepted?

dthybert (approved):

src/python/tests/test_orthofinder.py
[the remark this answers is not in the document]
ipilizota: I think it wouldn't. More precisely, `subprocess.run` would fail with anything other than `str`, `bytes` or `os.PathLike` object.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -0,0 +1,63 @@
...
+    """Docstring"""
+
+
+def prep_input_for_goc():
+    """Docstring"""
+
+
+def calculate_goc_scores():
+    """Docstring"""
+
```
thiagogenez, line 48: I didn't get why these functions are empty. Will they be implemented later?

thiagogenez: Okay - it may will, but just confirming..

src/python/lib/ensembl/compara/orthtools/orthofinder.py
```diff
@@ -0,0 +1,92 @@
...
+                that have been parallelised.(If not specified at all, defaults to
+                min{16, 1/8 * `number_of_threads`}.)
+
+    Raises:
+        FileNotFoundError: If OrthoFinder executable (`orthofinder_exe`) cannot be found.
+        subprocess.CalledProcessError: If executing OrthoFinder command fails for some reason,
+                including `input_dir` not found.
+
+    """
+    orthofinder_exe = "/hps/software/users/ensembl/ensw/C8-MAR21-sandybridge/linuxbrew/bin/orthofinder"
```
thiagogenez, line 89: Would it be a good idea to concentrate harder code path on a shared/common python file?

thiagogenez: It would be easy if things changes...

src/python/tests/flatfiles/config/mlss_conf.xml
```diff
@@ -0,0 +1,21 @@
...
+      <genome name="caenorhabditis_elegans"/>
+      <genome name="saccharomyces_cerevisiae"/>
+    </collection>
+
+    <collection name="test"/>
+
+  </collections>
+
+</compara_db>
+
```
thiagogenez, line 21: Another idea, It would be nice to add some comments about each XML types... just a brief explanation ?

thiagogenez (commented):
Looks all to to me - just left some suggestions 
Nice Work @ipilizota !

src/python/tests/flatfiles/config/mlss_conf.xml
thiagogenez: This would also apply to other XML files.

scripts/pipeline/orthology_benchmark.py
ipilizota: Yes, they will 🙂

src/python/lib/ensembl/compara/orthtools/orthofinder.py
ipilizota: I think I will do something like that when I move this pipeline into NextFlow 👍

src/python/tests/flatfiles/config/mlss_conf.xml
ipilizota: Do you mean explanations for `taxonomic_group`, `ref_for_taxon`?

src/python/lib/ensembl/compara/orthtools/orthofinder.py
CristiGuijarro: Sorry, nosy here: and if it shares NextFlow config for params etc (as one would expect)- this is already existing in the repo :innocent:

PR #367, opened 2021-11-22
Raise step size in ancestral allele dump pipeline

twalsh-ebi:
## Description

During the initial ancestral allele dump pipeline for Ensembl 106, a job failure occurred in the `get_ancestral_sequence` analysis for species `chlorocebus_sabaeus`.

In the `get_ancestral_sequence.pl` script used by this analysis, alignments are processed in step intervals of 10 Megabases, and there is an alignment in this species that spans a step boundary but has a large deletion in the ancestral sequence, such that there is no ancestral sequence corresponding to the interval before the step boundary, resulting in an error due to an undefined `seq` attribute for that part of the alignment.

**Related JIRA tickets:**
- ENSCOMPARASW-4977

## Overview of changes
This update circumvents the error in `chlorocebus_sabaeus` by adding a step-size parameter to `get_ancestral_sequence.pl` and setting this to 20-Megabases by default in the `DumpAncestralAlleles` pipeline, thus avoiding the error condition in the current dataset. The resource class of the `get_ancestral_sequence` analysis has been increased to `2Gb_job` to allow for the increased step size.

## Testing

A test ancestral-allele dump pipeline run was conducted ( pipeline URL `mysql://ensro@mysql-ens-compara-prod-3:4523/twalsh_test_dump_ancestral_alleles_20211122` ).

The output FASTA files were compared against the original release-106 dumps. All were found to be identical except for `chlorocebus_sabaeus_ancestor_scaffold.fa`, whose original version was affected by an issue reported in [ENSCOMPARASW-5011](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-5011). When accounting for that issue, all FASTA files were identical before and after this update.

This update does introduce differences in output BED files at odd-numbered multiples of 10-Megabases that were formerly step boundaries. For example, the following pair of intervals lies either side of the 10-Megabase point in the previously generated `gorilla_gorilla_ancestor_22.bed`:
```
22  9357179 10000000  (((((gorilla,(homo,(pan,pan))),pongo),nomascus),(chlorocebus,(macaca,macaca))),microcebus);
22  10000001  10000525  (((((gorilla,(homo,(pan,pan))),pongo),nomascus),(chlorocebus,(macaca,macaca))),microcebus);
```
These represent a single alignment that happened to overlap the 10-Megabase step boundary. In the copy of `gorilla_gorilla_ancestor_22.bed` generated by the test pipeline, there is no step boundary at 10 Megabases, so a single BED feature is generated:
```
22  9357179 10000525  (((((gorilla,(homo,(pan,pan))),pongo),nomascus),(chlorocebus,(macaca,macaca))),microcebus);
```
Increasing the step size as in this update reduces the number of step boundaries, and in turn causes a reduction in the number of split features such as this.


CristiGuijarro (approved):

twalsh-ebi:
> @twalsh-ebi this probably only needs the one reviewer, but I wanted to check: you have based this code on `release/106` - but this particular code change has not been used in production right? If the code has not been used in production for `e106` - then it should be more appropriately based against `master` as a fix for `e107`. Possibly this has been discussed, but wanted to check.

This PR is essentially equivalent to the code used in production for `chlorocebus_sabaeus` (i.e. using a 20-Megabase step size). This bugfix (or something like it) needs to be in place later in e106 so that the `chlorocebus_sabaeus` ancestral-allele dumps don't fail during the general FTP dumps.

(Though a different step size was used for other primate species, the test pipeline verified that the dumped FASTA files are identical regardless of the step size used.)

PR #369, opened 2021-11-25
Prepare input for OrthoFinder and run OrthoFinder

ipilizota:
## Description

So far, the pipeline dumps core dbs into fasta files. Now, let's run OrthoFinder on them.

**Related JIRA tickets:**
- ENSCOMPARASW-4952
- ENSCOMPARASW-4956

## Overview of changes
1) Prepare input for OrthoFinder (symlinks to the fasta dumps)
2) Run OrthoFinder

#### Change 1
- Function `prep_input_for_orth_tools()` creates a folder with symlinks to the input data for OrthoFinder. 
It will be extended in the future by adding the same functionality for other tools.

#### Change 2
- Function `run_orthology_tools()` just runs OrthoFinder. 
It will be extended in the future by adding the same functionality for other tools.

#### Misc
- Reverted back `.fa` to `.fasta` because `symlink_fasta.py` creates symlinks only for `.fasta` files. Removed `.fa` test files (`.fasta` are still there).
- Added some `print` statements to keep track of the current pipeline status.
- Added a small test collection to the `mlss_conf.xml` to facilitate full pipeline runs during development.

## Testing
- Full pipeline run with default OrthoFinder parameters:
```
ibsub -m 4gb python /hps/software/users/ensembl/repositories/compara/ivana/gene_tree_bench/ensembl-compara/scripts/pipeline/orthology_benchmark.py --mlss_conf /hps/software/users/ensembl/repositories/compara/ivana/gene_tree_bench/ensembl-compara/conf/benchmark/mlss_conf.xml --species_set test_test --host mysql-ens-compara-prod-10 --port 4648 --user ensro --out_dir $HPS_HOME/for_test2 --orthology_input $HPS_HOME/test_orthofinder
```
Folder with symlinks and OrthoFinder output: `/hps/nobackup/flicek/ensembl/compara/ivana/test_orthofinder`
- Full pipeline run with user specified options for OrthoFinder (numbers of threads, reconstruct gene trees from MSA) :
```
bsub -q standard -M16000 -R"select[mem>16000] rusage[mem=16000]" -n 1  -env all -Is python /hps/software/users/ensembl/repositories/compara/ivana/gene_tree_bench/ensembl-compara/scripts/pipeline/orthology_benchmark.py --mlss_conf /hps/software/users/ensembl/repositories/compara/ivana/gene_tree_bench/ensembl-compara/conf/benchmark/mlss_conf.xml --species_set test_test --host mysql-ens-compara-prod-10 --port 4648 --user ensro --out_dir $HPS_HOME/for_test2 --orthology_input $HPS_HOME/test_orthofinder_options --orthofinder_params "-t 8 -M msa"
```
Folder with symlinks and OrthoFinder output: `/hps/nobackup/flicek/ensembl/compara/ivana/test_orthofinder_options/`
- In addition, `pytest test_orthology_benchmark.py` passes.


scripts/pipeline/orthology_benchmark.py
```diff
@@ -157,12 +160,49 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
     return core_names
 
 
-def prep_input_for_orth_tools():
-    """Docstring"""
+def prep_input_for_orth_tools(source_dir: str, target_dir_ofinder: str) -> None:
```
CristiGuijarro, line 163: Can we just name all `target_dir_ofinder` `target_dir`? For a while, except that I realised it was consistent, my eyes kept telling me it was a typo and the intended was `target_dir_orthofinder`. However, since we know this script, aptly named `orthology_benchmark.py` is specific to orthofinder preparation, and for consistency with `source_dir` - `target_dir` makes good sense :+1:

scripts/pipeline/orthology_benchmark.py
```diff
@@ -157,12 +160,49 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+        FileNotFoundError: If directory `source_dir` does not exist or does not contain any fasta files.
+        subprocess.CalledProcessError: If creating symlinks fails for some other reason.
+    """
+    if not os.path.isdir(source_dir):
+        raise FileNotFoundError("Directory containing fasta files not found.")
+    if len(fnmatch.filter(os.listdir(source_dir), '*.fasta')) == 0:
+        raise FileNotFoundError("No fasta files found.")
+
+    # OrthoFinder
+    os.mkdir(target_dir_ofinder)
```
CristiGuijarro, line 185: The `symlink_fasta.py` script does this for you:
```suggestion
```

scripts/pipeline/orthology_benchmark.py
```diff
@@ -157,12 +160,49 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+    script_symlinks = os.path.join(Path(__file__).parents[2], "scripts", "pipeline", "symlink_fasta.py")
+    subprocess.run([script_symlinks, "-s", target_dir_ofinder, "-d", source_dir],
+                   capture_output=True, check=True)
```
CristiGuijarro, line 188: This really doesn't matter, and may be totally subjective, but I like order of data movement direction e.g.:
```suggestion
    subprocess.run([script_symlinks, "-d", source_dir, "-s", target_dir_ofinder],
                   capture_output=True, check=True)
```

scripts/pipeline/orthology_benchmark.py
```diff
@@ -157,12 +160,49 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+
+
+def run_orthology_tools(input_ofinder: str) -> None:
+    """Runs selected orthology inference tools: OrthoFinder.
```
CristiGuijarro, line 192: This is a generic description for a specific function, which only does the one thing. To improve and reduce confusion, could be more concise:
```suggestion
    """Runs OrthoFinder workflow.
```

scripts/pipeline/orthology_benchmark.py
```diff
@@ -157,12 +160,49 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+
+    Args:
+        input_ofinder: Path to the directory containing symlinks to the input fasta files for OrthoFinder.
```
CristiGuijarro, line 195:
```suggestion
        input_ofinder: Path to the directory containing the input fasta files for OrthoFinder.
```
Whilst in your script and methods the input is prepared using symlinks, and this is recommended based on our useage - it does not have to be the case, and running this function should have no impact on the content of the directory for OrthoFinder, so we can be more generic here.

CristiGuijarro, line 195: As with `target_dir_ofinder` I don't think it is necessary to call this `input_ofinder`, `input_dir` or something similar could be clearer.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -157,12 +160,49 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+
+    Raises:
+        FileNotFoundError: If OrthoFinder executable (`orthofinder_exe`) cannot be found.
+        subprocess.CalledProcessError: If executing OrthoFinder command fails for some reason,
+            including `input_ofinder` not found.
+    """
+    # OrthoFinder
+    orthofinder_exe = "/hps/software/users/ensembl/ensw/C8-MAR21-sandybridge/linuxbrew/bin/orthofinder"
+    subprocess.run([orthofinder_exe, "-t", str(32), "-a", str(8), "-f", input_ofinder],
+                   capture_output=True, check=True)
```
CristiGuijarro, line 205:
```suggestion
    s_threads = str(32)
    a_threads = str(8)
    subprocess.run([orthofinder_exe, "-t", s_threads, "-a", a_threads, "-f", input_ofinder],
                   capture_output=True, check=True)
```
I suggest this a little for transparency and clarity on what the parameters are pertaining (for those that know nothing about orthofinder), although I think what you have is also fine

scripts/pipeline/orthology_benchmark.py
ipilizota: I'd be up for `target_dir_orthofinder` but I thought it might be too long. The script `orthology_benchmark.py` is not specific to OrthoFinder and as I said in the description, `prep_input_for_orth_tools()` and `run_orthology_tools()` will be extended to handle other tools. What I didn't say is that will happen already in Dec/Jan 😄

scripts/pipeline/orthology_benchmark.py
ipilizota: `symlink_pasta.py` does `Path(symlink_dir).mkdir(parents=True, exist_ok=True)`. So, let's say that there was already a dir `dummy` with some other unrelated analyses and a user mistakenly sets it for `target_dir_ofinder`. 
- `os.mkdir(target_dir_ofinder)` would throw an error because the directory already exists. The script would stop and no symlinks would be created.
- If that line would be removed, `symlink_fasta.py` would create symlinks in `dummy` and the pipeline would continue running, potentially messing up its already existing content.

scripts/pipeline/orthology_benchmark.py
ipilizota: Me too, actually 😄

scripts/pipeline/orthology_benchmark.py
ipilizota: See one of the comments above 😉

scripts/pipeline/orthology_benchmark.py
ipilizota: It will be very soon 👼  `input_orthofinder`?

scripts/pipeline/orthology_benchmark.py
ipilizota: In fact, I had `number_of_threads` and `number_of_orthofinder_threads` in that abandoned [PR](https://github.com/Ensembl/ensembl-compara/pull/365) but JA was very critical of that:
> Futhermore, I would suggest you to avoid parameter renaming when wrapping external tools, to avoid confusion on usage for users.

Truth be told, the authors use `number_of_threads` and `number_of_orthofinder_threads` in [the manual](https://github.com/davidemms/OrthoFinder/blob/master/OrthoFinder-manual.pdf). 

Since here I don't provide an option to actually pass the number of threads via command line, and `s_threads` is not much less cryptic than `s` (same for `a`) coupled with adding unnecessary variable declarations, I will actually add options to pass these numbers  via command line (with let's say `-t 32 -a 8` as default).

Related to what JA said, what do you think about variable names?

ipilizota:
Thank you very much @CristiGuijarro for a quick review! I will commit all code changes tomorrow 😉 
You've probably received an email per comment but github.com (or my Chrome) is messed up again and it doesn't let me to "Start a review", only "Add single comment" 🙈

scripts/pipeline/orthology_benchmark.py
CristiGuijarro: The author's also use `search_threads` and `analysis_threads` :point_left: so these may be even more appropriate. This is not renaming the parameters, because the `-t` and `-a` flags remain in the code and have not been renamed, it is the values that are being named to accompany the flags in my suggestion. As parameters that are optimised for a specific environment setup, I would encourage providing them as default values, but also give flexibility to changing them as you mention in the commandline. For this dataset of just 2 genomes, these parameters could even be overkill, they are selected for a dataset of ~40 genomes. But as I said, I think it is fine without, it was just a suggestion :shrug:

scripts/pipeline/orthology_benchmark.py
CristiGuijarro: I am not sure that is a real problem. It's a technical possibility, in which there are hundreds we cannot possibly think of, but rather unlikely to happen. This is not far off what occurs in almost every single production pipeline we run, with `rm_mkdir` - in which the directories are silently overridden. In this case, there is the potential issue that orthofinder or other-orthotool is run on files we would not want it to run on, but these tools do not alter fasta files and worst case scenario is producing data we do not want. If this particular script was moved into a workflow manager such as NextFlow or SnakeMake the equivalent type of method would be used for creating the new directory too.

scripts/pipeline/orthology_benchmark.py
CristiGuijarro: This is really confusing, but I don't know the future! So the function will be used for other tools, but it has an orthofinder specific parameter passed in? Are you planning to then have a different parameter in the function for each of the other orthology-tools? And what of those that all take a simple `target_dir` of fasta files? To me it makes more sense to write this function more generically, and then it can be added to/altered to fit the next stages. It's nice to have a script that looks like it could be finished, without lots of silent placeholders no-one else knows about. If you were to win the lottery and decided to buy a tropical island without internet to live on for a year, someone else should be able to pick up the script, and add the functionality. What input do the other orthology-tools you have in mind take?

scripts/pipeline/orthology_benchmark.py
CristiGuijarro: A side note on this, if you see potential pitfalls in a compara script you are using in your code, you are encouraged to tweak it to fit your purposes, it makes it more generic, and it could be more useful to other purposes too - for example, passing in a flag to `symlink_fasta.py` to force it to die if the directory already exists.

scripts/pipeline/orthology_benchmark.py
CristiGuijarro: Your description still specifies `OrthoFinder` though - so in that case:
```suggestion
    """Runs the specified orthology inference tool.
```

scripts/pipeline/orthology_benchmark.py
CristiGuijarro: I'm :+1: on the name improvement - because who knows, there could be a future tool called `ofinder`! But I await your comments on how these other tool specific parameters will be passed in

scripts/pipeline/orthology_benchmark.py
ipilizota: I should've linked tickets for preparing input for other tools to the ticket ENSCOMPARASW-4952 so it'd make more sense for people not working on the project.

Yes, there will be input parameters for other tools. E.g. OMA will need `parameters.drw` file. I remember it requiring placing input fasta files in a particular location with respect to runnables and `parameters.drw`.

scripts/pipeline/orthology_benchmark.py
ipilizota: What's the problem with ensuring that OrthoFinder is run on the dataset the user wants?

CristiGuijarro: That isn't a problem, the user passes in the directory via the command line - so it should be the dataset the user wants to run it on. My suggestions are based on redundancy and my comments in response to your explanation are to further support my suggestion, but also to include that if you see it as an issue, perhaps it would be nice for you to alter the existing code in `symlink_fasta.py` to fit your purpose.

scripts/pipeline/orthology_benchmark.py
ipilizota: I didn't rename `-t` and `a`. I had `"-t", str(number_of_threads), "-a", str(number_of_orthofinder_threads)` 😉 

I like `search_threads` and `analysis_threads` 👍 The pipeline is being developed for any number of genomes, and as I said in the description, this is a small test collection to facilitate full pipeline run. Actually, I could've kept those changes in the `mlss_conf.xml` locally 🤔

CristiGuijarro: Yes, exactly, I think you are on the same page as me.

scripts/pipeline/orthology_benchmark.py
CristiGuijarro: For this, I would keep generic parameters shared by the tools, such as a `source_dir` (as you have) and a `target_dir` :point_left: because this is a common shared parameter that the majority of the tools, if not all of them, will use however you look at it, even if they require further processing in the prep stage. And for additional parameters that some tools do or do not have, perhaps you could incorporate those as and when needed in the future - although I suspect that might be more appropriate for the `run_orthology_tools`

scripts/pipeline/orthology_benchmark.py
```diff
@@ -157,12 +160,49 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+        raise FileNotFoundError("No fasta files found.")
+
+    # OrthoFinder
+    os.mkdir(target_dir_ofinder)
+    script_symlinks = os.path.join(Path(__file__).parents[2], "scripts", "pipeline", "symlink_fasta.py")
+    subprocess.run([script_symlinks, "-d", source_dir, "-s", target_dir_ofinder],
+                   capture_output=True, check=True)
+
+
+def run_orthology_tools(input_ofinder: str) -> None:
```
CristiGuijarro, line 191: As with comment on shared parameters, this too would likely be shared among the orthology tools, as `input_dir` or the like.

scripts/pipeline/orthology_benchmark.py
CristiGuijarro:
```suggestion
        input_dir: Path to the directory containing the input .fasta files for orthology tools.
```

scripts/pipeline/orthology_benchmark.py
CristiGuijarro, line 163:
```suggestion
def prep_input_for_orth_tools(source_dir: str, target_dir: str) -> None:
```

scripts/pipeline/orthology_benchmark.py
```diff
@@ -157,12 +160,49 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
     return core_names
 
 
-def prep_input_for_orth_tools():
-    """Docstring"""
+def prep_input_for_orth_tools(source_dir: str, target_dir_ofinder: str) -> None:
+    """Prepares input files for selected orthology inference tools.
 
+    For OrthoFinder: Creates symlinks to input fasta files in `target_dir_ofinder`.
+    # For OMA: ...
```
CristiGuijarro, line 167:
```suggestion
    """Prepares input files for selected orthology inference tools.
    Collects fasta files into single directory: `target_dir`
```

scripts/pipeline/orthology_benchmark.py
```diff
@@ -157,12 +160,49 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
 
-def run_orthology_tools():
-    """Docstring"""
+    Args:
+        source_dir: Path to the directory containing fasta files.
+        target_dir_ofinder: Path to the directory where symlinks to fasta files will be created
+            for OrthoFinder to use.
```
CristiGuijarro, line 172:
```suggestion
        target_dir: Path to the directory where symlinks to fasta files will be created.
```

scripts/pipeline/orthology_benchmark.py
```diff
@@ -157,12 +160,49 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+
+    Raises:
+        FileExistsError: If directory `target_dir_ofinder` already exists.
```
CristiGuijarro, line 175:
```suggestion
        FileExistsError: If directory `target_dir` already exists.
```

CristiGuijarro, line 175:
```suggestion
```

CristiGuijarro: (If adamant about not moving the `mkdir` line)

scripts/pipeline/orthology_benchmark.py
```diff
@@ -157,12 +160,49 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+    """
+    if not os.path.isdir(source_dir):
+        raise FileNotFoundError("Directory containing fasta files not found.")
+    if len(fnmatch.filter(os.listdir(source_dir), '*.fasta')) == 0:
+        raise FileNotFoundError("No fasta files found.")
+
+    # OrthoFinder
+    os.mkdir(target_dir_ofinder)
+    script_symlinks = os.path.join(Path(__file__).parents[2], "scripts", "pipeline", "symlink_fasta.py")
+    subprocess.run([script_symlinks, "-d", source_dir, "-s", target_dir_ofinder],
```
CristiGuijarro, line 187:
```suggestion
    subprocess.run([script_symlinks, "-d", source_dir, "-s", target_dir],
```

scripts/pipeline/orthology_benchmark.py
CristiGuijarro, line 191:
```suggestion
def run_orthology_tools(input_dir: str) -> None:
```

scripts/pipeline/orthology_benchmark.py
```diff
@@ -157,12 +160,49 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+        input_ofinder: Path to the directory containing symlinks to the input fasta files for OrthoFinder.
+
+    Raises:
+        FileNotFoundError: If OrthoFinder executable (`orthofinder_exe`) cannot be found.
+        subprocess.CalledProcessError: If executing OrthoFinder command fails for some reason,
+            including `input_ofinder` not found.
+    """
+    # OrthoFinder
+    orthofinder_exe = "/hps/software/users/ensembl/ensw/C8-MAR21-sandybridge/linuxbrew/bin/orthofinder"
+    subprocess.run([orthofinder_exe, "-t", str(32), "-a", str(8), "-f", input_ofinder],
```
CristiGuijarro, line 204:
```suggestion
    subprocess.run([orthofinder_exe, "-t", str(32), "-a", str(8), "-f", input_dir],
```

CristiGuijarro (changes requested):

scripts/pipeline/orthology_benchmark.py
ipilizota: Now I see you added a bunch of suggestions in the meantime (I was busy with other stuff and couldn't work on this) so I'm not sure if this is still relevant or relevant at all. At the moment, everything is read from the command line. Since the list of pipeline parameters is... becoming longer and longer, I'm thinking of passing the parameters via input file, e.g. json.

(I've been incorporating changes locally since the morning but other things come up... I will re-request your review once

scripts/pipeline/orthology_benchmark.py
ipilizota: Errr... searching the manual didn't give any results for `search_threads` and `analysis_threads`. I didn't manage to spot them by scrolling either.
I guess they have switched to `number_of_threads` and `number_of_orthofinder_threads`. `number_of_threads` also sounds generic enough if shared across different tools. So I went with them.

scripts/pipeline/orthology_benchmark.py
ipilizota: Yes, you're right - the user passes in the directory, thus it's prone to human error. Better safe than sorry.

BTW, it'd be great to extend the functionality of `symlink_fasta.py` (could also add the option for `.fa` files). At the moment it still looks like work in progress (no usage example, no pytest) so once that's done, I can extend it with the bits that'd be useful for `orthology_benchmark.py`. pytest would be helpful for future developments to ensure that the new code doesn't mess up the existing functionality which is desired to remain or be improved.

As of `os.mkdir(target_dir)`, I also have similar safety measure elsewhere in this script, e.g. https://github.com/Ensembl/ensembl-compara/pull/364/files#r750455647

Anyways, I'll go through the whole pipeline when moving it to NextFlow, some Python stuff will be removed so that will be a good time for a quick iteration of further polishing.

scripts/pipeline/orthology_benchmark.py
CristiGuijarro:
```
$ orthofinder --help

...

OPTIONS:
 -t <int>        Number of parallel sequence search threads [Default = 16]
 -a <int>        Number of parallel analysis threads
```

ipilizota: Oh, I see what you meant. I was referring to `search_threads`, `analysis_threads`, `number_of_threads`, `number_of_orthofinder_threads`.

CristiGuijarro (approved):

dthybert (changes requested):

scripts/pipeline/orthology_benchmark.py
```diff
@@ -157,12 +167,52 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+    """
+    if not os.path.isdir(source_dir):
+        raise FileNotFoundError("Directory containing .fasta files not found.")
+    if len(fnmatch.filter(os.listdir(source_dir), '*.fasta')) == 0:
+        raise FileNotFoundError("No .fasta files found.")
+
+    # OrthoFinder
+    os.mkdir(target_dir)
+    script_symlinks = os.path.join(Path(__file__).parents[2], "scripts", "pipeline", "symlink_fasta.py")
+    subprocess.run([script_symlinks, "-d", source_dir, "-s", target_dir],
```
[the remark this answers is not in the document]
ipilizota: It throws `subprocess.CalledProcessError`. 

Side note: Moving to NextFlow, I think this piece of Python code will be removed and replaced with a shell command.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -157,12 +167,52 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+        FileNotFoundError: If directory `source_dir` does not exist or does not contain any .fasta files.
+        subprocess.CalledProcessError: If creating symlinks fails for some other reason.
+    """
+    if not os.path.isdir(source_dir):
+        raise FileNotFoundError("Directory containing .fasta files not found.")
+    if len(fnmatch.filter(os.listdir(source_dir), '*.fasta')) == 0:
+        raise FileNotFoundError("No .fasta files found.")
+
+    # OrthoFinder
+    os.mkdir(target_dir)
```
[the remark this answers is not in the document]
ipilizota: If the directory exists, it will throw `FileExistsError`. I find that more favourable than adding new symlinks/fasta files into an existing directory. In case the existing directory already contains fasta files or symlinks to fasta files, OrthoFinder would use all of them as input.

ipilizota:
Following the meeting with DT, I:
* Rewrote functions to handle all potential errors with try-except blocks, providing verbose msg in case of an error
* Simplified and extended possibilities for additional OrthoFinder input parameters: all params & corresponding values (if any) are now passed in via command line as a single string

scripts/pipeline/orthology_benchmark.py
```diff
@@ -157,12 +165,57 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+    except subprocess.CalledProcessError as exc:
+        msg = f"Command '{exc.cmd}' returned non-zero exit status {exc.returncode}"
+        if exc.stdout:
+            msg += f"\n  StdOut: {exc.stdout}"
+        if exc.stderr:
+            msg += f"\n  StdErr: {exc.stderr}"
+        raise RuntimeError(msg) from exc
+
+
+def run_orthology_tools(input_dir: str, orthofinder_params: str) -> None:
```
CristiGuijarro, line 195: My only comment here is, since you have made changes suggested by David, I want to see which direction this is headed now: with all the orthofinder tool params wrapped in  a single param string how will it look with different orthology tools? Will each orthology tool require it's own `orthotool_params` ? Perhaps they could be provided in that more tool-flexible way since they are now a string? You would include in the end of the string the flag/param to include the input once processed, and otherwise, pass in tool and params in the same blackbox style we have been looking at - so no tool has to be hard-coded in the code as such - with another param/flag to determine the type of preprocessing required in the file prep. This might be a bit of a pain if you need this code quite soon, so I don't think I expect the changes this increment, but something to think about for the next.

CristiGuijarro (approved):

scripts/pipeline/orthology_benchmark.py
CristiGuijarro: Actually, one more comment, perhaps pulling in the most recent changes from `main` will help Travis get over his grievances. My approval is conditional on Travis passing, as one would expect :crossed_fingers:

dthybert: I do not understand your comment, could you please detail more because it seems interesting?

scripts/pipeline/orthology_benchmark.py
```diff
@@ -157,12 +165,57 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+        RuntimeError: If OrthoFinder command fails to execute for any reason.
+    """
+    # OrthoFinder
+    orthofinder_exe = "/hps/software/users/ensembl/ensw/C8-MAR21-sandybridge/linuxbrew/bin/orthofinder"
+    cmd = f"{orthofinder_exe} -f {input_dir} {orthofinder_params}"
+
+    try:
+        subprocess.run(cmd, capture_output=True, check=True, shell=True, text=True)
+    except subprocess.CalledProcessError as exc:
+        msg = f"Command '{exc.cmd}' returned non-zero exi status {exc.returncode}"
```
dthybert, line 213: returned non-zero exi  --> returned non-zero exit

dthybert (changes requested):
Once the typo is corrected, this can be considered as approved on my side

scripts/pipeline/orthology_benchmark.py
ipilizota: Thank you @CristiGuijarro for a speedy reaction and your thoughts on the future developments. The current plan is to move on with pipeline development by adding functionality to prepare input files for GOC score calculation (GTF files, files with OrthoFinder putative orthologs) and calculate GOC scores. Once that's done, i.e. I have a fully working pipeline, I'd like to move away from providing pipeline parameters via command line. All pipeline params would be provided in a (`json`) file. There I could also have a full path to `orthofinder_exe`. In case I add another tool to the pipeline before that, it would be dealt in the same way as OrthoFinder now: the pipeline would run the tool in its default mode if no additional params are provided, and if there are additional params, they'd be provided by corresponding `orthotool_params` (a user would specify them via string in a command line).

How does that sound to you?

P.S. The pipeline is conceived to be running all tools specified in `run_orthology_tools()`. It'd be nice to be able to run the pipeline only for a specified subset of orthology tools. I think that could be another small ticket. But let's make one step at a time 🙂

ipilizota:
Travis is failing for reasons not related to this code. Merging.

PR #380, opened 2021-12-13
Refactoring dump_genomes()

ipilizota:
## Description

In its current state, `dump_genomes()` takes as input a list of species names to dump and calls `get_core_names()` to determine names of the cores to dump. Core names will also be needed in a function to retrieve GTF files. To avoid calling `get_core_names()` twice, I refactored `dump_genomes()` so that `get_core_names()` is called before `dump_genomes()`, and the output from `get_core_names()` is used as input for `dump_genomes()`.

**Related JIRA tickets:**
- ENSCOMPARASW-5013

#### Change 1
- `get_core_names()` is not called from `dump_genomes()` anymore. It is called before `dump_genomes()` and its output is used as input for `dump_genomes()`.

#### Change 2
- [This wasn't necessary] In `dump_genomes()`, I put `os.mkdir()` and `subprocess.run()` within `try-except` blocks. This was inspired by the most recent PR review process 🙂  I also extended the dumping command with `id_type` option so the fasta dumps can have gene stable ids in the header.

#### Change 3
- Updated and extended `test_orthology_benchmark.py` accordingly.

## Testing
All `pytests` pass. In addition, I made a full pipeline run:
```
ibsub -m 4gb python /hps/software/users/ensembl/repositories/compara/ivana/gene_tree_bench/ensembl-compara/scripts/pipeline/orthology_benchmark.py --mlss_conf /hps/software/users/ensembl/repositories/compara/ivana/gene_tree_bench/ensembl-compara/conf/benchmark/mlss_conf.xml --species_set test_test --host mysql-ens-compara-prod-10 --port 4648 --user ensro --out_dir $HPS_HOME/test_refactoring --id_type gene --orthology_input $HPS_HOME/test_orthofinder
```
- Fasta dumps: `/hps/nobackup/flicek/ensembl/compara/ivana/test_refactoring/test_test`
- Orthofinder input and output: `/hps/nobackup/flicek/ensembl/compara/ivana/test_orthofinder`


CristiGuijarro (changes requested):
Could you put `dump_genomes` back where it was so that the changes are more apparent? This will make it quicker for reviewers to review, and therefore quicker to merge. This move makes it look like a whole new function to replace the old, and I don't think that is the case.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -42,46 +42,54 @@
...
-    for core in dump_cores:
+    for core in core_list:
 
         out_file = os.path.join(dumps_dir, f"{core}.fasta")
 
-        subprocess.run([script, "-core-db", core, "-host", host, "-port", str(port),
-                       "-outfile", out_file], capture_output=True, check=True)
+        try:
+            subprocess.run([script, "-core-db", core, "-host", host, "-port", str(port),
+                            "-outfile", out_file, "-id_type", id_type], capture_output=True, check=True)
```
CristiGuijarro, line 85: Just a small change that I think is required is to either provide a default `id_type` if not provided or change the `subprocess.run` call to not pass `id_type` if not given

CristiGuijarro (changes requested):

CristiGuijarro (approved):

dthybert (approved):
sounds good

PR #383, opened 2021-12-14
Include protists specific pipeconfigs

CristiGuijarro:
## Description

Very basic pipeconfigs for the Protists division - necessary for ease of running production and reducing human error in command line parameters, subject to alteration during production.

**Related JIRA tickets:**
- ENSCOMPARASW-5026 :
- ENSCOMPARASW-5028

## Overview of changes
Quite generally the process has mimicked that of the Metazoa division, to include all the pipeconfigs that will be necessary for e107. Whilst for `e107` we will only run the protein trees pipeline, the WGA related configs are included here for ease in the future too. Additional parameters come from legacy EG division specifications for Protists as they would have been run by the Microbes team.

## Testing
NA

## Notes
- Certain resources are expected to be updated in the live running of production, at the moment they are not known.
- Have included dnds analysis because it is included still for `Metazoa` division and was specified in the old EG division related config.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/Protists/ProteinTrees_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/Protists/ProteinTrees_conf.pm
@@ -64,10 +64,10 @@ sub default_options {
         ],
 
         # Extra analyses:
-        # Compute dNdS for homologies?
-        'do_dnds'                => 1,
         # Gain/loss analysis?
         'do_cafe'                => 0,
+        # Compute dNdS for homologies?
+        'do_dnds'                => 1,
         # Do we want the Gene QC part to run?
         'do_gene_qc'             => 0,
         # Do we need a mapping between homology_ids of this database to another database?
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/Protists/ProteinTrees_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/Protists/ProteinTrees_conf.pm
@@ -1,0 +1,80 @@
+=head1 LICENSE
+
+See the NOTICE file distributed with this work for additional information
+regarding copyright ownership.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


modules/Bio/EnsEMBL/Compara/PipeConfig/Protists/ProteinTrees_conf.pm
```diff
@@ -0,0 +1,80 @@
...
+use base ('Bio::EnsEMBL::Compara::PipeConfig::ProteinTrees_conf');
+
+
+sub default_options {
+    my ($self) = @_;
+
+    return {
+        %{$self->SUPER::default_options},   # inherit the generic ones
+
+        'division'   => 'protists',
```
ipilizota, line 49: I'm comparing files from this PR to those in `.../PipeConfig/Metazoa`. In Metazoa, after this line there is `'collection' => $self->o('division'),`. I'm aware it's not like that in all divisions. Why going without `'collection'` in Protists?

modules/Bio/EnsEMBL/Compara/PipeConfig/Protists/ProteinTrees_conf.pm
```diff
@@ -0,0 +1,80 @@
...
+                'taxa'          => [ 'all' ],
+                'thresholds'    => [ undef, undef, 25 ],
+            },
+        ],
+
+        # Extra analyses:
+        # Compute dNdS for homologies?
+        'do_dnds'                => 1,
+        # Gain/loss analysis?
+        'do_cafe'                => 0,
```
ipilizota, line 70:
```suggestion
        # Gain/loss analysis?
        'do_cafe'                => 0,
        # Compute dNdS for homologies?
        'do_dnds'                => 1,
```
Ordering params alphabetically 😉

ipilizota (changes requested):
Given that these configs work for Metazoa, I believe they should be good enough for Protists as well. I guess hive-related params in `ProteinTrees_conf.pm` will be added along the way, if needed. Overall, good to go 👍

modules/Bio/EnsEMBL/Compara/PipeConfig/Protists/ProteinTrees_conf.pm
CristiGuijarro: This was a change in `e106` across divisions to make sure that the `collection/clusterset.name` will be `default` instead of `Protists`. You might remember the discussions we had as a team about it: [ENSCOMPARASW-4706](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-4706)

ipilizota (approved):

twalsh-ebi (approved):
Looks good. 👍

PR #386, opened 2022-01-05
A function to find and copy a GTF file for a particular core db

ipilizota:
## Description

The main goal of the whole pipeline is the ability to benchmark orthology methods based on GOC scores for their predictions. To calculate a GOC score, we need to know neighbouring genes of the genes involved in a prediction. Hence, we need GTF files for the genomes involved.

This function finds and copies a GTF file for a specified core db.

**Related JIRA tickets:**
- ENSCOMPARASW-4982

## Overview of changes
New function and corresponding `pytest` with a test file.

## Testing
`pytest`. (All tests pass.)

## Notes
- We agreed at the Dev ** Table that I can utilise the structure of `production/ensemblftp` or Marc's personal dir to narrow down the search for a file. Making it more generic would mean traversing a lot of directories (takes way too long to find a single file let alone multiple files).
- Since this is a public repo, I was not sure how explicit I could be about `production/ensemblftp` or Marc's personal directory in the documentation.


dthybert (changes requested):
Hi Ivana, I have added my feedbacks, most are to improve readability of the code or to have it less complicated.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -175,6 +177,63 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+    """
+    species_name = core_name.split("_core_")[0]
+    rel_ver = core_name.split("_core_")[1].split("_")
+
+    # gtf_dirs contains all subdirs of source_dir which could contain the gtf file
+    # Vertebrates in `production/ensemblftp`:
+    gtf_dirs = [os.path.join(source_dir, "release-" + rel_ver[0], "gtf", species_name)]
+    # Vertebrates in MC's personal dir and
+    # all other divisions whether in `production/ensemblftp` or MC's dir
+    divisions = ["vertebrates", "plants", "metazoa", "bacteria", "fungi", "protists"]
```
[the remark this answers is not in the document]
dthybert: maybe you can use a list that filter out rather than filter in?

scripts/pipeline/orthology_benchmark.py
```diff
@@ -175,6 +177,63 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+    for division in divisions:
+        gtf_dirs.append(os.path.join(source_dir, "release-" + rel_ver[0], division, "gtf", species_name))
+
+    for directory in gtf_dirs:
+        try:
+            gtf_file = [file for file in glob.glob(os.path.join(directory, "*"))
```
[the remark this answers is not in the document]
dthybert: ha yes the assembly name is quite random, I see then you do not have the choice.

dthybert (approved):

scripts/pipeline/orthology_benchmark.py
```diff
@@ -175,6 +177,64 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+    release = core_name.split("_core_")[1].split("_")[0]
+
+    parent_dir = os.path.join(source_dir, "release-" + release)
+    # gtf_dirs contains all subdirs which could contain the gtf file
+    # Vertebrates in `production/ensemblftp`:
+    gtf_dirs = [os.path.join(parent_dir, "gtf", species_name)]
+    # Vertebrates in MC's personal dir and
+    # all other divisions whether in `production/ensemblftp` or MC's dir
+    potential_divisions = [f.name for f in os.scandir(parent_dir) if f.is_dir()]
+    for division in potential_divisions:
```
twalsh-ebi, line 210: Since you're only iterating over `potential_divisions` once, why not access the `scandir` iterator directly in the for loop?

twalsh-ebi (changes requested):
It's nice to see this function come to life. I just have a few questions/suggestions.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -175,6 +177,64 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+
+    for directory in gtf_dirs:
+        try:
+            gtf_file = [file for file in glob.glob(os.path.join(directory, "*"))
+                        if file.endswith(f".{release}.gtf.gz")][0]
+        except IndexError:
+            continue
+
+        try:
+            cmd = f"cp {gtf_file} {target_dir}"
```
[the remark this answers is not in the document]
ipilizota: If `target_dir` doesn't exist, `subprocess.run` fails and `RuntimeError` is raised. Would you strongly advise going with `shutil.copy` or `shutil.copyfile` instead?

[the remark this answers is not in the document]
twalsh-ebi: I would indeed advise to use an `shutil` function such as `shutil.copy` or `shutil.copyfile`. In this case `shutil.copy` would be more similar to the existing command since it allows the target to be a directory.

One caveat: if the target directory doesn't exist but its parent does, the copied file will be saved to a file with the same name as the intended target directory. To avoid this, you could ensure the target directory exists beforehand with `os.makedirs` as follows:
```
os.makedirs(target_dir, exist_ok=True)
shutil.copy(gtf_file, target_dir)
```
This doesn't handle the check of the `result` in the same way that the existing code does, but I'd advise using an `shutil` copy function anyway because it will make the code more concise.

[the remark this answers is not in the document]
ipilizota: It probably doesn't handle errors in the same way as `subprocess` does, either. So this requires a bit more rewriting. Just wanted to check if it was worth doing it. Thank you for your extensive reply, @twalsh-ebi.

scripts/pipeline/orthology_benchmark.py
ipilizota: Do you mean something like:
1) `for division in [f.name for f in os.scandir(parent_dir) if f.is_dir()]:` -> I think the current solution is more readable
2) `gtf_dirs += [os.path.join(parent_dir, f.name, "gtf", species_name) for f in os.scandir(parent_dir) if f.is_dir()]` -> Not sure I can read this without getting dizzy
3) Something else I can't think of at the moment?

twalsh-ebi: I was thinking more along the lines of:
```suggestion
    for f in os.scandir(parent_dir):
        if f.is_dir():
            gtf_dirs.append(os.path.join(parent_dir, f.name, "gtf", species_name))
```
In fairness, this is not very concise, but I'd venture that it's not entirely unreadable.

ipilizota: Thank you very much @twalsh-ebi. I've incorporated this locally so I will mark it as resolved here. I might swap it with the suggestion from Slack, tho 😄  But let me make the first iteration incorporating all the feedback from here 😉

scripts/pipeline/orthology_benchmark.py
```diff
@@ -175,6 +177,64 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+        core_name: Core db name.
+        source_dir: Path to the directory containing subdirectories with GTF files.
+            `production/ensemblftp` or MC's `nobackup/release_dumps`.
+        target_dir: Path to the directory where the GTF file will be copied.
+
+    Raises:
+        RuntimeError: If command to copy a file fails for any reason.
+
+    Warns:
+        UserWarning: If the GTF file was not found.
```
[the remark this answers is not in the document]
ipilizota: Good point. I made local changes so in case the function looks for the GTF file of e.g. a `release-53` core, and `release-53/` doesn't exist, it raises a warning that the GTF file was not found. Will push later.

dthybert (approved):

scripts/pipeline/orthology_benchmark.py
```diff
@@ -175,6 +178,49 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+        except IndexError:
+            continue
+
+        os.makedirs(target_dir, exist_ok=True)
+        shutil.copy(gtf_file, target_dir)
+        copied = True
+
+    try:
+        copied
+    except NameError:
```
twalsh-ebi, line 220: I would avoid testing for a successful copy using a `NameError`. If a `copied` variable is defined in the global scope, there will be no `NameError` and this will fail to warn if a GTF file was not found.

Two alternative ways to test if a GTF file was found:
1. Initialise `gtf_file` to `None` before the for loop, then issue a warning `if gtf_file is None` after the for loop.
2. Create a list of `gtf_files`, gathering all matching GTF files into that list (using a for loop, or maybe a list comprehension?), and then access the first element of that list. If you catch an `IndexError` at that point, you can be sure of the need for a 'GTF not found' warning.

twalsh-ebi (changes requested):
This is really shaping up. But I hope you don't mind, I've added one more round of comments/suggestions.

As a general point, I think the code to search for a GTF file could be more clearly separated from the code to copy it. In particular, I think that the calls to `os.makedirs` and `shutil.copy` should be moved out of the for loop iterating over `gtf_dirs_patterns`, and the check of whether a GTF file was found should be put immediately after the for loop. As things are, it is technically possible that two GTF files are matched and copied.

scripts/pipeline/orthology_benchmark.py
ipilizota: So true and so obvious... Thank you for spotting it 👍

ipilizota:
Addressed 👍

scripts/pipeline/orthology_benchmark.py
```diff
@@ -175,6 +178,47 @@ def get_core_names(species_names: List[str], host: str, port: int, user: str) ->
...
+        os.path.join(parent_dir, "gtf", species_name)  # vertebrates in `production/ensemblftp`
+    ]
+
+    gtf_file = None
+    gtf_file_pattern = f"{species_name.capitalize()}.*.{release}.gtf.gz"
+    for pattern in gtf_dirs_patterns:
+        try:
+            gtf_file = list(glob.glob(os.path.join(pattern, gtf_file_pattern)))[0]
+        except IndexError:
+            continue
```
twalsh-ebi, line 213: An optional suggestion to ensure the search stops as soon as there is a match:
```suggestion
        except IndexError:
            continue
        else:
            break
```

twalsh-ebi (approved):

PR #389, opened 2022-01-10
Configure sugarcane vs wheat LASTZ alignment for e107

ipilizota:
## Description

Configuring sugarcane vs wheat pairwise alignment to be computed in e107

**Related JIRA tickets:**
- ENSCOMPARASW-5064

## Overview of changes
Added appropriate line to the Plants `mlss_conf.xml`

## Testing
NA

### DIFF
--- a/conf/plants/mlss_conf.xml
+++ b/conf/plants/mlss_conf.xml
@@ -59,6 +59,7 @@
       </species_set>
     </one_vs_all>
 
+    <pairwise_alignment method="LASTZ_NET" ref_genome="triticum_aestivum" target_genome="saccharum_spontaneum"/>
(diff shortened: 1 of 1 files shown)


CristiGuijarro (commented):
Just a quick comment/check based on frozen-ness. If sugar cane is to be aligned to the wheat reference, do we still want it aligned to the rice reference too? If not, it should be excluded  explicitly after line 51. If it's fine to align it to both, triggering 2 lastz alignments and not just 1, then I see no reason not to approve :+1:

ipilizota:
@CristiGuijarro, I will get back to you when I find out the answer to your question.

(Just to have it written: I think Travis is failing because `saccharum_spontaneum` is not yet in `allowed_species.json`.)

ipilizota:
@CristiGuijarro, we will compute sugarcane vs rice as well.

CristiGuijarro (approved):

PR #392, opened 2022-01-20
Update GERP pipeline with the latest scripts

JAlvarezJarreta:
## Description

Updated GERP pipeline with the latest scripts available.

**Related JIRA tickets:**
- ENSCOMPARASW-5060

## Overview of changes
- Added `hal_to_maf.py` and `run_gerp.py` scripts
- Added required new parameters
- Renamed latest analysis to meet the FTP output (for now)

## Testing
No testing was done. Will be tested when at least the MAF to MFA analysis is ready.

### DIFF
--- a/pipelines/GERPonHAL/gerp_on_hal.nf
+++ b/pipelines/GERPonHAL/gerp_on_hal.nf
@@ -1,6 +1,6 @@
 #!/usr/bin/env nextflow
 
-genomes_file = file(params.genomes_file)
+keep_genomes_file = file(params.keep_genomes_file)
 tree_file = file(params.tree_file)
 
 // Initialise workflow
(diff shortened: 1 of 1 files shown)


pipelines/GERPonHAL/gerp_on_hal.nf
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env nextflow
 
+genomes_file = file(params.genomes_file)
```
twalsh-ebi, line 3: Would it be possible to name this parameter `keep_genomes_file` to match the parameter in the `hal_to_maf.py` script?

pipelines/GERPonHAL/gerp_on_hal.nf
```diff
@@ -1,5 +1,8 @@
...
+tree_file = file(params.tree_file)
```
twalsh-ebi, line 4: Would the Newick tree in the HAL file be sufficient for GERP? If yes, I'd suggest adding a process to do this. The command to extract the tree goes something like this:
```
halStats --tree input.hal > output.nh
```

twalsh-ebi (commented):
The pipeline itself looks fine. I've just made a couple of comments about the new parameters.

pipelines/GERPonHAL/gerp_on_hal.nf
JAlvarezJarreta: CACTUS needs a tree file as input, and I would expect that one and the one generated by `halStats` to be the same, so I think this can be left as it is (but good to know about this tool :+1:).

twalsh-ebi (approved):

PR #395, opened 2022-01-28
Examples related to the Pythonistas presentation on pytest

ipilizota:
* A file with the code I presented
* A file with more examples

They will be referenced in the presentation & alongside the presentation on Confluence.
The idea is to use them as templates for future tests. 

Making  a PR so we're sure this is where they should go.
And to hear what Travis thinks about them, of course.

P.S. I called the dir `unittests` to avoid confusion with the library `unittest`.


src/python/examples/unittests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,303 @@
...
+            ),
+            ([], "test", "gene", raises(ValueError, match=r"No cores to dump."))
+        ]
+    )
+    def test_dump_genomes(self, core_list: List[str], species_set_name: str,
+                          tmp_dir: Path, id_type: str, expectation: ContextManager) -> None:
+        """Tests :func:`orthology_benchmark.dump_genomes()` when server connection can be established.
+
+        Args:
+            species_list: A list of species (genome names).
```
JAlvarezJarreta, line 118:
```suggestion
            core_list: A list of core database names.
```

src/python/examples/unittests/test_data_dumps.py
```diff
@@ -0,0 +1,193 @@
...
+            ),
+            ([], "test", "gene", pytest.raises(ValueError))
+        ]
+    )
+    def test_dump_genomes(self, core_list: List[str], species_set_name: str,
+                          tmp_dir: Path, id_type: str, expectation: ContextManager) -> None:
+        """Tests :func:`data_dumps.dump_genomes()` when server connection can be established.
+
+        Args:
+            species_list: A list of species (genome names).
```
JAlvarezJarreta, line 161:
```suggestion
            core_list: A list of core db names.
```

src/python/examples/unittests/test_data_dumps.py
```diff
@@ -0,0 +1,193 @@
...
+            species_set_name: Species set (collection) name.
+            tmp_dir: Unit test temp directory (fixture).
+            expectation: Context manager for the expected exception, i.e. the test will only pass if that
```
JAlvarezJarreta, line 164:
```suggestion
            id_type: Type of identifier to use in the dumps.
            expectation: Context manager for the expected exception, i.e. the test will only pass if that
```

JAlvarezJarreta (changes requested):
The workaround that Thomas did to import these scripts may be quite hard to understand/follow to many people, but I still need some more time to prepare and present an alternative to this.

I have checked mainly the docstrings to ensure readibility, as I expect the rest to be as good as it was in the original branch :) The only other change I would propose is `unit_tests` instead of `unittests` in the folder name, to make it even more clear what we are talking about.

src/python/examples/unittests/test_data_dumps.py
```diff
@@ -0,0 +1,193 @@
...
+def test_find_latest_core_naive() -> None:
+    """Tests :func:`data_dumps.find_latest_core()` function when it does not raise an error.
+
+    Args:
+        core_names: A list of core database names.
+
+    """
+    assert data_dumps.find_latest_core(
+        ["mus_musculus_core_105_1", "mus_musculus_core_52_105_3", "mus_musculus_core_104_4"]
+            ) == "mus_musculus_core_52_105_3"
```
twalsh-ebi, line 67: I think assigning to a `core_names` variable makes it easier to compare this to the parametrised version of the test:
```suggestion
    core_names = ["mus_musculus_core_105_1", "mus_musculus_core_52_105_3", "mus_musculus_core_104_4"]
    assert data_dumps.find_latest_core(core_names) == "mus_musculus_core_52_105_3"
```

src/python/examples/unittests/test_data_dumps.py
```diff
@@ -0,0 +1,193 @@
...
+        """
+        with pytest.raises(RuntimeError):
+            data_dumps.dump_genomes(["mus_musculus", "naja_naja"], "fake",
+                                             "fake-host", 65536, tmp_dir, "protein")
+
+    def test_dump_genomes_fake_output_path(self) -> None:
+        """Tests :func:`data_dumps.dump_genomes()` with fake output path."""
+        with pytest.raises(OSError):
+            data_dumps.dump_genomes(["mus_musculus", "naja_naja"], "default",
+                                             self.host, self.port, "/compara", "protein")
```
twalsh-ebi, line 193: I'd be wary of assuming that a path does not exist in the root directory, but if it must be done, I'd suggest at least renaming it to make its purpose clear: 
```suggestion
                                             self.host, self.port, "/nonexistent/path", "protein")
```

src/python/examples/unittests/test_orthology_benchmark.py
```diff
@@ -0,0 +1,303 @@
...
+        """
+        with raises(RuntimeError):
+            orthology_benchmark.dump_genomes(["mus_musculus", "naja_naja"], "fake",
+                                             "fake-host", 65536, tmp_dir, "protein")
+
+    def test_dump_genomes_fake_output_path(self) -> None:
+        """Tests :func:`orthology_benchmark.dump_genomes()` with fake output path."""
+        with raises(OSError, match=r"Failed to create '/compara/default' directory."):
+            orthology_benchmark.dump_genomes(["mus_musculus", "naja_naja"], "default",
+                                             self.host, self.port, "/compara", "protein")
```
twalsh-ebi, line 150: As above, I'd suggest at least renaming the nonexistent path to be as explicit as possible about its purpose:
```suggestion
        with raises(OSError, match=r"Failed to create '/nonexistent/path/default' directory."):
            orthology_benchmark.dump_genomes(["mus_musculus", "naja_naja"], "default",
                                             self.host, self.port, "/nonexistent/path", "protein")
```

twalsh-ebi (commented):
I'll defer to @JAlvarezJarreta on the question of where these examples should go.

In terms of the code itself, I just have a couple of suggestions.

ipilizota:
All suggestions accepted.

ipilizota:
Travis is failing due to reasons not related to this PR.

twalsh-ebi (approved):

JAlvarezJarreta (approved):

PR #403, opened 2022-03-01
Use Ensembl-default parameters in hal_to_maf.py

twalsh-ebi:
## Description

The `hal_to_maf.py` script does not currently expose the `hal2maf` options `--maxBlockLen` and `--maxRefGap`,
nor does it use the default values used in existing Ensembl Compara code. These should be configured and set with
default values yielding output MAF blocks that are more consistent with existing Compara code for exporting genomic
alignment blocks from a HAL file.

**Related JIRA tickets:**
- ENSCOMPARASW-5290

## Overview of changes

Two new parameters have been added: `--max-block-length` and `--max-ref-gap`, corresponding to `hal2maf` parameters `--maxBlockLen` and `--maxRefGap`, respectively. Their default values are assigned consistently with existing Compara code for exporting MAF alignments from a HAL file.

Parameter `--min-block-size` has been renamed to `--min-block-length` for consistency with `--max-block-length`.

## Testing

The script was tested with reference genome GRCh38 and reference sequence chr19 in two test runs using the old and new default parameters, respectively.

Using the old default parameters:
```
python hal_to_maf.py --ref-genome GRCh38 --ref-sequence chr19 --genomes-file genomes.txt \
  --max-block-length 1000 --max-ref-gap 0 input.hal output_old_params.maf
```

Using the new default parameters:
```
python hal_to_maf.py --ref-genome GRCh38 --ref-sequence chr19 --genomes-file genomes.txt \
  --max-block-length 1000000 --max-ref-gap 500 input.hal GRCh38_chr19_new_params.maf
```

Bar plots of the distributions of the number of MAF blocks in each of the two output files show that the new default parameters have yielded a smaller number of longer MAF blocks, as expected.

### DIFF
--- a/scripts/hal_alignment/hal_to_maf.py
+++ b/scripts/hal_alignment/hal_to_maf.py
@@ -45,12 +45,34 @@ if __name__ == '__main__':
 
     parser.add_argument('--min-block-seqs', metavar='INT', type=int, default=3,
                         help="Minimum number of sequences per block.")
-    parser.add_argument('--min-block-size', metavar='INT', type=int, default=200,
+    parser.add_argument('--min-block-length', metavar='INT', type=int, default=200,
                         help="Minimum number of columns per block.")
+    parser.add_argument('--max-block-length', metavar='INT', type=int,
+                        help="Maximum number of columns per block.")
+    parser.add_argument('--max-ref-gap', metavar='INT', type=int,
+                        help="Maximum reference sequence gap length.")
     args = parser.parse_args()
 
     with open(args.genomes_file) as f:
         target_genomes = [line.rstrip() for line in f]
+
+    # Ensembl defaults
+    if len(target_genomes) > 1:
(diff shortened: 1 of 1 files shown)


scripts/hal_alignment/hal_to_maf.py
```diff
@@ -45,20 +45,44 @@
...
+
     command = [
         'hal2maf',
          args.hal_file,
         'stdout',
         '--refGenome', args.ref_genome,
         '--refSequence', args.ref_sequence,
-        '--targetGenomes', ','.join(target_genomes)
+        '--targetGenomes', ','.join(target_genomes),
+        '--maxBlockLen', str(args.max_block_length),
```
CristiGuijarro, line 84:
```suggestion
        '--maxBlockLength', str(args.max_block_length),
```
I just feel like this could be the complete word here

CristiGuijarro (approved):
:+1:

PR #404, opened 2022-03-02
New reference selector with query results directory

CristiGuijarro:
## Description

Require a script to run and select appropriate reference collection to query genome in Orthofinder pipeline, and manage the copy/creation of the appropriate directory. Works as a factory and post-factory set-up.

**Related JIRA tickets:**
- ENSCOMPARASW-5042

#### Change 1
- New script using compara taxonomy utils 

#### Change 2
- New pytest to test the script

## Testing
As above


scripts/pipeline/appropriate_ref_collection_fasta.py
```diff
@@ -0,0 +1,118 @@
...
+from sqlalchemy.orm import Session
+
+from ensembl.database import DBConnection
+from ensembl.compara.utils.taxonomy import (
+    collect_taxonomys_from_path,
+    match_taxon_to_reference,
+)
+
+
+def _get_collection_path(session: Session, taxon: str, ref_dir: Path) -> Path:
```
dthybert, line 47: the functions starting by underscore is a convention used in python to denominate private methods.  I think we need an harmonisation to define if we need to use this convention in function defined in script or in class definition only. So far in the different scripts I reviewed I did not see this for scripts. But happy to get opinion form other team members.

scripts/pipeline/appropriate_ref_collection_fasta.py
```diff
@@ -0,0 +1,118 @@
...
+        taxon: Scientific taxon name as in ncbi_taxonomy
+        dir: collections directory containing directories
+            named by taxonomic classification
+    """
+    refs_dir = collect_taxonomys_from_path(session, ref_dir)
+    ref_taxon = match_taxon_to_reference(session, taxon, refs_dir)
+    return os.path.join(ref_dir, ref_taxon)
+
+
+def _duplicate_collection_path(taxon: str, src: Path, dest: Path) -> Path:
```
dthybert, line 60: same remark about under score

scripts/pipeline/appropriate_ref_collection_fasta.py
```diff
@@ -0,0 +1,118 @@
...
+    Args:
+        taxon: Scientific taxon name as in ncbi_taxonomy
+        src: Collections directory named by taxonomic classification
+        dest: Path of new directory parent
+    """
+    dest = os.path.join(dest, taxon)
+    try:
+        new_collection = copytree(src, dest)
+        return new_collection
+    except FileExistsError:
```
dthybert, line 72: Is it possible that the dest directory exists but the references are not the same? If this is the case you would process the data with the wrong references set?

CristiGuijarro: It should not be possible in our very specific use-case for this script. Although I am not sure what else this script would be used for. It depends on if we would use this script for any other purpose than to be run in a workflow in which the dest dir is a temporary dir as it is in this case (as it heads to short-term storage). To note than in our use-case the dest dir is to be removed near the end of the pipeline - it is an intermediate step.

scripts/pipeline/appropriate_ref_collection.py
```diff
@@ -0,0 +1,80 @@
...
+    args = parser.parse_args(sys.argv[1:])
+
+    try:
+        dbc = DBConnection(args.url)
+        with dbc.session_scope() as sesh:
+            collection_path = get_collection_path(
+                sesh, args.taxon_name, args.ref_base_dir
+            )
+            print(collection_path)
+    except (TypeError, AttributeError, NoResultFound):
```
dthybert, line 79: I would either throw an exception here or stop with sys.exit(1) . This will give an error code for the end of the script at the system level. In this way the pipeline will stop at this error. in the way as it is done now in case of error the script will finished as if it does not have any error.

dthybert (changes requested):

dthybert (approved):

thiagogenez (approved):
All looks good to me

PR #405, opened 2022-03-03
A function to read in OrthoFinder predictions for a pair of species

ipilizota:
## Description

In order to calculate GOC scores for a pair of species, it's necessary to have OrthoFinder pairwise orthology predictions for the two species. This function parses OrthoFinder output and returns a list of those predictions.

**Related JIRA tickets:**
- ENSCOMPARASW-4972

## Overview of changes
- A new function and corresponding `pytest`.

## Testing
- corresponding `pytest`
- I also tested the function in the Python interpreter on an existing OrthoFinder output that I have:
```
input_dir = '/hps/nobackup/flicek/ensembl/compara/ivana/test_gtfs_orthofinder'
predictions = extract_orthologs(input_dir, "gallus_gallus", "homo_sapiens")
```
Looks ok.

## Notes
1. I could've also checked if the `*tsv` file existed in `*species2*species1*` directory but if it doesn't exist in `*species1*species2*` that's a sign that something went wrong in the OrthoFinder run/writing output. Maybe better to stop immediately.
2. Line 310 in the test script had to be added due to `pylint` complaints.


scripts/pipeline/orthology_benchmark.py
```diff
@@ -307,8 +307,51 @@ def run_orthology_tools(input_dir: str, orthofinder_params: str) -> None:
...
+        FileNotFoundError: If there is no `OrthoFinder/Results_*` directory in `input_dir`.
+
+    """
+    orthologs = []
+
+    try:
+        orthofinder_res = [f.path for f in os.scandir(os.path.join(input_dir, "OrthoFinder"))
+                           if f.is_dir() and f.name.startswith("Results")]
+    except FileNotFoundError as e:
+        raise FileNotFoundError("Could not find OrthoFinder output.") from e
```
dthybert, line 332: in this case I do not think you should catch the exception as the only think you are doing is adding is "Could not find OrthoFinder output" to the exception message and you raise it again. This is not crucial for a better understanding of the exception's default message. So I would either not catch the exception or just stop the script here (sys.exit() with the error message.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -307,8 +307,51 @@ def run_orthology_tools(input_dir: str, orthofinder_params: str) -> None:
...
+
+    latest_res = max(orthofinder_res, key=os.path.getmtime)
```
dthybert, line 334: I would add a comment here for why you select the lastest time. This is so specific to orthofinfer directory structure that in a few month/years we may forget why this is done that way

scripts/pipeline/orthology_benchmark.py
```diff
@@ -307,8 +307,51 @@ def run_orthology_tools(input_dir: str, orthofinder_params: str) -> None:
...
+            if "Orthogroup" in line:
+                continue
+
+            temp = line.replace("\n", "").split("\t")
+            genes1 = temp[1].replace(" ", "").split(",")
+            genes2 = temp[2].replace(" ", "").split(",")
+
+            for gene1 in genes1:
+                for gene2 in genes2:
+                    orthologs.append((gene1, gene2))
```
dthybert, line 352: Just a comment. Do you think it would be useful to have the type of orthology relation in this (one2one, one2many...)

dthybert (changes requested):
I Ivana I have a couple of minor changes and one question related to the type of orthologues. The later one does not necessary have to be done just a question and come be done in a next iteration if required.

scripts/pipeline/orthology_benchmark.py
dthybert: this is fine.

scripts/pipeline/orthology_benchmark.py
dthybert: ok

dthybert (approved):

scripts/pipeline/orthology_benchmark.py
```diff
@@ -307,8 +307,48 @@ def run_orthology_tools(input_dir: str, orthofinder_params: str) -> None:
...
+
+    orthofinder_res = [f.path for f in os.scandir(os.path.join(input_dir, "OrthoFinder"))
+                           if f.is_dir() and f.name.startswith("Results")]
+
+    latest_res = max(orthofinder_res, key=os.path.getmtime)
+
+    predictions_subdir = [f.path for f in os.scandir(os.path.join(latest_res, "Orthologues"))
+                          if f.is_dir() and re.match(f"(.*){species1}_core_(.*)", f.name)][0]
+    predictions_file = [f.path for f in os.scandir(predictions_subdir) if f.is_file() and
+                        re.match(f"(.*){species1}_core_(.*){species2}_core_(.*)", f.name)][0]
```
twalsh-ebi, line 336: Assuming that `species_key1` and `species_key2` were used by OrthoFinder to identify the two species of interest, the `predictions_subdir` and `predictions_file` paths could be composed directly using `os.path.join` or an equivalent method.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -307,8 +307,48 @@ def run_orthology_tools(input_dir: str, orthofinder_params: str) -> None:
...
+
+    with open(predictions_file) as file_handler:
+        for line in file_handler:
+            if "Orthogroup" in line:
+                continue
+
+            temp = line.replace("\n", "").split("\t")
+            genes1 = temp[1].replace(" ", "").split(",")
+            genes2 = temp[2].replace(" ", "").split(",")
```
twalsh-ebi, line 345: This seems like a great opportunity to use the csv module (`import csv`). The DictReader class is particularly useful here, as it will consume the initial header line and use the headings as keys to the dictionary it returns for every data row. Because of this, you could use `species_key1` and `species_key2` to access the respective fields. I've also suggested using the comma-space string as subfield delimiter to simplify the process of getting individual gene IDs.
```suggestion
    with open(predictions_file) as file_handle:
        reader = csv.DictReader(file_handle, delimiter="\t")
        for row in reader:
            genes1 = row[species_key1].split(", ")
            genes2 = row[species_key2].split(", ")
```

scripts/pipeline/orthology_benchmark.py
```diff
@@ -351,5 +391,4 @@ def calculate_goc_scores():
     prep_input_for_orth_tools(os.path.join(args.out_dir, args.species_set), args.orthology_input)
     print("Running orthology inference tools...")
     run_orthology_tools(args.orthology_input, args.orthofinder_params)
```
twalsh-ebi, line 393: Ideally we would know the exact path of the results directory so that we could pass this to `extract_orthologs`. The following suggestion (requiring `import datetime` and `import shlex`) would allow us to do this by: 1) checking that `--name` and `--output` parameters are not specified in `args.orthofinder_params`, since our approach assumes these are not set by the user; 2) setting an OrthoFinder results directory suffix based on an ISO 8601-formatted string of the current date and time; and 3) checking for a results directory uniquely matching the requested results directory name. This approach would necessitate user intervention in some situations, but in those cases the existing solution would arguably require careful monitoring anyway.
```suggestion
    for unsupported_flagset in [("-n", "--name"), ("-o", "--output")]:
        if any(x in unsupported_flagset for x in shlex.split(args.orthofinder_params)):
            raise RuntimeError(
                f"argument --orthofinder_params: '{'/'.join(unsupported_flagset)}' not supported")
    res_ts = datetime.datetime.now().strftime("%Y%m%dT%H%M%S")
    orthofinder_params = f"{args.orthofinder_params} --name {res_ts}"

    run_orthology_tools(args.orthology_input, orthofinder_params)

    res_path_spec = os.path.join(args.orthology_input, "OrthoFinder", f"Results_{res_ts}*")
    res_path, *clashing_paths = glob.glob(res_path_spec)
    if clashing_paths:
        raise RuntimeError(f"multiple directories match OrthoFinder results path spec: '{res_path_spec}'")
```

src/python/tests/test_orthology_benchmark.py
```diff
@@ -301,3 +301,12 @@ def test_prepare_gtf_files(core_names: str, tmp_dir: Path, expectation: ContextM
...
         assert file_cmp( tmp_dir / "Anopheles_albimanus.AalbS2.51.gtf", exp_out2)
+
+
+def test_extract_orthologs() -> None:
+    """Tests :func:`orthology_benchmark.extract_orthologs()` function.
+    """
+    # pylint: disable-next=no-member
+    test_files_dir = pytest.files_dir / "orth_benchmark"  # type: ignore[attr-defined]
+    assert orthology_benchmark.extract_orthologs(test_files_dir , "gallus_gallus", "homo_sapiens") == \
+           [("ENSGALG00000030005", "ENSG00000147255")]
```
twalsh-ebi, line 312: I would suggest to use the backtick sparingly for line continuation. You could break the line at the start of the expected output list:
```suggestion
    assert orthology_benchmark.extract_orthologs(test_files_dir, "gallus_gallus", "homo_sapiens") == [
        ("ENSGALG00000030005", "ENSG00000147255")
    ]
```
Alternatively, you could assign the observed and expected output to variables and compare these:
```suggestion
    obs_output = orthology_benchmark.extract_orthologs(test_files_dir, "gallus_gallus", "homo_sapiens")
    exp_output = [("ENSGALG00000030005", "ENSG00000147255")]
    assert obs_output == exp_output
```
In fact, there are multiple ways to write this `assert` statement that would all make `pylint` happy. So in this case, I would ask, which of these alternatives would best communicate what is happening to my human colleagues?

twalsh-ebi (changes requested):
The pipeline is really taking shape here. I just have a few comments and suggestions.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -307,8 +310,32 @@ def run_orthology_tools(input_dir: str, orthofinder_params: str) -> None:
         raise RuntimeError(msg) from exc
 
 
-def prep_input_for_goc():
-    """Docstring"""
+def extract_orthologs(res_dir: str, species_key1: str, species_key2: str) -> List[Tuple[str, str]]:
+    """Returns a list of putative orthologous pairs inferred by OrthoFinder for a specified pair
+    of species identificators used in the OrthoFinder analysis.
```
twalsh-ebi, line 315: Ideally the summary line of a multi-line docstring should fit on one line. One way to achieve this would be to summarise the function here and add a `Returns` section into which you could put the details of the return value that are currently given here.

twalsh-ebi (approved):
All looks well here. 👍  I've made an optional suggestion about the `extract_orthologs` docstring.

PR #407, opened 2022-03-07
Centralise repository's test data

JAlvarezJarreta:
## Description

Moved all test files (both databases and flatfiles from Perl and Python) into a single central place in `src/test_data`. For Perl, all folders have been symlinked keeping the existing structure. In Python I have opted to change the path in the code instead (since it is in a single place, whilst I fear it is not the same for Perl).

## Testing
I have run both Perl and Python unit tests and all have passed in Codon.


ipilizota (approved):

twalsh-ebi (approved):

JAlvarezJarreta:
I can confirm the patching script works as it was before :fireworks:

thiagogenez (approved):
Looks good to me.

thiagogenez:
By the way, is it possible to make `mypy` to suppress that error?

JAlvarezJarreta:
> By the way, is it possible to make `mypy` to suppress that error?

On it on the `main` branch :+1:

PR #409, opened 2022-03-17
A function to read in OrthoFinder paralogs for a species

ipilizota:
## Description

In order to calculate GOC scores, it's necessary to have OrthoFinder pairwise paralogs for a species so tandem repeats can be discarded. This function parses OrthoFinder output and returns a list of paralogs for a specified species.

**Related JIRA tickets:**
- ENSCOMPARASW-5308

## Overview of changes
- A new function and corresponding `pytest`.

## Testing
- Corresponding `pytest`
- I also tested the function in the Python interpreter on an existing OrthoFinder output that I have:
```
res_dir = "/hps/nobackup/flicek/ensembl/compara/ivana/test_gtfs_orthofinder/OrthoFinder/Results_Jan17"
species_key = "ivana_gallus_gallus_core_106_6"
paralogs = extract_paralogs(res_dir, species_key)
```
Looks ok.

## Notes
Almost copy-paste of https://github.com/Ensembl/ensembl-compara/pull/405 so it should be easy to review 😉

### DIFF
--- a/scripts/pipeline/orthology_benchmark.py
+++ b/scripts/pipeline/orthology_benchmark.py
@@ -357,6 +357,7 @@ def extract_paralogs(res_dir: str, species_key: str) -> List[Tuple[str, str]]:
 
     predictions_file = os.path.join(res_dir, "Gene_Duplication_Events", "Duplications.tsv")
 
+    gene_pattern = re.compile(f'{re.escape(species_key)}_(?P<gene_id>.+)')
     with open(predictions_file) as file_handle:
         reader = csv.DictReader(file_handle, delimiter="\t")
         for row in reader:
@@ -364,7 +365,6 @@ def extract_paralogs(res_dir: str, species_key: str) -> List[Tuple[str, str]]:
             genes1 = row["Genes 1"].split(", ")
             genes2 = row["Genes 2"].split(", ")
 
-            gene_pattern = re.compile(f'{re.escape(species_key)}_(?P<gene_id>.+)')
             if species == species_key:
                 for gene1 in genes1:
                     gene1_match = gene_pattern.fullmatch(gene1)
--- a/scripts/pipeline/orthology_benchmark.py
+++ b/scripts/pipeline/orthology_benchmark.py
@@ -341,6 +341,39 @@ def extract_orthologs(res_dir: str, species_key1: str, species_key2: str) -> Lis
     return orthologs
 
 
+def extract_paralogs(res_dir: str, species_key: str) -> List[Tuple[str, str]]:
+    """Reads in putative paralogs inferred by OrthoFinder.
+
+    Args:
+        res_dir: Path to the directory with OrthoFinder results.
+        species_key: OrthoFinder identificator of the species of interest.
+
+    Returns:
+        A list of putative paralogous pairs for a specified species identificator used in the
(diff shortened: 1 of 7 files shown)


ipilizota:
Travis is failing for reasons not related to the new code itself. A failed test related to this PR is on the test file:
```
not ok 27 - All lines of src/python/tests/flatfiles/orth_benchmark/OrthoFinder/Results_Mar03/Gene_Duplication_Events/Duplications.tsv have the same number of columns
#   Failed test 'All lines of src/python/tests/flatfiles/orth_benchmark/OrthoFinder/Results_Mar03/Gene_Duplication_Events/Duplications.tsv have the same number of columns'
#   at ./travisci/all-housekeeping/tsv.t line 32.
# died: Expected equal number of columns throughout the file. Got:
#       1 13
#       2 7
```
This could be fixed in Travis configuration while ensuring that the new config is compatible with other test `.tsv` files. The fix looks like a bit out of the scope of this PR, especially given other work commitments at the moment.

dthybert (approved):

scripts/pipeline/orthology_benchmark.py
```diff
@@ -341,6 +341,39 @@ def extract_orthologs(res_dir: str, species_key1: str, species_key2: str) -> Lis
...
+        for row in reader:
+            species = row["Species Tree Node"]
+            genes1 = row["Genes 1"].split(", ")
+            genes2 = row["Genes 2"].split(", ")
+
+            if species == species_key:
+                for gene1 in genes1:
+                    for gene2 in genes2:
+                        paralogs.append(
+                            (gene1.split(f"{species_key}_")[1], gene2.split(f"{species_key}_")[1])
```
twalsh-ebi, line 371: Though this may well work here, using `split` could have unexpected results on different data. It might also cause confusion about what is being done here, since the species key is a prefix and not an internal delimiter of the kind that `split` is typically used for.

One way to remove the species-key prefix would be to define a `re.Pattern` object outside the for loops...
```python
prefix_pattern = re.compile(f"^{re.escape(species_key)}_")
```
...and then use its `sub` method on each `gene1` and `gene2` field to replace the pattern with an empty string. For example:
```python
gene1_bare_id = prefix_pattern.sub("", gene1)
```
Then you could append the tuple of bare IDs (e.g. `(gene1_bare_id, gene2_bare_id)`) to the `paralogs` list.

One disadvantage of that suggestion is that it won't raise an exception if the prefix is not found, while the current approach using `split` will complain in that case. If you want to raise an exception in the case that the species key isn't found, you could define a `re.Pattern` object to match the full species-gene pair...
```python
gene_pattern = re.compile(f'^{re.escape(species_key)}_(?P<gene_id>.+)$')
```
...and then take the bare gene ID from the match's capture group. For example:
```python
gene1_match = gene_pattern.match(gene1)
gene1_bare_id = gene1_match["gene_id"]
```
This will raise an exception if the prefix is not found in `gene1`.

twalsh-ebi (approved):
It is true that these changes are very familiar. :-) LGTM, I just have a couple of optional suggestions.

ipilizota:
All current Travis failures are expected and unrelated to this PR.

twalsh-ebi (approved):

PR #410, opened 2022-03-31
OrthoFinder on RR Query Update Nextflow Pipeline

CristiGuijarro:
## Description

Need a NextFlow skeleton for the Update of a reference collection with a query genome

**Related JIRA tickets:**
- ENSCOMPARASW-4694 (various subtasks)

#### Change 1
- Move reuseable/useful code to new pipeline utilities script
- ascii ensembl logo
- JsonParserForSoloEntry - in place for DataFileAPI rest return, but generic enough for use in lots of other stuff

#### Change 2
- Add license to NextFlow files and docstrings to modules and functions/methods
- Add help methods to each main.nf

#### Change 3
- New modules in the `orthofinder_modules.nf` for query
- Rename of workflow files to follow DSL2 convention - `main.nf`
- Included stubs to use in workflow testing and debugging - these can stay for good practice

#### Change 4
- Rework of config files to fit with hybrid deployment and work more agnostically

#### Change 5
- New `main.nf` workflow for the query pipeline skeleton
- New parameter config for specifics - to be changed with the S3 buckets when related wrapper script is in place

## Testing
- Individual modules tested locally on codon
- Workflow tested as entirety in codon on lsf with `-stub-run` and completed

## Notes
- Data copy to FTP not included in skeleton because still needs to be discussed and agreed upon.

Screenshot from 2022-03-31 09-21-02


dthybert (approved):
I had a look at the directory structure for the organisation of these NF files and overall it seems sensible. Maybe the utilities could move in the future in a  utilities directory, but only if add more utilities function and the file have to be subdivided. Since this is the most advance set of pipeline we have I think it does make sense to follow the current organisation and change it if in the future we see that they are limitations.

ipilizota (approved):

PR #412, opened 2022-04-05
Record queries to references per release text files

CristiGuijarro:
## Description

Production team needs to know when a species needs to be re-run. The aim of this PR is to address the need when a reference genome is either updated or retired. In this PR the query half is executed to provide a list of query species for each reference genome. See figure.

**Related JIRA tickets:**
- ENSCOMPARASW-5089
- ENSCOMPARASW-5004

#### Change 1
- New runnable in Perl - required because of incorporation with Perl API and Compara Schema

## Testing
- Run in two pipelines for two species to ensure it works as one would hope.
- [Test 1](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-9&port=4647&dbname=cristig_blastocyst_1_105)
- [Test 2](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-9&port=4647&dbname=cristig_blastocyst_105)
- File output here:
```
/hps/nobackup/flicek/ensembl/compara/shared/species_set_record/105/
```

## Notes
The reference side (not yet written) will parse the genome specific text file output (by release and genome) upon update or retiring of a reference - concatenate the necessary files, uniquely sort them and pass a list of query species that need to be re-run via email to the production team.

species_update_method drawio (3)

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/ENV.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/ENV.pm
@@ -91,7 +91,8 @@ sub shared_default_options {
         'genome_dumps_dir'      => $self->o('shared_hps_dir') . '/genome_dumps/'.$self->o('division').'/',
         'ref_member_dumps_dir'  => $self->o('shared_hps_dir') . '/reference_dumps/',
         'sketch_dir'            => $self->o('shared_hps_dir') . '/species_tree/' . $self->o('division') . '_sketches/',
-
+        # Record of the species that have been run with each reference in RR
+        'rr_species_set_record' => $self->o('shared_hps_dir') . '/species_set_record/' . Bio::EnsEMBL::ApiVersion::software_version() . '/',
         # HMM library
         'hmm_library_version'   => '2',
         'hmm_library_basedir'   => $self->o('shared_hps_dir') . '/treefam_hmms/2019-01-02',
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/HomologyAnnotation_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/HomologyAnnotation_conf.pm
@@ -77,6 +77,8 @@ sub default_options {
         'ref_dump_dir' => $self->o('ref_member_dumps_dir'),
         # Directory for diamond and fasta files for query genome
         'members_dumps_dir' => $self->o('dump_path'),
+        # Directory for species set record of references to query
+        'species_set_record' => $self->o('rr_species_set_record'),
         # Compara schema file path
         'schema_file' => $self->o('schema_file_sql'),
         # Path to db_cmd.pl script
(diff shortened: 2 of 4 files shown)


CristiGuijarro:
> If I understand this based on the diagram and the explanation, the goal of this code is to store for each query genome the list of references so we have the information linking a query genome to the reference? If a reference is modified/retired etc... we can then identify what are the query genome associated to the references?
> 
> If this is the case in the code written in the RecordSpeciesSet.pm how to you link the query genome to the species set? At them moment the blastocyst will copy the references into a basedirectory but there I cannot see the link between query and references?

It's a little bit the other way around. The base directory stores for each release; the list of reference genomes as files, a query genome is added/written to each reference genome file as it is "compared" to it. This creates a list of queries that needs to be re-run when a reference genome is updated or retired. This is all this PR addresses in terms of code. After that, as in the figure, the work in the reference side is to sort/uniquely concatenate and compile the query genomes that are parsed from the relevant reference genome files as and when they are updated/retired. As mentioned previously, this code does not include the update/inclusion of an entirely new species set - that is to be addressed in another ticket, with the use of the taxonomic selection script to re-evaluate previously run queries - and written to the same list of species that need to be re-run to keep it simple for the production team in one place.

dthybert:
> 



> > If I understand this based on the diagram and the explanation, the goal of this code is to store for each query genome the list of references so we have the information linking a query genome to the reference? If a reference is modified/retired etc... we can then identify what are the query genome associated to the references?
> > If this is the case in the code written in the RecordSpeciesSet.pm how to you link the query genome to the species set? At them moment the blastocyst will copy the references into a basedirectory but there I cannot see the link between query and references?
> 
> It's a little bit the other way around. The base directory stores for each release; the list of reference genomes as files, a query genome is added/written to each reference genome file as it is "compared" to it. This creates a list of queries that needs to be re-run when a reference genome is updated or retired. This is all this PR addresses in terms of code. After that, as in the figure, the work in the reference side is to sort/uniquely concatenate and compile the query genomes that are parsed from the relevant reference genome files as and when they are updated/retired. As mentioned previously, this code does not include the update/inclusion of an entirely new species set - that is to be addressed in another ticket with another PR, with the use of the taxonomic selection script to re-evaluate previously run queries - and written to the same list of species that need to be re-run to keep it simple for the production team in one place.

I see that in the way you are specifying it you try to have the less data redundancy as possible. and it is also faster to get the list of species. based on the reference set.  However I see one issue is when two pipeline run in parallel then we can have concurrency issues when writing in the file if the two pipelines try to write to the same file. I know that at the moment only one pipeline is run using batches but in the future this may happen especially if we use orthofinder pipeline in parallel to the homology annotation pipeline. Do you see how we could solve this? could we create a table to a reference database that store this info? Or a separate db? We also need to support the inverse use for a query genome what is the list of references but this can be a separate ticket .

dthybert (approved):

PR #413, opened 2022-04-05
Only delete_tree for corresponsing collection/clusterset with test

CristiGuijarro:
## Description

We only want to update the `gene_member_hom_stats` for the appropriate tree `clusterset_id`/`collection` with  `gene_member`, rather than updating the stats for every instance of the `gene_member` in `gene_member_hom_stats`.

**Related JIRA tickets:**
- ENSCOMPARASW-5031

#### Change 1
- _Updated the update statement in the `GeneTreeAdaptor` `delete_tree()` method to include the `clusterset_id` / `collection`_

#### Change 2
- _Added a test because it is the easiest way to test I didn't break it_

## Testing
_As above_

## Notes
_Have not addressed the issue in which apparently only the last tree version should be updated on delete - because this is not a real issue. There should only be one gene to one tree with each `clusterset_id` at any one time, where this was previously seen as otherwise was a blip._


modules/Bio/EnsEMBL/Compara/DBSQL/GeneTreeAdaptor.pm
```diff
@@ -428,9 +403,19 @@ sub delete_tree {
     $tree->_load_tags;
 
     my $root_id = $tree->root->node_id;
-
+    my $clusterset_id = $tree->clusterset_id;
     # Query to reset gene_member_hom_stats
```
twalsh-ebi, line 407: I think this entire section of code dealing with gene-member homology stats could be enclosed in a conditional block such as the following that checks whether the given tree is a default/merged tree. The advantages of this would be as follows:

- It would be more explicit about the fact that gene-member homology stats are reset _only_ for default/merged gene trees. This may be obvious to someone who is familiar enough with Compara gene-tree pipelines to know that under normal circumstances the `$clusterset_id` matches the `collection` column only in the case of default/merged gene trees, but it will not necessarily be immediately apparent to everyone who will read this code, and it's no harm to provide a helping hand.
- It would avoid executing this code for linked trees, which for a default/merged tree are deleted by a recursive call to `delete_tree`. I wouldn't claim that this is a world-beating optimisation, but why do unnecessary work?
- It would be consistent with the approach taken elsewhere in this method for code that is relevant only to the default/merged tree.
```
    # Only for "default" trees
    unless ($tree->ref_root_id) {
        my $clusterset_id = $tree->clusterset_id;
        # Query to reset gene_member_hom_stats
        my $gene_member_hom_stats_sql = qq/
            UPDATE gene_member_hom_stats
                SET
                    gene_trees = 0,
                    orthologues = 0,
                    paralogues = 0,
                    homoeologues = 0
                WHERE
                    gene_member_id = ?
                        AND collection = "$clusterset_id"
        /;
        for my $leaf (@{$tree->get_all_leaves}) {
            if ($leaf->isa('Bio::EnsEMBL::Compara::GeneTreeMember')) {
                $self->dbc->do($gene_member_hom_stats_sql, undef, $leaf->gene_member_id);
            }
        }
    }
```

modules/t/geneTreeAdaptor.t
```diff
@@ -0,0 +1,52 @@
...
+    $gene_tree_adaptor->delete_tree($gene_tree_default);
+    $gene_member = $compara_dba->get_GeneMemberAdaptor->fetch_by_stable_id("ENSTSYG00000021671");
+    my $default_after = $gene_member->has_GeneTree("default");
+    my $other_after = $gene_member->has_GeneTree("other");
+    is($default_after, 0, "Num default tree after delete");
+    is($other_after, $other_before, "Num other tree after default delete");
+
+    done_testing();
+};
+
```
twalsh-ebi, line 51: This test looks almost trivial, but it would have failed on the previous version of the `delete_tree` method, and provides additional confirmation that the method is now more robust even to the most unbargained-for blip.
```suggestion

my $gene_member2 = $compara_dba->get_GeneMemberAdaptor->fetch_by_stable_id("ENSXMAG00000006974");
my $gene_tree2_default = $gene_tree_adaptor->fetch_default_for_Member($gene_member2, "default");
my $all_trees = $gene_tree_adaptor->fetch_all_by_Member($gene_member2, -CLUSTERSET_ID => "phyml-aa");
my $gene_tree2_linked = $all_trees->[0];

subtest "Test Bio::EnsEMBL::Compara::DBSQL::GeneTreeAdaptor delete_tree method on linked protein tree", sub {

    my $default_before = $gene_member2->has_GeneTree("default");
    is($default_before, 1, "Default tree status before linked-tree delete");
    $gene_tree_adaptor->delete_tree($gene_tree2_linked);
    $gene_member2 = $compara_dba->get_GeneMemberAdaptor->fetch_by_stable_id("ENSXMAG00000006974");
    my $default_after = $gene_member2->has_GeneTree("default");
    is($default_after, 1, "Default tree status after linked-tree delete");

    done_testing();
};

```

twalsh-ebi (changes requested):
The module changes and the additional unit test look grand to me, though I just have a couple of suggested changes.

modules/Bio/EnsEMBL/Compara/DBSQL/GeneTreeAdaptor.pm
CristiGuijarro: I don't agree. This seems unnecessary and additionally cumbersome. Keeping the code as I have provided it, regardless of the type of tree provides more flexibility in data movement. The code already works to remove the `gene_member_hom_stats` specifically for the gene and the `cluster_set_id` (`collection`). Currently, although we do not provide `gene_member_hom_stats` on the other "contributing" trees within our production analysis, to me the preferred behaviour would be to attempt to consistently and cleanly delete that data related to the tree that is specifically being deleted, and as this code is on the `GeneTreeAdaptor` - it is schema specific, and not so process specific, it should be decoupled from the method of merging/reconstructing supertrees for this particular aspect (I have not looked at the default/merging edge cases you have mentioned, but I might argue against those also). As far as the point made about the recursive call - I would agree, not a world beating optimisation. This PR addresses the issue in the ticket.

modules/t/geneTreeAdaptor.t
CristiGuijarro: As above :+1:

modules/Bio/EnsEMBL/Compara/DBSQL/GeneTreeAdaptor.pm
twalsh-ebi: You've made a lot of interesting points here, and I'd consider several of them to be debatable. But in the interest of keeping the PR review focused and constructive, I'm going to take what I think are your most compelling points and address those.

You argued that it is preferable for the gene-member homology stats to be reset consistently for a deleted tree, regardless of whether or not it's a default/merged tree. This is mentioned in the context of the provision of homology stats for "contributing" trees, though I'm not sure that there's a use case that would justify routinely inferring, storing and delivering homologies for unmerged TreeBeST trees. But you also mentioned the gene-tree overhaul elsewhere in your comment, and I have to admit that in that context you have a point, so I'm happy to defer to your judgement on this.

twalsh-ebi (approved):

PR #414, opened 2022-04-07
replace agrschema with agrparse and add constraint elements in the json files

dthybert:
src/python/scripts/run_gerp.py

## Description
The constraint elements identified by gerpelem have coordinates based on the genomic alignment bloc. To simplified the coordinate transformation required in the next step of the pipeline (concatenation of constraint element for each species) we want to have these constraint elements into the json file describing the start and end of the genomic alignment bloc for each species. the goal of this developemnt is to add a new attribute constraint elements in the json file which will contained the constraints elements.

**Related JIRA tickets:**
- ENSCOMPARASW-5353

## Overview of changes
We had two different changes in this developemnt. First one is switching argument passing system from argschema to argpars. This was not in the jira ticket by argshema was not adapted int this script as discussed with some team members. The second change aims to add the constraint element into the json file

#### Change 1
-  switching argshema to argparse and wrapping the main code of the script into a main fucntion that get as a parameter the argparse.Namespace

#### Change 2
- creation of a function which aim to add the conserved element of a genomic alignment bloc to its corresponding json file and in a attribut called : constraint_elems

## Testing
I tested by using a constraint element file with one and two and 0 constraint elements.
pass pylint with a score of 100%

## Notes
_Optional extra information._

### DIFF
--- a/src/python/scripts/run_gerp.py
+++ b/src/python/scripts/run_gerp.py
@@ -28,11 +28,10 @@ path where they can be found::
     $ python run_gerp.py --msa_file alignment.mfa --tree_file tree.nw --gerp_exe_dir path/to/gerp/
 
 """
-import os
-import subprocess
 import argparse
 import json
-
+import os
+import subprocess
 
 def add_ce_to_json(ce_file: str, json_file: str) -> None:
     """Enrich the json file with the constrained elements.
@@ -42,31 +41,32 @@ def add_ce_to_json(ce_file: str, json_file: str) -> None:
         json_file: json file with genome coordinate level.
 
     """
-    with open(json_file, 'r') as json_file_handler:
-        align_set = json.load(json_file_handler)
+    with open(json_file, 'r') as json_file_obj:
+        align_set = json.load(json_file_obj)
 
-    with open(ce_file) as ce_file_handler:
-        constrained_elements = [ce.split() for ce in ce_file_handler if ce.rstrip() != ""]
-
-    # enrich json file with constrained elements info
     constrained_elems = []
-    for ce in constrained_elements:
-        constrained_elem = {
-            "start": int(ce[0]),
-            "end": int(ce[1]),
-            "length": int(ce[2]),
-            "score": float(ce[3]),
-            "p-val": float(ce[4])
-        }
-        constrained_elems.append(constrained_elem)
-    align_set["constraint_elems"] = constrained_elems
+    with open(ce_file) as ce_file_obj:
+        # enrich json file with constrained elements info
+        for cons_elem in ce_file_obj:
+            if cons_elem.rstrip() == "":
+                continue
+            ce = cons_elem.split()
+            constrained_elem = {
+                "start": int(ce[0]),
+                "end": int(ce[1]),
+                "length": int(ce[2]),
+                "score": float(ce[3]),
+                "p-val": float(ce[4])
+            }
+            constrained_elems.append(constrained_elem)
+    align_set["constrained_elems"] = constrained_elems
 
-    with open(json_file, 'w') as json_file_handler:
-        json.dump(align_set, json_file_handler)
+    with open(json_file, 'w') as json_file_obj:
+        json.dump(align_set, json_file_obj)
 
 
 def main(param: argparse.Namespace) -> None:
-    ''' Main function of the run_gerp.py script
+    """ Main function of the run_gerp.py script
 
     This function is running gerpcol that define a gerpscore for every column of the genomic alignment bloc
     and gerpelem that identify constrained elements across the alignment bloc
@@ -74,9 +74,7 @@ def main(param: argparse.Namespace) -> None:
     Args:
         param: argparse.Namespace storing all the script parameters
 
-    Returns:
-        None
-    '''
+    """
 
     cmd = ["gerpcol", "-t", param.tree_file, "-f", param.msa_file]
     cmd[0] = os.path.join(param.gerp_exe_dir, cmd[0])
@@ -88,7 +86,7 @@ def main(param: argparse.Namespace) -> None:
     cmd += ["-d", str(param.depth_threshold)]
     subprocess.run(cmd, check=True)
 
-    add_ce_to_json(f"{param.msa_file}.rates.elems", param.msa_file.replace(".fa", ".json"))
+    add_ce_to_json(f"{param.msa_file}.rates.elems", f"{os.path.splitext(param.msa_file)[0]}.json")
 
 
 if __name__ == "__main__":
--- a/src/python/scripts/run_gerp.py
+++ b/src/python/scripts/run_gerp.py
@@ -35,12 +35,15 @@ import json
 
 
 def add_ce_to_json(ce_file: str, json_file: str) -> None:
-    '''Enrich the json file with the constraint elements
+    """Enrich the json file with the constraint elements
 
-    :param ce_file: file containing the constraint elements
-    :param json_file: json file with genome coordinate elevel
-    :return: None
-    '''
+    Args:
+        ce_file: file containing the constraint elements.
+        json_file: json file with genome coordinate level.
+
+    Returns:
+        None
+    """
     with open(json_file, 'r') as json_file_handler:
         align_set = json.load(json_file_handler)
 
@@ -69,8 +72,11 @@ def main(param: argparse.Namespace) -> None:
     This function is running gerpcol that define a gerpscore for every column of the genomic alignment bloc
     and gerpelem that identify constraint elements across the alignment bloc
 
-    :param param: argparse.Namespace storing all the script parameters
-    :return: None
+    Args:
+        param: argparse.Namespace storing all the script parameters
+
+    Returns:
+        None
     '''
 
     cmd = ["gerpcol", "-t", param.tree_file, "-f", param.msa_file]
@@ -88,7 +94,7 @@ def main(param: argparse.Namespace) -> None:
 
 if __name__ == "__main__":
 
-    parser = argparse.ArgumentParser(description='Process some integers.')
+    parser = argparse.ArgumentParser(description='Calculate GERP score and identify constraint elements.')
     parser.add_argument('--msa_file', type=str, required=True, help='MSA file (MFA format) (REQUIRED)')
     parser.add_argument('--tree_file', type=str, required=True, help='Tree file (Newick format). Must '
                                                 'include every species in the MSA. (REQUIRED)')
--- a/src/python/scripts/run_gerp.py
+++ b/src/python/scripts/run_gerp.py
@@ -97,10 +97,10 @@ if __name__ == "__main__":
                                                 "include every species in the MSA. (REQUIRED)")
     parser.add_argument("--depth_threshold", default=0.5, type=float, help="Constrained elements depth "
                                                 "threshold for shallow columns, in substitutions per site. "
-                                                "By default, 0.5.")
+                                                "The default is 0.5.")
     parser.add_argument("--gerp_exe_dir", default="", type=str,
-                        help="Path where 'gerpcol' and 'gerpelem' binaries can be found."
-                             " By default, these are assumed to be accessible from the $PATH.")
+                        help="Path where 'gerpcol' and 'gerpelem' executable binaries can be found."
+                             " The default is $PATH.")
 
     args = parser.parse_args()
     main(args)
--- a/src/python/scripts/run_gerp.py
+++ b/src/python/scripts/run_gerp.py
@@ -35,7 +35,7 @@ import json
 
 
 def add_ce_to_json(ce_file: str, json_file: str) -> None:
-    """Enrich the json file with the constrained elements
+    """Enrich the json file with the constrained elements.
 
     Args:
         ce_file: file containing the constrained elements.
--- a/src/python/scripts/run_gerp.py
+++ b/src/python/scripts/run_gerp.py
@@ -22,7 +22,7 @@ Typical usage example::
 
     $ python run_gerp.py --msa_file alignment.mfa --tree_file tree.nw
 
-If ``gerpcol`` and/or ``gerpelem`` are not accesible from ``$PATH``, you will need to provide the
+If ``gerpcol`` and/or ``gerpelem`` are not accessible from ``$PATH``, you will need to provide the
 path where they can be found::
 
     $ python run_gerp.py --msa_file alignment.mfa --tree_file tree.nw --gerp_exe_dir path/to/gerp/
@@ -37,7 +37,7 @@ def add_ce_to_json(ce_file: str, json_file: str) -> None:
     """Enrich the json file with the constrained elements.
 
     Args:
-        ce_file: file containing the constrained elements (csv format).
+        ce_file: file containing the constrained elements (space-delimited format).
         json_file: json file with genome coordinate level.
 
     """
--- a/src/python/scripts/run_gerp.py
+++ b/src/python/scripts/run_gerp.py
@@ -42,7 +42,7 @@ def add_ce_to_json(ce_file: str, json_file: str) -> None:
         json_file: json file with genome coordinate level.
 
     """
-    with open(json_file, 'r') as json_file_obj:
+    with open(json_file, "r") as json_file_obj:
         align_set = json.load(json_file_obj)
 
     constrained_elems = []
--- a/src/python/scripts/run_gerp.py
+++ b/src/python/scripts/run_gerp.py
@@ -41,8 +41,6 @@ def add_ce_to_json(ce_file: str, json_file: str) -> None:
         ce_file: file containing the constrained elements.
         json_file: json file with genome coordinate level.
 
-    Returns:
-        None
     """
     with open(json_file, 'r') as json_file_handler:
         align_set = json.load(json_file_handler)
--- a/src/python/scripts/run_gerp.py
+++ b/src/python/scripts/run_gerp.py
@@ -67,7 +67,7 @@ def add_ce_to_json(ce_file: str, json_file: str) -> None:
 
 
 def main(param: argparse.Namespace) -> None:
-    """ Main function of the run_gerp.py script
+    """Main function of the run_gerp.py script
 
     This function runs gerpcol to define a GERP score for every column of the genomic alignment block
     and gerpelem to identify constrained elements across the genomic alignment block.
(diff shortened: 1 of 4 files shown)


src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,75 @@
...
 
+def add_ce_to_json(ce_file: str, json_file: str) -> None:
+    '''Enrich the json file with the constraint elements
 
-class InputSchema(argschema.ArgSchema):
-    """Calculates the Genomic Evolutionary Rate Profiling (GERP) of a multiple sequence alignment (MSA)."""
+    :param ce_file: file containing the constraint elements
+    :param json_file: json file with genome coordinate elevel
+    :return: None
+    '''
```
CristiGuijarro, line 43: In keeping with the rest of the python code base, I think it is preferable to write the docstrings in this format: https://github.com/Ensembl/ensembl-compara/blob/main/scripts/hal_alignment/hal_gene_liftover.py#L53-L62 with the `Args:` and `Returns:`

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,75 @@
...
+    cmd[0] = os.path.join(param.gerp_exe_dir, cmd[0])
+    cmd += ["-d", str(param.depth_threshold)]
     subprocess.run(cmd, check=True)
+
+    add_ce_to_json(f"{param.msa_file}.rates.elems", param.msa_file.replace(".fa", ".json"))
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description='Process some integers.')
```
CristiGuijarro, line 91: I think we can come up with something a little more informative than "Process some integers." :slightly_smiling_face:

CristiGuijarro (changes requested):
Aside from the docstrings - which is really formatting, one request I have is to really distinguish between "constrained" and "constraint" in "constrained elements" - from the evolutionary perspective of the multiple alignments, it should always be "constrained", "constraint" should only be used without "elements" and for the purpose of limitation/restriction in an algorithm of sorts. Having these terms used more appropriately will make the code more readable. Otherwise, appears fine functionally and with change to `argparse` as long as Travis states all tests pass :+1:

dthybert:
> 

Thank you for the explanation ,this has improved my English :)  , I will do the correction accordingly.

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,81 @@
...
-        description="Path where 'gerpcol' and 'gerpelem' binaries can be found. By default, resort to $PATH."
-    )
+    Returns:
+        None
+    """
+    with open(json_file, 'r') as json_file_handler:
+        align_set = json.load(json_file_handler)
 
+    with open(ce_file) as ce_file_handler:
+        constrained_elements = [ce.split() for ce in ce_file_handler if ce.rstrip() != ""]
```
twalsh-ebi, line 51: The constrained elements file is being processed here in two for loops, but these could be merged into one for loop in which each iteration involves reading and processing one line from the input file.

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,81 @@
...
-    mod = argschema.ArgSchemaParser(schema_type=InputSchema)
+    # enrich json file with constrained elements info
+    constrained_elems = []
+    for ce in constrained_elements:
+        constrained_elem = {}
+        constrained_elem["start"] = ce[0]
+        constrained_elem["end"] = ce[1]
+        constrained_elem["length"] = ce[2]
+        constrained_elem["score"] = ce[3]
+        constrained_elem["p-val"] = ce[4]
```
twalsh-ebi, line 61: I'd advocate converting the constrained element field values to their respective numeric types here, so that they'll be stored as those types in the output JSON file.

I can see this is not a simple change though, because when I tried to implement it as follows...
```suggestion
        constrained_elem = {}
        constrained_elem["start"] = int(ce[0])
        constrained_elem["end"] = int(ce[1])
        constrained_elem["length"] = int(ce[2])
        constrained_elem["score"] = float(ce[3])
        constrained_elem["p-val"] = float(ce[4])
```
...I encountered the `mypy` error:
```
error: Incompatible types in assignment (expression has type "float", target has type "int")
```
It looks like this is happening because `mypy` infers the dictionary value type to be `int` based on the types of the `start`, `end` and `length` fields.

One workaround (requiring us to `import typing`) would be to add a variable annotation to the `constrained_elem` dictionary, to explicitly say that the dictionary value could be either an `int` or `float`:
```suggestion
        constrained_elem: typing.Dict[str, typing.Union[float, int]] = {}
        constrained_elem["start"] = int(ce[0])
        constrained_elem["end"] = int(ce[1])
        constrained_elem["length"] = int(ce[2])
        constrained_elem["score"] = float(ce[3])
        constrained_elem["p-val"] = float(ce[4])
```

A second approach would be to set all the dictionary values in a single statement, so that `mypy` correctly infers the dictionary value type:
```suggestion
        constrained_elem = {
            "start": int(ce[0]),
            "end": int(ce[1]),
            "length": int(ce[2]),
            "score": float(ce[3]),
            "p-val": float(ce[4])
        }
```

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,81 @@
...
+        constrained_elems.append(constrained_elem)
+    align_set["constraint_elems"] = constrained_elems
```
twalsh-ebi, line 63: I'd suggest changing this for consistency with the changes suggested previously by @CristiGuijarro:
```suggestion
    align_set["constrained_elems"] = constrained_elems
```

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,81 @@
...
+
+    with open(json_file, 'w') as json_file_handler:
+        json.dump(align_set, json_file_handler)
```
twalsh-ebi, line 66: As above, I'd suggest changing `json_file_handler` to something like `json_file_handle`, `json_file_object` or `json_file_obj`.

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,81 @@
...
+
+
+def main(param: argparse.Namespace) -> None:
+    ''' Main function of the run_gerp.py script
```
twalsh-ebi, line 70: I'd suggest using a triple double-quoted string and ending with a period to be consistent with the docstring convention in PEP 257.
```suggestion
    """Main function of the run_gerp.py script.
```

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,81 @@
...
+
+    Args:
+        param: argparse.Namespace storing all the script parameters
 
-    cmd = ["gerpcol", "-t", mod.args["tree_file"], "-f", mod.args["msa_file"]]
-    if "gerp_exe_dir" in mod.args:
-        cmd[0] = os.path.join(mod.args['gerp_exe_dir'], cmd[0])
+    Returns:
+        None
+    '''
```
twalsh-ebi, line 80: This would need to be changed if you accept the suggestion above to use a triple double-quoted docstring.
```suggestion
    """
```

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,81 @@
...
+    and gerpelem that identify constrained elements across the alignment bloc
+
+    Args:
+        param: argparse.Namespace storing all the script parameters
 
-    cmd = ["gerpcol", "-t", mod.args["tree_file"], "-f", mod.args["msa_file"]]
-    if "gerp_exe_dir" in mod.args:
-        cmd[0] = os.path.join(mod.args['gerp_exe_dir'], cmd[0])
+    Returns:
+        None
```
twalsh-ebi, line 79: As above, I would omit the `Returns:` section of this docstring.
```suggestion
```

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,81 @@
...
+    '''
+
+    cmd = ["gerpcol", "-t", param.tree_file, "-f", param.msa_file]
+    cmd[0] = os.path.join(param.gerp_exe_dir, cmd[0])
```
twalsh-ebi, line 83: I initially didn't understand why you'd changed the default value of `gerp_exe_dir` to an empty string, but I see now that it allows for a more elegant solution here. 👍

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,81 @@
...
-    if "gerp_exe_dir" in mod.args:
-        cmd[0] = os.path.join(mod.args['gerp_exe_dir'], cmd[0])
-    if "depth_threshold" in mod.args:
-        cmd += ["-d", str(mod.args["depth_threshold"])]
+    cmd = ["gerpelem", "-f", f"{param.msa_file}.rates"]
+    cmd[0] = os.path.join(param.gerp_exe_dir, cmd[0])
+    cmd += ["-d", str(param.depth_threshold)]
     subprocess.run(cmd, check=True)
+
+    add_ce_to_json(f"{param.msa_file}.rates.elems", param.msa_file.replace(".fa", ".json"))
```
twalsh-ebi, line 92: I advise against using the `str.replace` method to replace the FASTA file extension with a JSON file extension. Using the `str.replace` method in this way would replace any occurrence of the substring `.fa` in the string, and if the MSA file has any other FASTA extension (e.g. `.fasta`, `.mfa`), it will quietly pass back the MSA filename unmodified. 

Another way to approach this would be to replace the FASTA extension using an anchored regex that includes all the FASTA file extensions that might be reasonably expected (e.g. `re.sub("(?<=.)\.(fa|fasta|fna|mfa)$", "", filename)`), though this still won't work if an unfamiliar FASTA extension is used.

A second alternative would be to use the `os.path.splitext` function to remove the file extension. This would have the advantage of not having to know the file extension in advance.

A third option would be to add a `--json_file` argument to the script, so that the user would specify the JSON file explicitly. This puts a slight extra burden on the caller, but it also gives them the freedom to name the JSON file however they like, and it removes a possible source of error from the script.

twalsh-ebi (changes requested):
It's nice to see the extra functionality being added. 👍  I've included some suggestions and comments.

src/python/scripts/run_gerp.py
dthybert: yes I found it easer in this way

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,78 @@
...
+    cmd[0] = os.path.join(param.gerp_exe_dir, cmd[0])
+    cmd += ["-d", str(param.depth_threshold)]
     subprocess.run(cmd, check=True)
+
+    add_ce_to_json(f"{param.msa_file}.rates.elems", f"{os.path.splitext(param.msa_file)[0]}.json")
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description='Calculate GERP score and identify constraint elements.')
```
CristiGuijarro, line 94:
```suggestion
    parser = argparse.ArgumentParser(description='Calculate GERP score and identify constrained elements.')
```

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,78 @@
...
-if __name__ == "__main__":
-    mod = argschema.ArgSchemaParser(schema_type=InputSchema)
 
-    cmd = ["gerpcol", "-t", mod.args["tree_file"], "-f", mod.args["msa_file"]]
-    if "gerp_exe_dir" in mod.args:
-        cmd[0] = os.path.join(mod.args['gerp_exe_dir'], cmd[0])
+def main(param: argparse.Namespace) -> None:
+    """ Main function of the run_gerp.py script
+
+    This function is running gerpcol that define a gerpscore for every column of the genomic alignment bloc
```
CristiGuijarro, line 71:
```suggestion
    This function runs gerpcol to define a GERP score for every column of the genomic alignment block
```

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,78 @@
...
+import json
 import os
 import subprocess
 
-import argschema
+def add_ce_to_json(ce_file: str, json_file: str) -> None:
+    """Enrich the json file with the constrained elements.
 
+    Args:
+        ce_file: file containing the constrained elements.
```
CristiGuijarro, line 40: It would be helpful to specify the type/format of the `ce_file`, for example, is it a `tsv` file? A `csv` file?

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,78 @@
...
-    mod = argschema.ArgSchemaParser(schema_type=InputSchema)
 
-    cmd = ["gerpcol", "-t", mod.args["tree_file"], "-f", mod.args["msa_file"]]
-    if "gerp_exe_dir" in mod.args:
-        cmd[0] = os.path.join(mod.args['gerp_exe_dir'], cmd[0])
+def main(param: argparse.Namespace) -> None:
+    """ Main function of the run_gerp.py script
+
+    This function is running gerpcol that define a gerpscore for every column of the genomic alignment bloc
+    and gerpelem that identify constrained elements across the alignment bloc
```
CristiGuijarro, line 72:
```suggestion
    and gerpelem to identify constrained elements across the genomic alignment block.
```

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,78 @@
...
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description='Calculate GERP score and identify constraint elements.')
+    parser.add_argument('--msa_file', type=str, required=True, help='MSA file (MFA format) (REQUIRED)')
+    parser.add_argument('--tree_file', type=str, required=True, help='Tree file (Newick format). Must '
+                                                'include every species in the MSA. (REQUIRED)')
+    parser.add_argument('--depth_threshold', default=0.5, type=float, help='Constrained elements depth '
+                                                'threshold for shallow columns, in substitutions per site. '
+                                                'By default, 0.5.')
```
CristiGuijarro, line 100:
```suggestion
                                                'The default is 0.5.')
```

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,78 @@
...
+    parser.add_argument('--gerp_exe_dir', default="", type=str, help='Path where "gerpcol" and "gerpelem" '
+                                                'binaries can be found. By default, resort to $PATH.')
```
CristiGuijarro, line 102:
```suggestion
    parser.add_argument('--gerp_exe_dir', default="", type=str, help='Path to "gerpcol" and "gerpelem" '
                                                'executable binaries. The default is $PATH.')
```

CristiGuijarro (commented):
Just added a few doc string etc edit suggestions

src/python/tests/conftest.py
```diff
@@ -70,7 +70,7 @@ def test_method(..., dir_cmp: DirCmp, ...):
     """
     # Get the source and temporary absolute paths for reference and target root directories
     ref = Path(request.param['ref'])  # type: ignore[attr-defined]
-    ref_src = ref if ref.is_absolute() else pytest.files_dir / ref  # type: ignore[attr-defined]
+    ref_src = ref if ref.is_absolute() else pytest.files_dir / ref  # type: ignore[attr-defined,operator]
     ref_tmp = Path(tmp_dir) / str(ref).replace(os.path.sep, '_')
     target = Path(request.param['target'])  # type: ignore[attr-defined]
     target_src = target if target.is_absolute() else pytest.files_dir / target  # type: ignore[attr-defined]
```
twalsh-ebi, line 76:
```suggestion
    target_src = target if target.is_absolute() else pytest.files_dir / target  # type: ignore[attr-defined,operator]
```

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,78 @@
...
+
+    parser = argparse.ArgumentParser(description="Calculate GERP score and identify constrained elements.")
+    parser.add_argument("--msa_file", type=str, required=True, help="MSA file (MFA format) (REQUIRED)")
+    parser.add_argument("--tree_file", type=str, required=True, help="Tree file (Newick format). Must "
+                                                "include every species in the MSA. (REQUIRED)")
+    parser.add_argument("--depth_threshold", default=0.5, type=float, help="Constrained elements depth "
+                                                "threshold for shallow columns, in substitutions per site. "
+                                                "By default, 0.5.")
+    parser.add_argument("--gerp_exe_dir", default="", type=str, help="Path where 'gerpcol' and 'gerpelem' "
+                                                "binaries can be found. By default, resort to $PATH.")
```
twalsh-ebi, line 102:
```suggestion
    parser.add_argument("--gerp_exe_dir", default="", type=str,
                        help="Path where 'gerpcol' and 'gerpelem' binaries can be found."
                             " By default, these are assumed to be accessible from the $PATH.")
```

src/python/scripts/run_gerp.py
```diff
@@ -28,45 +28,78 @@
...
-    gerp_exe_dir = argschema.fields.InputDir(
-        required=False,
-        description="Path where 'gerpcol' and 'gerpelem' binaries can be found. By default, resort to $PATH."
-    )
+    constrained_elems = []
+    with open(ce_file) as ce_file_obj:
+        # enrich json file with constrained elements info
+        for cons_elem in ce_file_obj:
+            if cons_elem.rstrip() == "":
+                continue
```
twalsh-ebi, line 52: Is it a frequent occurrence for a constrained element file to contain empty lines? Would an empty line be output during a normal run, or could it potentially be a sign that something has gone wrong?

CristiGuijarro (approved):

src/python/scripts/run_gerp.py
```diff
@@ -22,51 +22,86 @@
...
+                "length": int(ce[2]),
+                "score": float(ce[3]),
+                "p-val": float(ce[4])
+            }
+            constrained_elems.append(constrained_elem)
+    align_set["constrained_elems"] = constrained_elems
 
-if __name__ == "__main__":
-    mod = argschema.ArgSchemaParser(schema_type=InputSchema)
+    with open(json_file, 'w') as json_file_obj:
```
twalsh-ebi, line 65: An optional post-approval suggestion. Must've missed this last time around:
```suggestion
    with open(json_file, "w") as json_file_obj:
```

twalsh-ebi (approved):
It's looking good! I just have one optional suggestion for consistency.

PR #416, opened 2022-04-20
Deprecate/fetch by stable

CristiGuijarro:
## Description

In the not so distant future we are going to end up with duplicate `stable_id` in compara for `gene_member` and `seq_member` schemas. We need to be able to support this, and soon the schema will also be changed. Here is deprecation warning for `fetch_by_stable_id` to be replaced with additional `GenomeDB` constraint, so that `stable_id` only need be unique within a genome and not across the entire schema.

**Related JIRA tickets:**
- ENSCOMPARASW-5270
- ENSCOMPARASW-5360

#### Change 1
- Add deprecation warning on `DEPRECATED.md` and for the `MemberAdaptor::fetch_by_stable_id()` method

#### Change 2
- Was later realised that the deprecation needs to point to an alternative for users, which we hadn't yet provided, so included the new replacement `MemberAdaptor::fetch_by_stable_id_genomeDB()`

## Testing

New unit tests were written for the `MemberAdaptor.t` for both `SeqMemberAdaptor` and `GeneMemberAdaptor` - these didn't exist, it looks like they were previously neglected when the `member` table was replaced by `seq_member` and `gene_member`.

## Notes

The key thing to note in the new tests is the data tables provided with the duplicate `stable_id` for a different `genome_db` in both `gene_member.txt` and `seq_member.txt`.
I have written the new method to be very similar with edgecases matching that of the one being deprecated, but it is possible it could be better written - I went with the safe option of keeping what works.


modules/t/memberAdaptor.t
```diff
@@ -38,86 +39,82 @@ $hs_gdb->db_adaptor($hs_dba);
...
+    is( $s_member->sequence_id, $s_sequence_id, "Fetching seq_member by stable_id genomedb sequence_id" );
 
     done_testing();
 };
 
-=cut
-
 done_testing();
 
+1;
```
twalsh-ebi, line 120: This isn't necessary in a Perl test script.
```suggestion
```

twalsh-ebi (changes requested):
It's looking well! 🙂  I just have a few comments and suggestions.

modules/t/memberAdaptor.t
```diff
@@ -38,86 +39,82 @@ $hs_gdb->db_adaptor($hs_dba);
...
+    is( $g_member->version, $g_version, "Fetching gene_member by stable_id genomedb version" );
+    is( $g_member->display_label, $g_display_label, "Fetching gene_member by stable_id genomedb display_label" );
+    is( $g_member->description, $g_description, "Fetching gene_member by stable_id genomedb description" );
+    is( $g_member->source_name, $g_source_name, "Fetching gene_member by stable_id genomedb source_name" );
+    is( $g_member->dnafrag_id, $g_dnafrag_id, "Fetching gene_member by stable_id genomedb dnafrag_id" );
+    is( $g_member->dnafrag_start, $g_dnafrag_start, "Fetching gene_member by stable_id genomedb dnafrag_start" );
+    is( $g_member->dnafrag_end, $g_dnafrag_end, "Fetching gene_member by stable_id genomedb dnafrag_end" );
+    is( $g_member->dnafrag_strand, $g_dnafrag_strand, "Fetching gene_member by stable_id genomedb dnafrag_end" );
+    is( $g_member->taxon_id, $g_taxon_id, "Fetching gene_member by stable_id genomedb taxon_id" );
+    is( $g_member->genome_db_id, $g_genome_db_id, "Fetching gene_member by stable_id genomedb genome_db_id" );
```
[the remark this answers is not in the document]
CristiGuijarro: I simply forgot them, adding them in :+1:

twalsh-ebi (approved):
It's looking good!

modules/Bio/EnsEMBL/Compara/DBSQL/MemberAdaptor.pm
```diff
@@ -112,6 +103,51 @@ sub fetch_by_stable_id {
...
+        if ( !$genome_db_id ) {
+            throw( "[$genome_db] does not have a dbID" );
+        }
+    }
+
+    $self->bind_param_generic_fetch( $stable_id, SQL_VARCHAR );
+    $self->bind_param_generic_fetch( $genome_db_id, SQL_INTEGER );
+    my $m = $self->generic_fetch_one( 'm.stable_id = ? AND m.genome_db_id = ?' );
+    return $m if $m;
+
```
dthybert, line 137: I would add a comment explaining in which condition you are going to the last bit of code

dthybert (changes requested):

modules/Bio/EnsEMBL/Compara/DBSQL/MemberAdaptor.pm
CristiGuijarro: Do you mean for the versioned stable_ids? For example:
```suggestion
    # In the case of versioned stable_ids $m is not returned so continue with a versioned expectation
```

dthybert: > Do you mean for the versioned stable_ids? For example:

yes

PR #421, opened 2022-04-28
Add Compara data dump-related Jira tickets

twalsh-ebi:
## Description

We need to update the Compara dump registry files each Ensembl release before 
the commencement of the Compara data dump, so that the information is correct 
for the current release.

We also need tickets to keep track of server usage during the data 
dumps, to avoid placing too great a burden on any one server.

## Overview of changes

Data dump-related tickets are added:
- one ticket to update Compara data dump registry files; and
- one server usage tracker ticket for each of six divisions.

## Testing

The updated `jira_recurrent_tickets.json` file was confirmed as being valid JSON.

## Notes

Draft SOPs can be found at the URLs in the ticket descriptions.

### DIFF
--- a/conf/relco/jira_recurrent_tickets.json
+++ b/conf/relco/jira_recurrent_tickets.json
@@ -57,6 +57,48 @@
             "component": "Relco tasks",
             "description": "https://www.ebi.ac.uk/seqdb/confluence/display/EnsCom/Static+documentation+update#Staticdocumentationupdate-Schemadocumentation",
             "summary": "Update Compara schema documentation in Ensembl plugins"
+         },
+         {
+            "assignee": "<RelCo>",
+            "component": "Relco tasks",
+            "description": "https://www.ebi.ac.uk/seqdb/confluence/display/EnsCom/Flat+file+dumps#Flatfiledumps-Updatingdatadumpregistryfiles",
+            "summary": "Update Compara data dump registry files"
+         },
+         {
+            "assignee": "<RelCo>",
+            "component": "Relco tasks",
+            "description": "https://www.ebi.ac.uk/seqdb/confluence/display/EnsCom/Flat+file+dumps#Flatfiledumps-Trackingdatadumpserverusage",
+            "summary": "Fungi data dump server usage"
+         },
+         {
+            "assignee": "<RelCo>",
+            "component": "Relco tasks",
+            "description": "https://www.ebi.ac.uk/seqdb/confluence/display/EnsCom/Flat+file+dumps#Flatfiledumps-Trackingdatadumpserverusage",
+            "summary": "Metazoa data dump server usage"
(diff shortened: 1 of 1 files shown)


conf/relco/jira_recurrent_tickets.json
```diff
@@ -57,6 +57,48 @@
...
+            "assignee": "<RelCo>",
+            "component": "Relco tasks",
+            "description": "https://www.ebi.ac.uk/seqdb/confluence/display/EnsCom/Flat+file+dumps#Flatfiledumps-Updatingdatadumpregistryfiles",
+            "summary": "Update Compara data dump registry files"
+         },
+         {
+            "assignee": "<RelCo>",
+            "component": "Relco tasks",
+            "description": "https://www.ebi.ac.uk/seqdb/confluence/display/EnsCom/Flat+file+dumps#Flatfiledumps-Trackingdatadumpserverusage",
+            "summary": "Fungi data dump server usage"
```
CristiGuijarro, line 71:
```suggestion
            "summary": "Fungi data dump server usage",
            "server": "mysql-ens-compara-prod-4"
```
I put this suggestion here with the idea discussed that Production could incrementally copy the required tables back to the existing release databases on our servers - in which case the server will be set. This would also follow our usage for the next couple of releases, while Production team are understaffed, and we will need to run the FTP dumps. ourselves. Not sure if this will work syntactically with the Jira API - but it's an idea.

CristiGuijarro (changes requested):
We may need to revisit the SOP before this entire consideration - for example, the next couple of releases FTP dumps may come back to us ( @dthybert you mentioned this? ) - in which case we will want to have the FTP dumps tickets for individual divisions in those jira conf files. I leave a line specific comment for the generic SOP usage of servers that has been mentioned before. Overall, I think the whole SOP needs checking and discussing properly within the Compara team and with Production team - this is a change that may need to be saved for e109/e110 :grimacing:

conf/relco/jira_recurrent_tickets.json
twalsh-ebi: Setting the `server` field like this is not currently supported by `create_compara_release_JIRA_tickets.pl`. But even if it we could, I feel we need to ask if we should. Wouldn't this cause the data-dump tickets to be displayed on the Compara server usage dashboard from Compara branching onwards? I'm not sure it would be useful to display them so early.

twalsh-ebi:
I think this may be useful in future, but given the changing situation, it doesn't appear to be needed right now.

conf/relco/jira_recurrent_tickets.json
CristiGuijarro: Not unless the ticket was put in progress from branching would the server usage show the server as being used for the ticket at the time

PR #426, opened 2022-05-05
New production process specific core database cleanup script

CristiGuijarro:
## Description

Since databases are often copied to our production servers using a pattern, on occasion the wrong database for a core species may be copied over - this may be a version intended for another division, or it may be an alternative organism we do not want to process in Compara, for example: `crassostrea_gigas_core_54_107_1` and  `crassostrea_gigas_core_54_107_2` with differing `metakey` `species.poduction_name`.

**Related JIRA tickets:**
- ENSCOMPARASW-5256

#### Change 1
- New script with easy default options. See pod of script.
- `perl ensembl-compara/scripts/production/cleanup_inappropriate_core_dbs.pl --dry_run`

## Testing
This was tested by copying an additional core db into a production server and altering the `meta` table in that, then running the script to ensure it was picked up. It was. This was tested first with the `--dry_run` option, and then without - which promptly removed the imposter core database.

## Notes
There may be other edge cases to watch out for, which would need to be added to the script as and when encountered, I addressed issues seen for Protists in e107 only. The script should naturally die if a core database is missing also - based on inability to load from database. For now, the SOP will be to run with `--dry_run` to check output first. This script cannot be run on servers with which Compara members do not have rw permissions.


ipilizota (approved):
Much appreciated to have a cleanup script for inappropriate core dbs 👍

dthybert (approved):

PR #432, opened 2022-05-09
Function to read in info from GTF file required for GOC score calculation

ipilizota:
## Description

Reading in gene location and strand for all genes in a GTF file - necessary for GOC score calculation.

**Related JIRA tickets:**
- ENSCOMPARASW-5306

## Overview of changes
New function and corresponding `pytest`.
Parsing is done using `gtfparse` library -> added to `requirements.txt`

## Testing
- `pytest`

## Notes
In `requirements.txt`: Shall I keep `gtfparse==1.2.1` or perhaps put `gtfparse>=1.2.1`? Do we have any guidelines on this?

Side note: Genebuild's python gtf adaptor is in development. They use perl solution at the moment.

### DIFF
--- a/scripts/pipeline/orthology_benchmark.py
+++ b/scripts/pipeline/orthology_benchmark.py
@@ -403,9 +403,9 @@ def read_in_gtf(species_name: str, gtf_dir: str) -> pandas.DataFrame:
 
     df = gtfparse.read_gtf(gtf_file)
     df_genes = df[df["feature"] == "gene"][["seqname", "gene_id", "start", "strand"]]
-    df_genes = df_genes.sort_values(["seqname", "start"])
-    df_genes = df_genes.reset_index()
-    df_genes = df_genes.drop("index", 1)
+    df_genes.sort_values(["seqname", "start"], inplace=True)
+    df_genes.reset_index(inplace=True)
+    df_genes.drop("index", 1, inplace=True)
 
     return df_genes
 
--- a/scripts/pipeline/orthology_benchmark.py
+++ b/scripts/pipeline/orthology_benchmark.py
@@ -394,7 +394,6 @@ def read_in_gtf(species_name: str, gtf_dir: str) -> Union[pandas.DataFrame, type
         UserWarning: If the GTF file was not found.
 
     """
-    gtf_file = None
     gtf_file_pattern = f"{species_name.capitalize()}.*.gtf"
 
     try:
--- a/scripts/pipeline/orthology_benchmark.py
+++ b/scripts/pipeline/orthology_benchmark.py
@@ -41,10 +41,10 @@ import re
 import shlex
 import shutil
 import subprocess
-from typing import Dict, List, Tuple, Union
+from typing import Dict, List, Tuple
 import warnings
 
-import gtfparse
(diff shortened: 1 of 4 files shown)


scripts/pipeline/orthology_benchmark.py
```diff
@@ -41,10 +41,12 @@
 import shlex
 import shutil
 import subprocess
-from typing import Dict, List, Tuple
+from typing import Dict, List, Tuple, Union
 import warnings
 
+import gtfparse
```
dthybert, line 47: could you please add a comment beside the import with the link to the library.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -377,6 +379,38 @@ def extract_paralogs(res_dir: str, species_key: str) -> List[Tuple[str, str]]:
...
+        UserWarning: If the GTF file was not found.
+
+    """
+    gtf_file_pattern = f"{species_name.capitalize()}.*.gtf"
+
+    try:
+        gtf_file = list(glob.glob(os.path.join(gtf_dir, gtf_file_pattern)))[0]
+    except IndexError:
+        warnings.warn(f"GTF file for '{species_name}' not found.")
+        return None
```
dthybert, line 403: Is it not dangerous to return a None when there is such error? Why do you not let the exception be raised to the next level?

scripts/pipeline/orthology_benchmark.py
```diff
@@ -377,6 +379,38 @@ def extract_paralogs(res_dir: str, species_key: str) -> List[Tuple[str, str]]:
...
+
+    df = gtfparse.read_gtf(gtf_file)
+    df_genes = df[df["feature"] == "gene"][["seqname", "gene_id", "start", "end", "strand"]]
+    df_genes_sorted = df_genes.sort_values(["seqname", "start"])
+    df_genes_sorted_reset = df_genes_sorted.reset_index()
+    df_genes_final = df_genes_sorted_reset.drop("index", 1)
+
```
dthybert, line 410: you get a new dataframe each time, I would suggest to keep the same variable (i.e df_genes) for all the operation and return df_genes at the end

dthybert (changes requested):

scripts/pipeline/orthology_benchmark.py
ipilizota: Initially I wasn't returning anything here, just raised a warning. In the subsequent lines, there was an `if` statement to check if there was a `gtf_file`, and if yes, proceeded with parsing and returned a data frame. But `pylint` complained because it wanted a function to always return something, not just when the `if` condition was satisfied. Several rounds of attempts led to this.
 
I will replace that warning with an error. I put a warning to kind of keep it consistent with `get_gtf_file` function where a warning is raised if a GTF file isn't found.

scripts/pipeline/orthology_benchmark.py
ipilizota: Actually, I simplified this further 🙂

ens-sb (approved):

dthybert (approved):

PR #436, opened 2022-05-16
A function to collapse tandem paralogs

ipilizota:
## Description

Collapsing tandem paralogs to prevent inflation of GOC scores.

**Related JIRA tickets:**
- ENSCOMPARASW-5305

## Overview of changes
- A new function and corresponding tests

## Testing
- `pytest`

### DIFF
--- a/scripts/pipeline/orthology_benchmark.py
+++ b/scripts/pipeline/orthology_benchmark.py
@@ -41,7 +41,7 @@ import re
 import shlex
 import shutil
 import subprocess
-from typing import Dict, List, Tuple
+from typing import Dict, List, Tuple, Union
 import warnings
 
 import gtfparse  # https://pypi.org/project/gtfparse/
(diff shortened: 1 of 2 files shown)


dthybert (changes requested):
While the code is simple , it may be difficult to  understand it straight away because of pandase collumn name that are not necessary explicite. Since you cannot change this, to solve this I would  give a brief description of the algorithm used to collapse in the main comment section of the function . I would also provide more small comment at key point of the algorithm so we understand in which part of the algorithm we are.

ipilizota:
Column names could be changed but I prefer to keep them as they appear in GTF files. Nevertheless, I'm open to changing them if that'd be more convenient than sticking to the GTF file convention. 

Also, please note that the current code relies on all putative paralogous relationships being recorded, and does not exploit the transitivity of within species paralogy when a relationship is not explicitly listed.

dthybert (approved):

PR #437, opened 2022-05-16
Species tree pipeline based on BUSCO sets: alignment step

ens-sb:
## Description

**Related JIRA tickets:**
- ENSCOMPARASW-5438

#### Change 1
Added BUSCO postprocessing step.
#### Change 2

Added alignment step.

## Testing
The pipeline was tested on a small artificial dataset.


CristiGuijarro (commented):
I think it looks okay. Would it be possible to copy paste or screenshot the terminal with the `nextflow run` command for the test output, so that it can be better visualised?

PR #446, opened 2022-05-26
Placeholders and new modules, update to configs

CristiGuijarro:
## Description

This is an attempt to break up a PR, so it is parts of a couple of subtasks, incomplete as a ticket, but complete as a test functional increment on it's own for the modules and not yet the inclusion in the workflows. More work is in progress, but the review of this PR should make the next reviews easier to spot with changes to fit process.

**Related JIRA tickets:**
- ENSCOMPARASW-5375
- ENSCOMPARASW-4695
- ENSCOMPARASW-5414
- ENSCOMPARASW-5417

#### Change 1
- The `output_path` has been partially implemented as a parameter to specify the FTP directory in which to publish the results
- Another additional parameter is the use of `embassy` - by default this is `false` which will mean it will run only in lsf. Passing this parameter as `true` will put the modules (and the workflow not in this PR) into hybrid lsf/slurm usage.
- The reference directory parameters have been altered to fit the above, to better manage the differences between codon and embassy - at the moment set for testing/development - this may be changed for production.

#### Change 2
- An update on the shared pipelines `nextflow.config` for resources as improvements

#### Change 3
- Tweaks/improvements in the `orthoFinderFactory` and `runOrthoFinder` processes to better fit production purposes and tie in with the query-side of the project
- Addition of `stub` for testing in the reference side modules & the use of `scratch` for faster disk stuff

#### Change 4
- Inclusion of module/process to gunzip a fasta file if it is `gz` compressed, allowing flexibility to pass in either format
- This in the workflow is an optional step based on the file format detected - but will be in another PR

#### Change 5
- Improvements to `queryTaxonSelectionFactory` and `MergeRefToQuery` to work better with nextflow `publishDir` and such.

#### Change 6
- New module to trigger the running of the update orthofinder pipeline in hybrid mode - having split the workflow in two (this is not reflected in this PR, only the module existence)
- Update to module to update orthofinder results with query

### Change 7
- Inclusion of (commented out as an incomplete process) module to copy query orthofinder results to FTP

## Testing
These have been run in Codon and Embassy with and without stubs

## Notes
Emphasis on this is a partial but tested as a complete functional addition PR

### DIFF
--- a/pipelines/orthofinder_modules.nf
+++ b/pipelines/orthofinder_modules.nf
@@ -18,10 +18,13 @@ nextflow.enable.dsl=2
 
 params.orthofinder_exe = ''
 params.dir = ''
-params.ref_destination = ''
+params.ref_destination_farm = ''
+params.ref_destination_cloud = ''
 params.query_destination = ''
 params.taxon_selector_exe = ''
 params.ncbi_taxonomy_url = ''
+params.output_path = ''
+params.embassy = false
 
 /**
 *@input takes list of directories containing fasta files
@@ -31,17 +34,23 @@ process orthoFinderFactory {
 
     label 'slurm_default'
 
-    publishDir "$params.ref_destination"
+    publishDir "${params.ref_destination_cloud}/", mode: 'copy', overwrite: 'true'
 
     input:
     val x
 
     output:
-    val x
+    path "${x}/"
 
     shell:
     """
-    rsync -Lavz ${params.dir}/${x} ${params.ref_destination}/
+    rsync -Lavz ${params.dir}/${x}/ ${x}
+    """
+
+    stub:
+    """
+    mkdir -p ${params.dir}/${x}
+    echo "hello" > ${params.dir}/${x}/test.txt
     """
 
 }
@@ -51,14 +60,18 @@ process orthoFinderFactory {
 */
 process runOrthoFinder {
 
-    label 'slurm_default'
+    label 'slurm_64Gb_32C'
+
+    scratch true
+
+    publishDir "${params.ref_destination_cloud}/${x}", mode: 'copy', overwrite: 'true'
 
     input:
-    val x
+    path x
 
     shell:
     """
-    ${params.orthofinder_exe} -t 32 -a 8 -f ${params.ref_destination}/${x}
+    ${params.orthofinder_exe} -t 32 -a 8 -f ${x}
     """
 
     stub:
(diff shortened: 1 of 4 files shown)


pipelines/orthofinder_modules.nf
```diff
@@ -51,14 +60,18 @@ process orthoFinderFactory {
...
+    publishDir "${params.ref_destination_cloud}/${x}", mode: 'copy', overwrite: 'true'
 
     input:
-    val x
+    path x
 
     shell:
     """
-    ${params.orthofinder_exe} -t 32 -a 8 -f ${params.ref_destination}/${x}
+    ${params.orthofinder_exe} -t 32 -a 8 -f ${x}
```
thiagogenez, line 74: Do the `t` and `a` parameters could be not hard-coded here?

CristiGuijarro: They don't have to be, but for a robust production pipeline, it would be easiest - the idea being that the workflow and resources are optimal for the production environment. This could be seen as an improvement with the idea that the process resource could be dynamic too - in a similar way: https://github.com/Ensembl/ensembl-compara/pull/455#discussion_r898015589

pipelines/orthofinder_modules.nf
```diff
@@ -125,34 +160,34 @@ process mergeRefToQuery {
...
-    val "${query}"
+    path "${query}/"
 
     shell:
     """
-    mkdir ${query};
-    rsync -Lavz ${ref_dir} ${params.query_destination}/${query}/;
-    cp ${query_fasta} ${params.query_destination}/${query};
+    rsync -Lavz ${ref_dir}/ ${query};
+    cp ${query_fasta} ${query}/;
```
thiagogenez, line 176: Is there any sanity check  (task) after this process has been done? Or if this fails, the pipeline will stop anyway, isn't?

thiagogenez:
Hey @ens-sb, As you are working with Nextflow pipelines, would you mind looking at this PF, please?  It looks acceptable to me as it is...

pipelines/orthofinder_modules.nf
CristiGuijarro: There is going to be - there is a ticket for this to include CHECKSUMS etc

ens-sb (approved):

thiagogenez (approved):
just left a few more comments..

PR #449, opened 2022-06-07
fixing pylint

thiagogenez:
## Description

fixing the following error
```
************* Module pyproject.toml
pyproject.toml:1:0: W0012: Unknown option value for '--disable', expected a valid pylint message and got 'no-self-use' (unknown-option-value)
pyproject.toml:1:0: W0012: Unknown option value for '--disable', expected a valid pylint message and got 'relative-import' (unknown-option-value)
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,8 +22,6 @@ warn_unused_configs = true
 max-line-length = 110
 disable = [
     "invalid-name",
-    "no-self-use",
-    "relative-import",
     "similarities",
     "too-few-public-methods",
     "too-many-arguments",


thiagogenez:
Apparently, the solution is just to remove `no-self-use` and `relative-import` from the disabled list. Easy one

CristiGuijarro:
Will need to be applied to main and release/107 and release/106 probably as well as propagated to feature branches too

thiagogenez:
> Will need to be applied to main and release/107 and release/106 probably as well as propagated to feature branches too

For older releases, I will run `git cherry-pick`

CristiGuijarro:
Can we just merge this so I don't keep getting the spam emails from Travis about pylint? It does not require 2 reviewers :sweat_smile:

PR #451, opened 2022-06-09
Update SeqMember and GeneMember adaptor methods to fetch_by_stable_id_GenomeDB()

CristiGuijarro:
## Description
The methods depended on `fetch_by_stable_id` which is to be deprecated in e109. These now first collect the appropriate species and collect the member using `fetch_by_stable_id_GenomeDB` in an adaptor reversal.

**Related JIRA tickets:**
- ENSCOMPARASW-5361

#### Change 1
- Additional lines in the `fetch_by_Transcript()` and `fetch_by_Translation()` methods in the `SeqMemberAdaptor.pm` and `fetch_by_Gene()` in GeneMemberAdaptor.pm`

#### Change 2
- Update to the test db and `memberAdaptor.t` to include the updated methods.

## Testing
New tests, first run locally.

## Notes
This is only a small portion of the changes required for the deprecation, more to come, trying to keep it small.

PR #452, opened 2022-06-09
Update currently unused methods to stop using `fetch_by_stable_id()` …

CristiGuijarro:
…from GeneMember

## Description

There are a few runnables that require an update upon deprecation of `fetch_by_stable_id()` from the `MemberAdaptors`.

**Related JIRA tickets:**
- ENSCOMPARASW-5361

#### Change 1
- Replace the use of stable_ids with dbIDs where possible. Requires method to `fetch_by_Gene()` see PR: https://github.com/Ensembl/ensembl-compara/pull/451
- Pass the `gene_member_id` to other unused method in params instead of `stable_id`. 

#### Change 2
- Cleanup PODs and some commented out parts

## Testing
This has been tested syntactically, and by running in standalone mode to ensure the used methods are not adversely affected.
```
standaloneJob.pl Bio::EnsEMBL::Compara::RunnableDB::ncRNAtrees::NCRecoverEPO --url <ncrnadb> --compara_db <ncrnadb> --mlss_id 40163 --gene_tree_id 1400000003 --epo_db <compara_prev> --debug 9
```

## Notes
Since the methods here that have the change are not used at all, there is no impact on the production running of this code. A unittest has not been written because the methods are unused, and this is very specific to the EPO_EXTENDED MSA method - which in the long term - we are **probably** swapping for Cactus. On line 182 one can see that a method still (probably detrimentally) refers to the `stable_id` in a way in which duplicates could be found. I have opted to leave this in the interest of development time - since the method is unused anyway, and to change it may have a larger functional impact.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/ncRNAtrees/NCRecoverEPO.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/ncRNAtrees/NCRecoverEPO.pm
@@ -15,15 +15,14 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 
-=cut
-
-# POD documentation - main docs before the code
-
-=pod 
-
 =head1 NAME
 
 Bio::EnsEMBL::Compara::RunnableDB::ncRNAtrees::NCRecoverEPO
+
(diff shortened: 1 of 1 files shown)


dthybert (approved):

ipilizota (approved):
Replacements make sense and keep the code equally readable 👍

PR #453, opened 2022-06-10
Update mafft_update and related runnables to `fetch_by_stable_id_GenomeDB()`

CristiGuijarro:
## Description
There are a few runnables that require an update upon deprecation of fetch_by_stable_id() from the MemberAdaptors.

**Related JIRA tickets:**
- ENSCOMPARASW-5361

#### Change 1
- Made change in `Bio::EnsEMBL::Compara::RunnableDB::GeneTrees::FlagUpdateClusters` to store tags to include the `genome_db_id` with the `stable_id`.
- Replaced method to retrieve members from the gene tree adaptors - this was necessary to retrieve the genome_dbs to use the `fetch_by_stable_id_GenomeDB` method rather than the `fetch_by_stable_id` alone in `Bio::EnsEMBL::Compara::RunnableDB::GeneTrees::CopyTreesFromDB` and `Bio::EnsEMBL::Compara::RunnableDB::ProteinTrees::MafftUpdate`.
- Updated `Bio::EnsEMBL::Compara::RunnableDB::ProteinTrees::CopyAlignmentsFromDB` to reflect new `gene_tree_root_tag` values to include `stable_id` and `genome_db_id`.

#### Change 2
- Hoping to rename the `update_mafft` runnable - which is only used in one place anyway.

#### Change 3
- POD cleanup
- Comment clean up and clarify/update
- Removal of redundant code comments to improve readability

## Testing
This has been tested syntactically locally and via Travis.

## Notes
This runnable is included in the protein trees config - but it was part of an unused development project to use a `topup`  clustering method (which as far as I know never worked out - but at least the logic is there). It will likely never be used as it is now when gene trees are overhauled, and could probably be removed from the protein trees pipeline - but at least it should be in working condition as a standalone.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/CopyTreesFromDB.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/CopyTreesFromDB.pm
@@ -1,4 +1,3 @@
-
 =head1 LICENSE
 
 See the NOTICE file distributed with this work for additional information
@@ -18,16 +17,6 @@ limitations under the License.
 
 =cut
 
-=head1 CONTACT
-
-Please email comments or questions to the public Ensembl
-developers list at <http://lists.ensembl.org/mailman/listinfo/dev>.
-
-Questions may also be sent to the Ensembl help desk at
-<http://www.ensembl.org/Help/Contact>.
-
-=cut
-
 =head1 NAME
 
 Bio::EnsEMBL::Compara::RunnableDB::GeneTrees::CopyTreesFromDB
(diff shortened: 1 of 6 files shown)


thiagogenez:
related to #458 and #451

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/CopyTreesFromDB.pm
```diff
@@ -278,15 +253,24 @@ sub write_output {
...
+        my $old_member_id = $leaf->dbID;
+        my $sql = qq/
+        SELECT
+            seq_member_id_current
+        FROM
+            seq_member_id_current_reused_map
+        WHERE
+            stable_id = "$stable_id"
+                AND seq_member_id_reused = $old_member_id;
+            /;
```
thiagogenez, line 273: Is it a common approach to include SQL code like this here? Is there any specified layer for querying the DB?

thiagogenez (approved):
Nice clean up by the way

modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/CopyTreesFromDB.pm
CristiGuijarro: In this case it is because this is an ad-hoc(ish) ehive pipeline specific table in the pipeline database, there is no API/adaptor for this table, so it has to be done like this

thiagogenez: Ah I see - so Is an adaptor need to be created via a ticket?

CristiGuijarro: No, it wouldn't be worth the development, it is purely for an ehive specific process/analysis used in production and should not be part of the Compara API - one of the sorts of things that needs to be considered as we move away from Perl and eHive :grimacing:

ipilizota (approved):

PR #458, opened 2022-06-20
Fix `ImportAltAlleGroupAsHomologies` dependency on `gene_member.stabl…

CristiGuijarro:
…e_id` to use `genome_db`

## Description

There are a few runnables that require an update upon deprecation of `fetch_by_stable_id()` from the `MemberAdaptors`.

**Related JIRA tickets:**
- ENSCOMPARASW-5361

#### Change 1
- Reworked the `copy_and_fetch_gene()` method to first retrieve the correct `gene_member` using the `genome_db` and `stable_id` - altered the copy statement to use the `gene_member.gene_member_id`.

## Testing
Travis and syntactically locally.

## Notes
There may be more examples in the code base in which data is copied using the `[gene|seq]_member.stable_id` which will need changing once the schema changes, particularly for the use of `Bio::EnsEMBL::Compara::Utils::CopyData::copy_data_with_foreign_keys_by_constraint()` method.

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/ImportAltAlleGroupAsHomologies.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/ImportAltAlleGroupAsHomologies.pm
@@ -77,11 +77,15 @@ sub copy_and_fetch_gene {
     my $self = shift;
     my $gene = shift;
 
-    copy_data_with_foreign_keys_by_constraint($self->param('member_dbc'), $self->compara_dba->dbc, 'gene_member', 'stable_id', $gene->stable_id, undef, 'expand_tables');
-
     # Gene Member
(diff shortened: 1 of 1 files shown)


thiagogenez:
#451

thiagogenez (approved):
In line with https://github.com/Ensembl/ensembl-compara/pull/458

would be more locations in the code base you suspect that a similar update would be needed @CristiGuijarro?

CristiGuijarro:
> In line with #458
> 
> would be more locations in the code base you suspect that a similar update would be needed @CristiGuijarro?

You mean for what I mentioned in the note? Yes, if so, I plan to compile a list of those I can find to include them as in line with the schema change - for now they will not break anything, but they will need to be sorted before the schema change :crossed_fingers:

dthybert (approved):

PR #471, opened 2022-07-01
Species tree pipeline improvements

ens-sb:
## Description

**Related JIRA tickets:**
- ENSCOMPARASW-5679
- ENSCOMPARASW-5680

#### Change 1
Added the prepareGenomes process to handle gzipped input genomes.

#### Change 2
Added code to filter out highly repetitive BUSCO genes in order to speed up the GenBlast step.

#### Change 2
Tweaked process resource classes and added a new resource class to the shared nextflow config.

## Testing
The pipeline was run on a set of 17 vertebrate species.

### DIFF
--- a/pipelines/SpeciesTreeFromBusco/main.nf
+++ b/pipelines/SpeciesTreeFromBusco/main.nf
@@ -81,6 +81,7 @@ process buscoAnnot {
     --num_threads ${params.cores} \
     --max_intron_length 100000 \
     --run_busco \
+    --genblast_timeout  32400\
     --busco_protein_file $busco_prot
     """
 }
--- a/pipelines/SpeciesTreeFromBusco/main.nf
+++ b/pipelines/SpeciesTreeFromBusco/main.nf
@@ -43,16 +43,52 @@ if (!params.dir) {
 *@output path to fasta with longest protein isoforms per gene
 *@output path to gene list TSV
 */
+process prepareGenome {
+    label 'rc_4Gb'
+    input:
+      path genome
+
+    output:
+        path "processed/*", emit: proc_genome
+    script:
+        id = (genome =~ /(.*)\..*$/)[0][1]
+        ext = (genome =~ /.*\.(.*)$/)[0][1]
+        if (ext == 'gz')
+            """
+                mkdir -p processed
+                zcat $genome > processed/$id
+            """
+        else
+            """
+                mkdir -p processed
+                ln -s `realpath $genome` processed/$id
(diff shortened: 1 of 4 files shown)


pipelines/SpeciesTreeFromBusco/main.nf
```diff
@@ -260,9 +297,10 @@ process runIqtree {
 workflow {
     println(ensemblLogo())
     prepareBusco(params.busco_proteins)
-    genomes = Channel.fromPath("${params.dir}/*.fas")
+    genomes = Channel.fromPath("${params.dir}/*.*")
+    prepareGenome(genomes)
```
CristiGuijarro, line 301: Is there any possibility that the `params.dir` directory could contain both the compressed and uncompressed genome files? Thinking about a checking step to see if a genome file has already been processed and already exists in the `processed/` dir or skipping if it already exists?

pipelines/SpeciesTreeFromBusco/nextflow.config
```diff
@@ -18,10 +18,11 @@ includeConfig "$ENSEMBL_ROOT_DIR/ensembl-compara/pipelines/nextflow.config"
 
 params {
     dir = "$HPS_HOME/Workspace/busco_tree/species_fasta"
+    gzipped = "true"
```
CristiGuijarro, line 21: Are you still using this? I cannot find it anywhere in this commit?

CristiGuijarro (commented):
Just a couple of questions/comments nothing major. Will re-review with responses when Travis gives the go ahead :+1: :crossed_fingers:

pipelines/SpeciesTreeFromBusco/nextflow.config
ens-sb: No, this slipped through. Not using it actually so will remove it.

pipelines/SpeciesTreeFromBusco/main.nf
ens-sb: I think it should also work in that case as it will only decompress files with ".gz" extension. If that is not present it will just create a symlink.

ens-sb: I think the possibility of having both gzipped and non-gzipped files in the input directory is not that high. But from nextflow's perspective, they would show up as different files so they would create a duplication of species which would probably cause the pipeline to fail downstream.

CristiGuijarro (approved):

PR #483, opened 2022-07-26
Function to calculate GOC score for a pair of genomes

ipilizota:
## Description

This function calculates GOC scores for all putative orthologs between 2 genomes. It assumes orthologs and paralogs have been inferred by OrthoFinder. It uses earlier implemented functions to read in OrthoFinder output files and GTF files. Then it uses `calculate_goc_genes()` to calculate GOC score for each pair of putative orthologs.

**Related JIRA tickets:**
- ENSCOMPARASW-5557

## Overview of changes
New function, corresponding `pytest` and test files.

## Testing
`pytest`

---

## PR review checklist


scripts/pipeline/orthology_benchmark.py
```diff
@@ -628,6 +628,68 @@ def calculate_goc_genes(gene1: str, gene2: str, neighbourhood1: pandas.DataFrame
...
+
+    for pair in orthologs:
+        gene1 = pair[0]
+        gene2 = pair[1]
+
+        # Get gene neighbourhood for each gene
+        try:
+            neighbours1 = get_neighbours(gene1, genome1, n)
+        except ValueError:  # gene1 is a collapsed tandem paralog or somehow doesn't appear in the GTF
+            continue
```
CristiGuijarro, line 672: I wonder if it would be helpful to the user to provide this as an uninterruptible warning - in the case that it somehow does not appear in the GTF as a level of possible debugging: 
```
import warnings
warnings.warn(f"Gene {gene1} is either a collapsed tandem paralog or not present in GTF file")
```
This would be in place of the comment also.

scripts/pipeline/orthology_benchmark.py
```diff
@@ -628,6 +628,68 @@ def calculate_goc_genes(gene1: str, gene2: str, neighbourhood1: pandas.DataFrame
     return (shared_orthologs / (2 * n)) * 100
 
 
+def calculate_goc_genomes(core_name1: str, core_name2: str, gtf_dir: str, res_dir: str, n: int, m: int)\
```
CristiGuijarro, line 631: I wonder if it could be worth providing default values for `n` and `m` in this function? This could be a useless suggestion because I cannot yet picture how this will be used - and I imagine it to be dynamic in the benchmarking use-case - in which if always dynamic a default value would be useless?

CristiGuijarro (approved):
Minor suggestions that are not :100: necessary for approval, to implement/address at @ipilizota 's discretion :+1:

scripts/pipeline/orthology_benchmark.py
ipilizota: I'll do that 🙂

scripts/pipeline/orthology_benchmark.py
ipilizota: In fact, `n=2` by default in function `calculate_goc_genes()` so there is no reason for not having it here as well. I think I'll go with `n=2, m=3` 👍 Thanks 🙂

ipilizota:
Travis issue is being addressed in https://github.com/Ensembl/ensembl-compara/pull/482.

PR #485, opened 2022-07-28
Function to write GOC scores to a file

ipilizota:
## Description

A function to write GOC scores to a file.

**Related JIRA tickets:**
- ENSCOMPARASW-5424

## Overview of changes
New function and corresponding tests.

## Testing
`pytest`

## Notes
Not sure if these `OSErrors` should be tested at all. For some reason it rings a bell that the conclusion from one of the Dev Round Tables was no - because they originate from the system.

---

## PR review checklist


scripts/pipeline/orthology_benchmark.py
```diff
@@ -694,6 +694,37 @@ def calculate_goc_genomes(core_name1: str, core_name2: str, gtf_dir: str, res_di
...
+            msg = f"Could not create a file '{output_file}' for writing."
+            raise OSError(msg) from e
+
+    try:
+        with open(output_file, "a") as file_handler:
+            for i in goc_scores:
+                file_handler.write(f"{i[0]}\t{i[1]}\t{i[2]}\n")
+    except OSError as e:
+        msg = f"Could not open a file '{output_file}' for writing."
+        raise OSError(msg) from e
```
CristiGuijarro, line 725: What about 
```suggestion
        print(f"Could not open a file '{output_file}' for writing.")
        sys.exit(1)
```
Instead of raising OSError after having excepted it? I gather that the purpose of excepting and then raising it is to print the `msg` for the user - I think this would achieve the same with less pain?

CristiGuijarro (commented):
For the purpose of this comment: 

> Not sure if these OSErrors should be tested at all. For some reason it rings a bell that the conclusion from one of the Dev Round Tables was no - because they originate from the system.

(And passing the tests) I have made a suggestion - this suggestion should mean you don't have to worry about what was discussed at the dev round table regarding testing OSErrors - and you just have to test that the message is printed and `raises(CalledProcessError)`? Not sure if this is much simpler, but it is something that was suggested to me at some point :thinking:

ens-sb (approved):

PR #487, opened 2022-08-08
making delimiter flexible into map_row_to_header

thiagogenez:
## Description

A row containing a space as the content of some field may cause an incorrect split action. As a consequence, the number of tokens (words) in a row may not match the number of tokens (words) in the header, and therefore the function `map_row_to_header` dies.

**Related JIRA tickets:**
- ENSCOMPARASW-5406

## Overview of changes
A new non-mandatory argument (called `delimiter_pattern`) has been added to the `map_row_to_header` subroutine. Its default value has been set to `\\s` because it is already in use.

#### Change 1

The addition of a non-mandatory argument with a default value set to `\s+` if nothing is passed.

```Perl
$delimiter_pattern //= '\s+';
```

#### Change 2
The 'LIMIT' parameter of the Perl 'split' function has been set to a negative value (`-1`) to avoid the removal of empty trailing fields. As things are currently implemented, an empty 'strain' field is removed by 'split', which calls for very special handling ([here](https://github.com/Ensembl/ensembl-compara/blob/feature/homology_annotation/modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromMetadataFactory.pm#L383) and [here](https://github.com/Ensembl/ensembl-compara/blob/c1e262ff33b70780a592d567d00af416c1e3bbfd/modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromMetadataFactory.pm#L408)) to ensure that the 'strain' field is not an empty string. 

Therefore, setting the 'LIMIT' to a negative value would leave such empty fields intact and would remove the need for special (hacky-ish?) handling. 

#### Change 3
Improve the error message in the case the function dies due to the different numbers of tokens in the given line and header. The proposal message is to include the number of tokens found in the given parameters. 

```Perl
die "Number of columns in header (", (scalar @head_cols),") do not match row (", (scalar @cols),")" unless scalar @cols == scalar @head_cols;
```

## Testing
Yes, new test have been included.

## Notes
_Optional extra information._

---

## PR review checklist

- [x] Is the PR against an appropriate branch?
- [x] Does the code adhere to coding guidelines?
- [x] Does the code do what it claims to do?
- [x] Is the code readable? Is it appropriately documented?
- [x] Is the logic in the correct place?
- [x] Was the code tested appropriately? Are there unit tests? Are unit tests self-contained and non-redundant?
[...]

PR #493, opened 2022-08-12
applying tab as column separator

thiagogenez:
Description described here #487.

Applying "\t" as field separator when calling the  `map_row_to_header` subroutine to avoid its death when dealing with spaces in the 'strain' field.

**Related JIRA tickets:**
- ENSCOMPARASW-5406

---

## PR review checklist

- [x] Is the PR against an appropriate branch?
- [x] Does the code adhere to coding guidelines?
- [x] Does the code do what it claims to do?
- [x] Is the code readable? Is it appropriately documented?
- [x] Is the logic in the correct place?
- [x] Was the code tested appropriately? Are there unit tests? Are unit tests self-contained and non-redundant?
- [x] Did Travis CI pass for the code in the PR? Is Codecov acceptable based on the included/updated unit tests?
- [x] Will the new code fail gracefully?
- [x] Does the code follow good practice for writing performant code (e.g. using a database transaction rather than repeated queries outside of a transaction)?
- [x] Does it bring in an unnecessary dependency?
- [x] If you are reviewing a new analysis, is it future-proof and pluggable?
- [x] Does the PR meet agile guidelines?

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromMetadataFactory.pm
+++ b/modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromMetadataFactory.pm
@@ -381,7 +381,7 @@ sub _parse_rr_genome_files {
         next if $_ !~ /[a-z]/;
         # the strain column can be blank, leading to mismatch of column numbers
         $_ = $_ . 'NULL' if $_ !~ /[a-z0-9]$/;
-        my $row = map_row_to_header( $_, $header );
+        my $row = map_row_to_header( $_, $header, "\t" );
         # only interested in the genome name
         push @species_names, $row->{name};
     }
@@ -406,7 +406,7 @@ sub _parse_rename_rr_files {
             next if $_ !~ /[a-z]/;
             # the strain column can be blank, leading to mismatch of column numbers
             $_ = $_ . 'NULL' if $_ !~ /[a-z0-9]$/;
-            my $row = map_row_to_header( $_, $header );
+            my $row = map_row_to_header( $_, $header, "\t" );
             # only interested in the genome name and old_name
             $renamed_genomes{$row->{name}} = $row->{old_name};
         }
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromMetadataFactory.pm
```diff
@@ -381,7 +381,7 @@ sub _parse_rr_genome_files {
         next if $_ !~ /[a-z]/;
         # the strain column can be blank, leading to mismatch of column numbers
         $_ = $_ . 'NULL' if $_ !~ /[a-z0-9]$/;
```
twalsh-ebi, line 383: Now that `map_row_to_header` uses a LIMIT of -1, it will not remove empty trailing columns, so It should be possible to remove these lines.

modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromMetadataFactory.pm
```diff
@@ -406,7 +406,7 @@ sub _parse_rename_rr_files {
             next if $_ !~ /[a-z]/;
             # the strain column can be blank, leading to mismatch of column numbers
             $_ = $_ . 'NULL' if $_ !~ /[a-z0-9]$/;
```
twalsh-ebi, line 408: As above, it should now be safe to remove these lines.

twalsh-ebi (commented):
Notwithstanding Travis issues, this looks good to me. My only suggestion would be to remove the lines that are intended to handle an empty strain column, because with recent changes to `map_row_to_header`, such empty rightmost columns will be preserved.

modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromMetadataFactory.pm
thiagogenez: 🦺  - thanks

modules/Bio/EnsEMBL/Compara/RunnableDB/PrepareMaster/UpdateGenomesFromMetadataFactory.pm
thiagogenez: 🦺

PR #581, opened 2023-03-14
fixing HALXS makefile and upgrade perl to 5.25.2

thiagogenez:
## Description

Bumping Perl version 5.14 to 5.26 on the bits regarding HALXS and improving the makefile contents

#### Change 1
Bumping Perl version 5.14 to 5.26 in both files 
```perl
use v5.26.2;
```
#### Change 2
- Improving the message by explaining the reason we *must* use `hdf5@1.10`
- Adding `LDDLFLAGS` into `Makefile-Linuxbrew.PL` to force the inclusion of `hdf5@1.10` location into the `LDDLFLAGS` of Makefile. Otherwise, a manual intervention into the generated `Makefile` would be necessary because the compilation calls for the dynamic link of `hdf5@1.10`'s library.
```perl
LDDLFLAGS => "-shared -O2 -L$hdf5_location/lib,/usr/local/lib,$lib_location/opt/libnsl/lib -fstack-protector-strong",
```

#### Disclaimer
Others `Makefile-*.PL` still needs review as it may be not working anymore

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/HAL/HALXS/Makefile-Linuxbrew.PL
+++ b/modules/Bio/EnsEMBL/Compara/HAL/HALXS/Makefile-Linuxbrew.PL
@@ -14,8 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
-use 5.014002;
+use v5.26.2;
 use ExtUtils::MakeMaker;
 
 # Version of Makefile.pl that assumes that all the dependencies can be found via linuxbrew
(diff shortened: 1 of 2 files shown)


ens-sb (approved):

twalsh-ebi (approved):
This looks good to me, @thiagogenez.

It looks like this also fixes the warning message (`WARNING: Setting ABSTRACT via file 'lib/HALXS.pm' failed`) appearing when installing the `HALXS` module. 👍

PR #590, opened 2023-03-31
removing clear definition of resource class to pick the new default o…

thiagogenez:
…f 1GB

## Description

Remove the clear definition of the resource class of `500Mb_job` which will not be pointing to the new resource class of `1Gb_job`

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/ENV.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/ENV.pm
@@ -145,8 +145,8 @@ sub resource_classes_single_thread {
     my ($self) = @_;
     my $reg_requirement = '--reg_conf '.$self->o('reg_conf');
     return {
-        # 250 Mb seems to be the minimum we need nowadays, especially since we load the registry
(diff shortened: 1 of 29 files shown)


modules/Bio/EnsEMBL/Compara/PipeConfig/ENV.pm
```diff
@@ -145,8 +145,8 @@ sub resource_classes_single_thread {
     my ($self) = @_;
     my $reg_requirement = '--reg_conf '.$self->o('reg_conf');
     return {
-        # 250 Mb seems to be the minimum we need nowadays, especially since we load the registry
-        'default'      => {'LSF' => ['-C0 -M250   -R"select[mem>250]   rusage[mem=250]"', $reg_requirement],               'LOCAL' => [ '', $reg_requirement ] },
+        # 1Gb seems to be the minimum we need nowadays (Because of the fatter Registry ?), especially since we load the registry
```
twalsh-ebi, line 148:
```suggestion
        # 1Gb seems to be the minimum we need nowadays
```

ens-sb (approved):

PR #602, opened 2023-04-20
Precalculate genomes in overlap between collections

twalsh-ebi:
## Description

While running the `cleanup_strains_clusters` step of the `protostomes` protein trees during Ensembl 110 production, there was unduly heavy usage of the pipeline and master database servers, which appeared to be due to the approximately 80,000 `cleanup_strains_clusters` jobs querying the master database for the same information: the overlap genome set that is used to determine whether a cluster contains only genes in the genomes overlapping between the given collection and its reference collection.

With this PR, the overlap genome set is precalculated once in each non-default protein-trees pipeline and saved to a pipeline-wide parameter, so that the necessary database queries are not repeated in every `cleanup_strains_clusters` job.

**Related JIRA tickets:**
- ENSCOMPARASW-6401

## Overview of changes

- A `FindOverlappingGenomes` runnable is added, which calculates the set of genomes overlapping
between a given collection and its reference collection(s), then saves the list of overlap
genomes as the pipeline-wide parameter `overlapping_genomes`.
- All non-default collection gene-tree pipelines are configured to use the `FindOverlappingGenomes` runnable before removing any overlapping data.
- Overlap-species checks are removed from the runnables `RemoveOverlappingClusters` and `RemoveOverlappingDataByMember`, and these runnables now take the overlapping genomes from the `overlapping_genomes` parameter.

## Testing

The changes in this PR were tested on a subset of the insect species expected to be processed in e111.

Additional pipelines for Murinae protein trees, Murinae ncRNA trees and Wheat cultivars were initialised to confirm that the changes in this PR had not caused problems for the gene-tree pipelines in other divisions.

See ENSCOMPARASW-6401 for details on testing.

---

## PR review checklist

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/Metazoa/ComplementaryProteinTrees_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/Metazoa/ComplementaryProteinTrees_conf.pm
@@ -80,6 +80,15 @@ sub core_pipeline_analyses {
         @{$self->SUPER::core_pipeline_analyses},
 
         # include complementary collection-specific analyses
+        {
+            -logic_name => 'find_overlapping_genomes',
+            -module     => 'Bio::EnsEMBL::Compara::RunnableDB::GeneTrees::FindOverlappingGenomes',
+            -parameters => {
+                'collection' => $self->o('collection'),
+            },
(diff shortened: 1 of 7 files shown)


ens-sb (approved):

thiagogenez (approved):
Much better doing this way

PR #608, opened 2023-05-09
Add resource classes with Slurm time limits

twalsh-ebi:
## Description

Compara-specific Hive resource classes are configured in `Bio::EnsEMBL::Compara::PipeConfig::ENV`. With the transition to Slurm, we will need to configure resource classes for the Slurm meadow.

Due to the need to set a time limit on Slurm jobs, it will be necessary to set, for each existing resource class, multiple new resource classes configuring different Slurm time limits, resulting in a significant increase in the number of resource classes.

**Related JIRA tickets:**
- ENSCOMPARASW-6467

## Overview of changes

This PR revises the Hive configuration in the Compara `PipeConfig::ENV` module to mitigate the expansion in the number of resource classes by explicitly configuring only the scheduler submission command arguments of the resource classes (configured by default with a 24-hour time limit on the Slurm meadow), and then dynamically generating the counterparts of those resource classes with time limits set to 1 hour, 168 hours, and for a subset of resource classes, 720 hours.

- Standard resource classes are configured as 'templates', containing only the scheduler submission command arguments, which are then used to dynamically generate the resource-class config. A small number of atypical resource classes are configured directly.
- Resource classes previously configured to use less than 1 Gigabyte are changed to their 1-Gigabyte counterpart (e.g. `250Mb_6_hour_job` becomes `1Gb_6_hour_job`). All 64-core resource classes are changed to be 48-core resource classes, to fit better with the Slurm cluster (e.g. `16Gb_64c_job` becomes `16Gb_48c_job`). All instances of use of these resource classes are updated to use the closest appropriate resource class in the updated configuration.
- A small number of errors in the previous provisional Slurm config are fixed.
- Utility script `dump_compara_hive_rc_config.pl` is added to enable the full set of resource classes to be dumped to a JSON file.

## Testing

The utility script `dump_compara_hive_rc_config.pl` was used to generate a JSON file (attached to ENSCOMPARASW-6467) showing the dynamically generated resource class configuration.

## Notes

- Slurm configuration of MPI resource classes are not included in this update.
- Time limits are not configured for the LSF meadow except in the 6-hour and 720-hour resource classes.

---

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/ENV.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/ENV.pm
@@ -141,303 +141,447 @@ sub executable_locations {
 }
 
 
+
+# Methods 'resource_classes_single_thread' and 'resource_classes_multi_thread' generate
+# a set of resource classes from the resource-class templates, and the former also has
+# additional resource classes. By the end of the process, every resource class should
+# have config for supported meadows, with each meadow config being a two-element arrayref
+# of the form: '[$submission_cmd_args, $worker_cmd_args]'.
+
 sub resource_classes_single_thread {
     my ($self) = @_;
-    my $reg_requirement = '--reg_conf '.$self->o('reg_conf');
-    return {
+
+    my $resource_class_templates = {
         # 1 Gb seems to be the minimum we need nowadays
-        'default' => {
-            'LSF'   => ['-C0 -M1000 -R"select[mem>1000] rusage[mem=1000]"', $reg_requirement],
-            'SLURM' => ['--partition=standard --time=24:00:00 --mem=1g', $reg_requirement],
-            'LOCAL' => ['', $reg_requirement],
-        },
-
-        '500Mb_job'    => {
-            'LSF'   => ['-C0 -M500 -R"select[mem>500] rusage[mem=500]"', $reg_requirement],
-            'SLURM' => ['--partition=standard --time=24:00:00 --mem=500m', $reg_requirement],
-            'LOCAL' => [ '', $reg_requirement ]
-        },
-
         '1Gb_job' => {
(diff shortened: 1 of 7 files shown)


ens-sb (approved):

modules/Bio/EnsEMBL/Compara/PipeConfig/ENV.pm
```diff
@@ -141,303 +141,447 @@ sub executable_locations {
...
         },
 
         '512Gb_job' => {
-            'LSF'   => ['-q bigmem -C0 -M512000 -R"select[mem>512000] rusage[mem=512000]"', $reg_requirement],
-            'SLURM' => ['--partition=bigmem --time=24:00:00 --mem=512g', $reg_requirement],
-            'LOCAL' => ['', $reg_requirement],
+            'LSF'   => '-q bigmem -C0 -M512000 -R"select[mem>512000] rusage[mem=512000]"',
+            'SLURM' => '--partition=bigmem --mem=512g',
         },
+    };
```
thiagogenez, line 210: Suggested code:
```perl
my $resource_class_templates = {};
my @mem_sizes = qw/1 2 4 8 16 24 32 48 64 96 512/;
foreach my $size (@mem_sizes) {
    my $size_str = $size . 'Gb_job';
    my $mem = $size * 1000;
    $resource_class_templates->{$size_str} = {
        'LSF'   => "-C0 -M$mem -R\"select[mem>$mem] rusage[mem=$mem]\"",
        'SLURM' => "--partition=standard --mem=$size"g"",
    };
}
$resource_class_templates->{'512Gb_job'} = {
    'LSF'   => '-q bigmem -C0 -M512000 -R"select[mem>512000] rusage[mem=512000]"',
    'SLURM' => '--partition=bigmem --mem=512g',
};
```

thiagogenez (approved):
Good idea of having `dump_hive_rc_config.pl`

PR #625, opened 2023-05-17
not ready yet for examl_48_cores

thiagogenez:
## Description

During the test of the protein-tree pipeline, the following message comes up from the `beekeeper`

```bash
Could not find a local analysis named 'examl_48_cores' (dataflow from analysis 'raxml_decision')
Could not find a local analysis named 'examl_48_cores' (dataflow from analysis 'examl_32_cores')
```

Indeed, there is no analysis named `examl_48_cores`. I believe this might be an error from the SLURM configuration. Perhaps, we should enforce the LSF meadow for these analyses.  Just reverting this temporary

---

### DIFF
--- a/modules/Bio/EnsEMBL/Compara/PipeConfig/ProteinTrees_conf.pm
+++ b/modules/Bio/EnsEMBL/Compara/PipeConfig/ProteinTrees_conf.pm
@@ -2486,7 +2486,7 @@ sub core_pipeline_analyses {
                     # examl can handle ~4x more patterns
                     '(#raxml_cores# >  16) && (#raxml_cores# <= 32)'    => 'examl_8_cores',
                     '(#raxml_cores# >  32) && (#raxml_cores# <= 48)'    => 'examl_16_cores',
-                    '(#raxml_cores# >  48) && (#raxml_cores# <= 96)'    => 'examl_32_cores',
+                    '(#raxml_cores# >  48) && (#raxml_cores# <= 128)'   => 'examl_32_cores',
                     '(#raxml_cores# >  96)'                             => 'examl_64_cores',
                 ),
             },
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/Compara/PipeConfig/ProteinTrees_conf.pm
```diff
@@ -2487,7 +2487,7 @@ sub core_pipeline_analyses {
                     '(#raxml_cores# >  16) && (#raxml_cores# <= 32)'    => 'examl_8_cores',
                     '(#raxml_cores# >  32) && (#raxml_cores# <= 48)'    => 'examl_16_cores',
                     '(#raxml_cores# >  48) && (#raxml_cores# <= 96)'    => 'examl_32_cores',
```
twalsh-ebi, line 2489: This was previously set to 128 cores. Would it be possible to revert this too?
```suggestion
                    '(#raxml_cores# >  48) && (#raxml_cores# <= 128)'    => 'examl_32_cores',
```

modules/Bio/EnsEMBL/Compara/PipeConfig/ProteinTrees_conf.pm
```diff
@@ -2487,7 +2487,7 @@ sub core_pipeline_analyses {
...
-                    '(#raxml_cores# >  96)'                             => 'examl_48_cores',
+                    '(#raxml_cores# >  96)'                             => 'examl_64_cores',
```
twalsh-ebi, line 2490:
```suggestion
                    '(#raxml_cores# >  128)'                             => 'examl_64_cores',
```

twalsh-ebi (commented):
Thanks @thiagogenez, I must have missed this when updating the resource classes. Since these are MPI jobs it should be OK to keep them as 64-core jobs.

twalsh-ebi (approved):

PR #688, opened 2023-11-06
Species tree pipeline: using miniprot as default annotation engine

ens-sb:
## Description

Certain genomes (e.g. ones with high levels of softmasking) cannot be annotated using GenBlast and ensembl-anno. Hence miniprot was added as a default annotation engine, which also has the benefit of being faster.

**Related JIRA tickets:**
- ENSCOMPARASW-7072

#### Change 1
Miniprot was added as default annotation engine.

#### Change 2
The filtering criteria when collating the annotation results are now stricter: multiple stop codons are disallowed and minimum taxa fraction filtering is performed again at the end of the filtering process.

#### Change 2
Some label and resource class definition updates.

## Testing
The pipeline was tested on the vertebrates and plants divisions with the vertebrates and viridiplantae BUSCO sets respecitively. A test run was performed using the eukaryotic BUSCO set as well.

---

### DIFF
--- a/pipelines/SpeciesTreeFromBusco/main.nf
+++ b/pipelines/SpeciesTreeFromBusco/main.nf
@@ -101,6 +101,8 @@ process dumpVersions {
     (java -jar $params.astral_jar 2>&1| grep "This is") >> $out
     echo "macse:" >> $out
     (java -jar $params.macse_jar -help 2>&1| grep "This is") >> $out
+    echo "miniprot:" >> $out
+    ${params.miniprot_exe} --version >> $out
     """
 
 }
@@ -132,7 +134,7 @@ process prepareGenome {
 *@output path to input genomes info csv
 */
 process prepareGenomeFromDb {
-    label 'retry_with_8gb_mem_c1'
+    label 'retry_with_16gb_mem_c1'
 
     publishDir "${params.results_dir}", pattern: "input_genomes.csv", mode: "copy", overwrite: true
 
@@ -211,7 +213,7 @@ process linkAnnoCache {
 *@output tuple of path to annotation GTF and genome fasta
 */
 process buscoAnnot {
-    label 'retry_with_32gb_mem_c32'
+    label 'retry_with_128gb_mem_c32'
 
     publishDir "${params.results_dir}/anno_cache/$genome", pattern: "annotation.gtf", mode: "copy",  overwrite: true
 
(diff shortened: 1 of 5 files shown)


Simarpreet-Kaur-Bhurji (approved):

pipelines/SpeciesTreeFromBusco/main.nf
```diff
@@ -221,21 +223,35 @@ process buscoAnnot {
...
+    if (params.use_anno == "")
+    """
+        SENS=""
+        if [ "${params.sensitive}" != "" ];
+        then
+            SENS="-M0 -k5"
+        fi
+        ${params.miniprot_exe} \$SENS -t ${params.cores} -d genome.mpi $genome
+        ${params.miniprot_exe} -N 0 -Iu -t ${params.cores} --gff genome.mpi $busco_prot | grep -v '##PAF' \
+        | awk '/mRNA/{split(\$9,a,/ID=|;/); split(\$9,b,/Target=|\s/)}{sub(a[2], b[2]); print}' > annotation.gtf
```
twalsh-ebi, line 235:
```suggestion
        | awk -F "\t" '$3=="mRNA" {match($9, /Target=([^; ]+)/, m); sub(/ID=[^; ]+/, sprintf("ID=%s", m[1]), $9); print}' > annotation.gtf
```
This includes several suggested changes:
- setting an explicit tab delimiter (`-F "\t"`) — to avoid whitespace issues;
- replacing the feature-type column regex with an exact string match `$3=="mRNA"` — to avoid using a regex where possible;
- replacing the `split` function calls with a `match` of the `Target` attribute and a `sub` of the matched target ID into the `ID` attribute — to make the intent clearer.

twalsh-ebi: Whoops! 🤦 I've now put the `match` and `sub` calls into separate blocks so that the substitution would be done on every line, and not just the `mRNA` line.

pipelines/SpeciesTreeFromBusco/nextflow.config
```diff
@@ -45,6 +45,10 @@ params {
     min_taxa = 0.5
     // Name of outgroup taxon:
     outgroup = ""
+    // Use sensitive miniprot mode:
```
twalsh-ebi, line 48:
```suggestion
    // If `sensitive` is a nonempty string, use sensitive miniprot mode:
```

pipelines/SpeciesTreeFromBusco/nextflow.config
```diff
@@ -45,6 +45,10 @@ params {
...
+    sensitive = ""
+    // Use ensembl-anno/GenBlast instead of miniprot:
```
twalsh-ebi, line 50:
```suggestion
    // If 'use_anno' is a nonempty string, use ensembl-anno/GenBlast instead of miniprot:
```

pipelines/SpeciesTreeFromBusco/nextflow.config
```diff
@@ -45,6 +45,10 @@ params {
...
+    use_anno = ""
```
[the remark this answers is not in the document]
twalsh-ebi: I've made specific suggestions for these changes.

pipelines/SpeciesTreeFromBusco/main.nf
```diff
@@ -221,21 +223,34 @@ process buscoAnnot {
...
+    if (params.use_anno == "")
+    """
+        SENS=""
+        if [ "${params.sensitive}" != "" ];
+        then
+            SENS="-M0 -k5"
+        fi
+        ${params.miniprot_exe} \$SENS -t ${params.cores} -d genome.mpi $genome
+        ${params.miniprot_exe} -N 0 -Iu -t ${params.cores} --gff genome.mpi $busco_prot | grep -v '##PAF' \
+        | awk -F "\t" '$3=="mRNA" {match($9, /Target=([^; ]+)/, m)} {sub(/ID=[^; ]+/, sprintf("ID=%s", m[1]), $9); print}' > annotation.gtf
```
twalsh-ebi, line 235: Not forgetting to escape the dollar signs: 🤦
```suggestion
        | awk -F "\t" '\$3=="mRNA" {match(\$9, /Target=([^; ]+)/, m)} {sub(/ID=[^; ]+/, sprintf("ID=%s", m[1]), \$9); print}' > annotation.gtf
```

twalsh-ebi (approved):

PR #718, opened 2024-01-10
WormBase ParaSite ProteinTrees After release WBPS19 fixes

digrigor:
## Description

We need to lower QC threshold for a low quality genome and request more memory for an analysis.


twalsh-ebi (approved):
👍

PR #736, opened 2024-01-30
Resource class updates for Slurm in Protein trees pipeline

Simarpreet-Kaur-Bhurji:
Updated the resource class for Slurm in Protein trees pipeline.


twalsh-ebi (commented):
This is just a few comments to get us started.

twalsh-ebi (approved):

PR #738, opened 2024-02-02
Compara schema update

Simarpreet-Kaur-Bhurji:
Updated the meta table of Compara schema to make the meta_value nullable and created a patch_112_113_b.sql for the same.


Simarpreet-Kaur-Bhurji:
No longer needs to be updated.

PR #913, opened 2025-03-03
Add NHX tag to 'synthetic' subgenomes

JAlvarezJarreta:
## Description

As discussed in Slack, to avoid hardcoding subgenome exceptions for tools like OrthoTree (mainly the subgenomes created to group together the unassigned scaffolds on polyploid chromosome-level assemblies), @twalsh-ebi proposed to add the NHX tag `[&&NHX:unassigned_region_component=1]` instead, so any naming can be used for these subgenomes and the code will not need to be changed.


twalsh-ebi (approved):
Thanks @JAlvarezJarreta, this looks good to me.

PR #954, opened 2025-07-15
[Plants] Update e116 conf files

JAlvarezJarreta:
## Description

As we usually do prior to start working a release, here are all the expected changes related with e116 plans.

### Notes:
I didn't know what number to put in the new protein tree, so put 40 to follow more or less the existing numbering.

As a reminder, whilst _Cajanus cajan_, _Gossypium raimondii_, _Nymphaea colorata_, _Quercus suber_, _Vigna unguiculata_, _Beta vulgaris vulgaris_, _Cannabis sativa_ are being updated, we do not want their corresponding LastZ alignments to be recomputed. Unfortunately I could not recall how to put this into the MLSSs file, so that may be missing.

### DIFF
--- a/conf/plants/mlss_conf.xml
+++ b/conf/plants/mlss_conf.xml
@@ -546,14 +546,6 @@
       </species_set>
     </one_vs_all>
 
-    <!-- Reference species for specific clades -->
-    <one_vs_all method="LASTZ_NET" ref_genome="medicago_truncatula" against="Fabids"/>
-    <one_vs_all method="LASTZ_NET" ref_genome="oryza_sativa">
-      <species_set in_collection="default">
-        <taxonomic_group taxon_name="Liliopsida"/>
-      </species_set>
-    </one_vs_all>
-
     <!-- Reference wheat vs all wheat cultivars and related species -->
     <one_vs_all method="LASTZ_NET" ref_genome="triticum_aestivum">
       <species_set in_collection="wheat_cultivars">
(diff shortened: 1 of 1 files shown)


conf/plants/mlss_conf.xml
```diff
@@ -383,6 +640,10 @@
       homology_range_index="27"
       prefer_for_genomes="hordeum_vulgare"
     />
+    <protein_trees
+      collection="oat_cultivars"
+      homology_range_index="40"
```
twalsh-ebi, line 645: I feel this is a little on the high side, @JAlvarezJarreta, as the way the `homology_range_index` is used to offset gene tree- and homology-related tables (in [lines 55-59 of the OffsetTables runnable](https://github.com/Ensembl/ensembl-compara/blob/93862cf7d4ff1e9169bdb8abfc668992877dff59/modules/Bio/EnsEMBL/Compara/RunnableDB/GeneTrees/OffsetTables.pm#L55-L59)) could leave us with less headroom than we might like in tables such as `gene_tree_node`.

That said, I'm wary of suggesting a specific number here without the opportunity to seek the input of @Simarpreet-Kaur-Bhurji  and @jitender-ebi .

JAlvarezJarreta: Happy to change it for the value you suggest. Just went for a random value that felt "right" based on the others, without having a clue what I was doing 😅 But it has helped me learn something new (or most likely remember something most likely you already shared), so I get something back 😁

twalsh-ebi: Thanks @JAlvarezJarreta. Following some discussion, I've suggested the number 34. This should leave plenty of space in either direction.

twalsh-ebi (approved):

Simarpreet-Kaur-Bhurji (approved):

PR #958, opened 2025-07-18
[Plants] Update e116 conf files - take 2

JAlvarezJarreta:
## Description

- Revert 2 species updates that will not happen at the end in e116.
- Change oat reference genome to the new preference from the community.


conf/plants/mlss_conf.xml
```diff
@@ -635,6 +635,7 @@
     <protein_trees
       collection="oat_cultivars"
       homology_range_index="34"
+      prefer_for_genomes="avena_sativa_ot3098"
```
twalsh-ebi, line 638: The `prefer_for_genomes` attribute is unnecessary in this case because `avena_sativa_ot3098` is only in one cultivar protein-tree collection.
```suggestion
```

twalsh-ebi (approved):

PR #965, opened 2025-08-27
Update Metazoa and Plants reference sets

JAlvarezJarreta:
## Description

Updated Metazoa and Plants reference genomes for the coming update. It also addresses a couple of issues flagged in the previous update.

_Note:_ bear in mind some of these genomes may not be "available" until Beta 14 is released.


conf/references/mlss_conf.xml
```diff
@@ -513,7 +513,7 @@
       <!-- Myrtales -->
       <genome name="eucalyptus_grandis_gca000612305v1"/>
       <!-- Rosales -->
-      <genome name="malus_domestica_gca002114115v1"/>
+      <genome name="malus_domestica_gca916615275v3_gca002114115v1"/>
```
twalsh-ebi, line 516:
```suggestion
      <genome name="malus_domestica_gca916615275v3"/>
```

PR #1033, opened 2026-08-28
Update genomes of interest for the ETM project

JAlvarezJarreta:
## Description

As requested I have updated the `additional_species.json` list with the genomes we are interested to bring on top of e116 genomes for the ETM project.

I have also added the missing genomes to the `species_tree.topology.nw` tree.

And important disclaimer: if it is going to be trouble, happy to use the e116 core databases, but for `avena_sativa_ot3098`, `hordeum_vulgare_barke`, `oryza_sativa` and `oryza_sativa_ir64` we would like to use the new site's core databases instead as they have been reloaded and a few issues have been corrected (however, I cannot recall if these were just metadata issues or something more profound, thus why the alternative may be fine for us).

### DIFF
--- a/conf/etm/additional_species.json
+++ b/conf/etm/additional_species.json
@@ -1,22 +1,14 @@
 {
     "plants": [
-        "aegilops_tauschii",
-        "arabidopsis_thaliana",
         "arabidopsis_thaliana_gca001651475v1gb",
         "arabidopsis_thaliana_gca978657495v1gb",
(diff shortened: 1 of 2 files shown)


twalsh-ebi:
Thanks @JAlvarezJarreta, this looks good for the additional nine ETM genomes.

I will approve this and prepare another PR including these and the default genomes, would be grateful for your review on those.

twalsh-ebi (approved):

